PR #63, opened 2026-03-27
fix: Engine.TensorAdd/TensorSubtract broadcast support

ooples:
## Summary

When tensor shapes don't match, `TensorAdd` and `TensorSubtract` now fall back to `BroadcastAdd`/`BroadcastSubtract` instead of throwing `ArgumentException`. This matches PyTorch/NumPy behavior.

## Why this matters

**Blocks ALL 244+ diffusion models in AiDotNet.** UNet ResBlocks add time embeddings `[1,C,1,1]` to feature maps `[1,C,H,W]` via `Engine.TensorAdd`. Without broadcasting, every diffusion model fails with:
```
ArgumentException: Tensor shapes must match. Got [1, 128, 32, 32] and [1, 128, 1, 1].
```

`TensorMultiply` already had this fallback (`TensorBroadcastMultiply`). This PR adds the same pattern to Add and Subtract.

## Test plan
- [x] 1203 existing tests pass, 0 regressions
- [x] Source library builds clean on both targets

### DIFF
--- a/src/AiDotNet.Tensors/Engines/CpuEngine.cs
+++ b/src/AiDotNet.Tensors/Engines/CpuEngine.cs
@@ -1951,8 +1951,8 @@ public class CpuEngine : ITensorLevelEngine
         if (b == null) throw new ArgumentNullException(nameof(b));
         if (!ShapesMatch(a._shape, b._shape))
         {
-            throw new ArgumentException(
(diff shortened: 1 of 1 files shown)


ooples:
Wrong approach — the fix belongs in AiDotNet's diffusion code which should call BroadcastAdd instead of Engine.TensorAdd when shapes differ.

PR #143, opened 2026-04-12
fix: RBFKernel backward swapped args + missing epsilon gradients

ooples:
## Summary
- Fixed swapped `epsilons`/`output` arguments in `BackwardFunctions.RBFKernelBackward` that caused `epsilons must be 1D [numCenters], got rank 2`
- Fixed `DifferentiableOps.RecordIfActive` to include the full `epsilons` tensor as `inputs[2]` (was only recording `input` and `centers`)
- Backward now properly accumulates epsilon gradients for non-uniform width training
- Added fallback for legacy recordings that only saved a scalar epsilon

## Root cause
The RBFKernel backward function was calling:
```csharp
engine.RBFKernelBackward(gradOutput, inputs[0], inputs[1], output, gammaT)
```
But the signature is `(gradOutput, input, centers, epsilons, output)`, so `output` (rank 2) was passed as `epsilons` and `gammaT` (rank 1) was passed as `output`.

## Test plan
- [ ] All 8 RadialBasisFunctionNetwork tests in AiDotNet should pass
- [ ] RBF backward produces correct gradients for non-uniform widths

### DIFF
--- a/src/AiDotNet.Tensors/AiDotNet.Tensors.csproj
+++ b/src/AiDotNet.Tensors/AiDotNet.Tensors.csproj
@@ -37,6 +37,7 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="AiDotNet" />
+    <InternalsVisibleTo Include="AiDotNetTests" />
     <InternalsVisibleTo Include="AiDotNet.Tensors.Tests" />
     <InternalsVisibleTo Include="AiDotNet.Tensors.Benchmarks" />
   </ItemGroup>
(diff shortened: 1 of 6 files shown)


ooples:
Superseded by PR #146 which fixes the same RBF backward bugs plus Issue #144 (octonion backward), CPU multiplication table sign errors, missing GPU dispatch, and adds 4 integration tests with numerical gradient verification. The FFT/zero-copy changes from this PR are already in main via PR #140.

PR #152, opened 2026-04-12
feat: add ScaleParameters to PersistentTensorRole enum

ooples:
## Summary
- Adds `ScaleParameters` variant to `PersistentTensorRole` enum
- For per-unit scale/width parameters (RBF kernel widths, learnable scales)
- Fixes: layers registering both weights and widths under `Weights` role get silently deduped, causing parameter count mismatches

## Context
RBFLayer registers `_centers` and `_widths` as trainable parameters. Both used `PersistentTensorRole.Weights`, so `RegisterTrainableParameter` replaced `_centers` with `_widths` (same role = replace). Result: only 1 registered parameter instead of 2, causing `SetTrainableParameters` to throw.


ooples:
Closing — this PR accidentally included unrelated spectral filter commits from a local feature branch. Replaced by a clean single-commit PR.

Issue #165, opened 2026-04-15
feat: expose EnableCheckpointing on ICompiledTrainingPlan interface


ooples:
## Motivation

`CompiledTrainingPlan<T>.EnableCheckpointing(int segmentSize)` exists and works (see `src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs:100`), but the class is `internal sealed` so consumer code that receives `ICompiledTrainingPlan<T>` from `CompiledModelCache.GetOrCompileTraining` cannot call it without reflection. This blocks AiDotNet from wiring gradient checkpointing into `TrainWithTape` for large-model memory reduction.

## Proposed change

Add `EnableCheckpointing` to the `ICompiledTrainingPlan<T>` interface:

```csharp
public interface ICompiledTrainingPlan<T> : IDisposable
{
    Tensor<T> Step();
    Tensor<T>[] Gradients { get; }
    int ForwardStepCount { get; }
    int BackwardStepCount { get; }
    void ConfigureOptimizer(...);

    /// <summary>
    /// Enables gradient checkpointing for this plan, reducing memory from
    /// O(N) to O(sqrt(N)) at the cost of ~33% more compute (each segment's
    /// forward runs twice during backward).
    /// </summary>
    /// <param name="segmentSize">Steps per checkpoint segment. 0 = auto (sqrt(N)).</param>
    void EnableCheckpointing(int segmentSize = 0);
}
```

The implementation already exists on the concrete class — just surface it.

## Consumer-side wiring after this ships

In `NeuralNetworks/NeuralNetworkBase.TrainWithTape` we'll wire a single `plan.EnableCheckpointing(segmentSize: 0)` call at plan-creation time, gated on a builder-level opt-in (`ConfigureActivationCheckpointing()`). Memory goes from O(layers) to O(sqrt(layers)) with ~33% compute overhead — the right tradeoff for large transformers training on memory-constrained hardware.

## Acceptance criteria

- `EnableCheckpointing(int)` on `ICompiledTrainingPlan<T>` with matching default parameter
- Unit test: `plan.EnableCheckpointing(4)` followed by `plan.Step()` produces identical gradients to non-checkpointed `plan.Step()` within floating-point tolerance
- Memory profiling test showing peak allocation reduction matches the expected sqrt(N) ceiling

## Reference

Plan: `lucky-waddling-knuth` (AiDotNet repo). Consumer PR is blocked on this landing.

PR #177, opened 2026-04-15
feat: plan stitching — ICompiledPlan.Then() for pipeline composition (#170)

### LINKED ISSUE #170: feat: plan stitching — concatenate compiled plans for pipeline composition
## Motivation

Production pipelines like `tokenizer → encoder → transformer → classifier` are 4 separate compiled plans today, each with its own input/output materialization. Between plans: a full tensor copy and a cache miss on the next plan's input boundary.

**If we can stitch plans, the whole pipeline becomes one compiled kernel** — no inter-plan copies, one trace, one replay. No existing ML framework does this at the pipeline level; PyTorch's `torch.compile` can fuse within a model but the framework doesn't know about pipeline composition.

## Proposed API

```csharp
public interface ICompiledPlan<T> : IDisposable
{
    // ... existing members ...

    /// <summary>
    /// Compose this plan with <paramref name="next"/> — the output of this plan
    /// feeds directly into next's input, without materializing the intermediate
    /// tensor (shared buffer, no copy). Throws if output shape doesn't match
    /// next's input shape.
    /// </summary>
    ICompiledPlan<T> ThenAsync(ICompiledPlan<T> next);
}
```

## Semantics

- Stitched plan runs `this.forward[*]` followed by `next.forward[*]` as one flat delegate array
- The output tensor of `this` becomes the captured input of `next` — same object reference, no copy
- Stitching is associative: `(a.Then(b)).Then(c) == a.Then(b.Then(c))`
- Re-entrant: stitched plans can be stitched with other stitched plans

## Consumer-side impact (AiDotNet)

`AiModelBuilder` already supports pipeline composition via multiple `Configure*` calls. Once plan stitching ships, the builder detects when it has N compile-eligible components and stitches them into ONE plan at Build time.

Example: a RAG pipeline `tokenize → embed → search → rerank → generate` becomes ONE compiled plan. Latency drops by the sum of all inter-component serialization overhead (typically 20-30% of total pipeline time on small batches).

## Acceptance criteria

- `a.Then(b).Execute(input)` produces bitwise-identical results to `b.Execute(a.Execute(input))` for 100 random inputs.
- Allocation profile shows **zero** intermediate tensor materialization between `a` and `b` in the stitched plan.
- Stitching mismatched shapes throws a descriptive exception at stitch time, not at execute time.

## Reference

Plan: `lucky-waddling-knuth` Part C PR 10. No framework competitor — unique AiDotNet feature.

ooples:
## Summary

- Adds `Then(ICompiledPlan<T> next)` to the public `ICompiledPlan<T>` interface — composes two compiled inference plans into one stitched plan whose `Execute()` runs both as a single flat execution.
- Implementation in `CompiledInferencePlan<T>`: splices step arrays with a synthetic in-place buffer-copy "stitch.boundary" step. Both boundary buffers were pre-allocated at compile time, so no new `Tensor<T>` is materialized between A and B.
- Naming: issue spec called this `ThenAsync`, but the return type is synchronous `ICompiledPlan<T>`. Used `Then` to match standard C# convention (Async suffix only on Task-returning methods).

## Why

Production pipelines like `tokenizer → encoder → transformer → classifier` are 4 separate compiled plans today, each with its own input/output materialization between stages. Stitching turns the whole pipeline into one compiled kernel — no inter-plan tensor copies, one trace, one replay. No existing ML framework does this at the pipeline level (PyTorch's `torch.compile` only fuses within a single model).

This is a building block. Once shipped, `AiModelBuilder` can detect N compile-eligible components in a pipeline and stitch them at Build time (the issue's stated downstream consumer).

## Acceptance criteria coverage

| Criterion | Where verified |
|---|---|
| `a.Then(b).Execute()` is bitwise-identical to `b.Execute(a.Execute())` for 100 random inputs | `Then_Execute_BitwiseIdenticalToSequentialExecution_Across100RandomInputs` |
| Allocation profile shows zero intermediate tensor materialization | `Then_StitchedPlan_HasExactlyOneBoundaryStep_NoNewTensorMaterialization` (structural) — stitched StepCount = A.StepCount + 1 (boundary memcpy) + B.StepCount, no tensor allocations |
| Shape mismatch throws descriptive exception at stitch time, not execute time | `Then_WithMismatchedShapes_ThrowsAtStitchTime_NotExecuteTime` — error message names both shapes |

## Beyond the criteria

| Test | Invariant |
|---|---|
| `Then_WithNullNext_ThrowsArgumentNullException` | Standard arg validation |
| `Then_WithNonBuiltInImplementation_ThrowsNotSupportedException` | Stitching needs concrete-type access to the step array; foreign implementers can opt in with their own stitcher rather than us guessing |
| `Then_IsAssociative_LeftFoldEquivalentToRightFold` | `(a.Then(b)).Then(c)` ≡ `a.Then(b.Then(c))` — same step count, same numeric output |
| `Then_IsReEntrant_StitchedPlanCanBeStitchedAgain` | A stitched plan can be the operand of another `Then` call |
| `Then_StitchedPlan_Dispose_DoesNotDisposeOriginalPlans` | Stitched plan owns no new pinned handles; originals stay alive — lets one plan participate in multiple stitched pipelines |

## Trade-off vs the issue's literal "same object reference, no copy"

True buffer aliasing (one Tensor object's storage swapped for another's) would require Tensor surgery — each step's `Execute` closure captured the original Tensor reference at trace time, so swapping `_data` references on the boundary tensors only works if all closures dereference through that field. That's a deeper change.

Shipping the in-place memcpy gets the **allocation** acceptance criterion (no new Tensor materialized) and the structural step-count invariant. It does move data once at the boundary (memcpy on existing buffers — no allocation). True aliasing is a follow-up if profiling shows the boundary memcpy is meaningful relative to the surrounding GEMM work (it shouldn't be — boundary buffer is the model's intermediate representation, typically much smaller than each plan's compute).

## Breaking change

Adding `Then()` to `ICompiledPlan<T>` is **source-breaking** for downstream consumers that implement the interface directly, and **binary-breaking** at runtime for already-compiled external implementers. Same situation as #165 (EnableCheckpointing). Documented in interface xmldoc with the same rationale: no default-interface-member polyfill possible because net471 doesn't support DIMs.

In-repo there's only one implementer (`CompiledInferencePlan<T>`, internal sealed) — updated alongside the interface. No external mocks/stubs in the test project.

## Test plan

- [x] 8/8 new stitching tests pass on both net471 and net10.0
- [x] Compilation/Compiled regression sweep: 178/178 pass on net10.0 (4 skipped = pre-existing perf gates)
- [x] Source builds clean on both targets, 0 warnings, 0 errors

Closes #170

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledInferencePlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledInferencePlan.cs
@@ -45,6 +45,20 @@ internal sealed class CompiledInferencePlan<T> : ICompiledPlan<T>
     private bool _disposed;
 
     /// <summary>
+    /// Upstream final-output tensor that this plan was last stitched onto,
+    /// or <c>null</c> if this plan has never been a stitch target.
+    /// <see cref="ThenAsync"/> rebinds this plan's captured input to the
+    /// upstream's output, so a second stitch targeting this plan must
+    /// either reuse the same upstream tensor (idempotent; fine — the
+    /// associativity case <c>(A.Then B).Then C</c> vs <c>A.Then(B.Then C)</c>
+    /// rebinds every intermediate to the same target) or go through a fresh
+    /// compiled plan. Pointing the rebind at a different upstream would
+    /// silently rewire any earlier stitched pipeline — the bug we're
+    /// guarding against.
+    /// </summary>
+    private Tensor<T>? _lastStitchUpstream;
+
+    /// <summary>
     /// Whether this plan has been disposed. Internal because
     /// <see cref="ThenAsync"/>'s stitched-plan Execute uses it to detect
     /// a caller who disposed a source plan while the stitched plan is
@@ -156,6 +170,24 @@ internal sealed class CompiledInferencePlan<T> : ICompiledPlan<T>
             throw new ArgumentException(
                 "Next plan has no captured input tensor — it cannot be stitched onto.", nameof(next));
 
+        // Reject fan-out reuse from a DIFFERENT upstream. Stitching rebinds
+        // nextPlan's input storage to this plan's output; calling Then again
+        // on the same nextPlan from a different upstream would silently
+        // rewire any earlier stitched pipeline because nextPlan is shared by
+        // reference. Idempotent re-stitch to the same upstream is allowed —
+        // the associativity case (A.Then B).Then C vs A.Then(B.Then C) ends
+        // up rebinding every intermediate to the same target tensor.
+        if (nextPlan._lastStitchUpstream is not null &&
+            !ReferenceEquals(nextPlan._lastStitchUpstream, _finalOutput))
+        {
+            throw new InvalidOperationException(
+                "This plan has already been stitched onto a different upstream. " +
+                "Reusing it as a stitch target with a new upstream would rebind " +
+                "its input storage and silently invalidate the earlier pipeline. " +
+                "Recompile the downstream plan to get a fresh instance for each " +
+                "distinct upstream.");
+        }
+
         // Validate at stitch time, not at execute time, per acceptance
         // criterion #3. Compare _shape arrays element-wise so the error
         // message can name the mismatching dim.
@@ -176,6 +208,7 @@ internal sealed class CompiledInferencePlan<T> : ICompiledPlan<T>
         // if the invariants don't hold (pre-filtered above, so in practice
         // this is a belt-and-suspenders check).
         nextPlan._compiledInputTensor.RebindStorageFrom(_finalOutput);
+        nextPlan._lastStitchUpstream = _finalOutput;
 
         // Splice: [thisSteps..., nextSteps...]. No boundary step — the
         // rebind above is a one-shot stitch-time operation, not a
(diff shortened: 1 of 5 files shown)


ooples:
## Audit follow-up: all three gaps addressed in `9c8d06b`

Independent audit of this PR against issue #170 flagged three divergences from the spec. All three are now fixed:

### (a) `Then` → `ThenAsync` — renamed to match the issue verbatim

Still synchronous (no `Task` return — stitch-time work is just shape validation + storage rebind + step-array splice, nothing to yield on). The interface xmldoc now includes a "why `ThenAsync` without `Task`" note with a pointer to VSTHRD200-style suppression for anyone whose analyzer is noisy about it.

### (c) True buffer aliasing via `Tensor.RebindStorageFrom`

New internal method on `TensorBase<T>` that swaps `_data` and `_storage` references. Both fields had their `readonly` modifier removed — documented on both fields that rebind is the ONLY intended mutation path. After `ThenAsync`, next's captured input tensor **shares backing storage** with this's final output (same `Vector<T>`, same `TensorStorage<T>` — verified by the new `ThenAsync_AfterStitch_NextInputAndThisOutputShareStorage` test).

Consequence: the boundary memcpy step is **gone**. Stitched step array is now `[this.steps, next.steps]` verbatim — one flat delegate array per the issue spec. `StepCount = A.StepCount + B.StepCount` (no +1). Updated the structural test from `HasExactlyOneBoundaryStep_NoNewTensorMaterialization` to `StepCountIsSumOfSources_NoBoundaryStep` to reflect this.

Trade-off documented in xmldoc: the rebind mutates `next`'s input storage, so `next` running standalone afterwards reads data from `this`'s most recent execution. Callers who need both standalone and stitched use should compile `next` twice.

View caveat also called out: tensor views constructed before the rebind continue to reference the old storage. Doesn't bite the common case (plans don't typically capture views of their own input leaf) but the edge is on the record.

### (b) Disposed-source-plan defensive check

Stitched plan now holds strong refs to source plans via a new `_sourcePlans` field (empty sentinel for non-stitched plans — zero overhead for the common path). Two behaviours:

- `ThenAsync` throws `ObjectDisposedException` immediately if either source is already disposed.
- `Execute()` checks each source's `IsDisposed` at the top before iterating the shared step array — "caller disposed A before running the stitched plan" now produces a crisp error with the source index named, instead of UB through freed GCHandles.

### Tests (12/net10.0, 11/net471)

11 stitching tests + 1 allocation probe. New additions this round:
- `ThenAsync_AfterStitch_NextInputAndThisOutputShareStorage` (verifies literal storage-reference identity — the point of option c)
- `ThenAsync_Execute_WithDisposedSourcePlan_ThrowsObjectDisposedException` (option b, violation path)
- `ThenAsync_WithDisposedSource_ThrowsAtStitchTime` (option b, early detection for disposed `this` or `next`)

Regression sweep: 182/182 Compilation tests pass on net10.0. 34/34 focused compilation suite on net471.

PR #194, opened 2026-04-17
feat(tensor): add Tensor<T>.CopyTo(Span<T>) for strided-aware single-pass copy

ooples:
## Summary

Adds a first-class API on ``Tensor<T>`` for copying the tensor's logical contents into a flat destination span without allocating an intermediate materialized tensor. Handles both contiguous and strided (post-``Permute`` / post-``Transpose`` / post-``Slice``) layouts transparently.

## Motivation

Today, copying a permuted tensor into a flat buffer requires the ``.Contiguous().AsSpan().CopyTo(dst)`` idiom:

\`\`\`csharp
var hwc = engine.TensorPermute(chw, [1, 2, 0]);
hwc.Contiguous().AsSpan().CopyTo(dst);  // 2 allocs, 3 copies
\`\`\`

``Contiguous()`` allocates a full-size tensor and copies through the strided iterator; ``AsSpan()`` then takes one more copy into the destination. On per-sample hot paths (e.g. CIFAR / CIFAR-100 / EuroSat loaders processing tens of thousands of 32x32x3 samples) that's hundreds of thousands of spurious tensor allocations and hundreds of MB of redundant copy work.

With ``CopyTo(Span<T>)``:

\`\`\`csharp
var hwc = engine.TensorPermute(chw, [1, 2, 0]);
hwc.CopyTo(dst);  // 0 extra allocs, 1 copy
\`\`\`

Approximately halves both allocation count and per-element work on the motivating data-loader path.

## API

\`\`\`csharp
public void CopyTo(Span<T> destination);
public void CopyTo(Memory<T> destination);
\`\`\`

Behavior:
- Contiguous fast path (with or without storage offset) delegates to ``Span<T>.CopyTo``, which lowers to ``Buffer.Memmove`` on modern runtimes.
- Strided fallback mirrors the existing ``Contiguous()`` stride-decomp loop so the two methods stay behaviorally identical.
- Throws ``ArgumentException`` on an undersized destination (consistent with ``Span<T>.CopyTo`` semantics).
- Throws ``NotSupportedException`` via ``ThrowIfSparse()`` for ``SparseTensor<T>``.
- Oversized destinations have trailing elements untouched.

## Tests

``tests/AiDotNet.Tensors.Tests/LinearAlgebra/TensorCopyToTests.cs`` — 9 cases:

- ``Contiguous_Copy_MatchesSource``
- ``Contiguous_Copy_ToOversizedDestination_LeavesTrailUntouched``
- ``Contiguous_Copy_ToUndersizedDestination_Throws``
- ``StridedAfterPermute_Copy_MatchesContiguousThenCopy`` — locks the CHW→HWC axis mapping; the specific regression that motivated this API
- ``StridedAfterTranspose_Copy_MatchesLogicalOrder``
- ``StridedAfterSlice_Copy_MatchesLogicalOrder``
- ``Memory_Overload_DelegatesToSpan``
- ``FloatTensor_CopyTo_WorksSymmetrically`` — generic-dispatch sanity check
- ``Contiguous_Copy_DoesNotAllocate`` — net5+ only; uses per-thread allocation counter to assert the fast path is zero-alloc

\`\`\`
Passed! - Failed: 0, Passed: 9, Skipped: 0, Total: 9, Duration: 271 ms
\`\`\`

## Followup

Once this ships in a Tensors NuGet version, the CIFAR / CIFAR-100 / EuroSat NHWC fixes in [ooples/AiDotNet#1153](https://github.com/ooples/AiDotNet/pull/1153) can drop the ``.Contiguous().AsSpan().CopyTo(dst)`` idiom in favor of the one-line ``.CopyTo(dst)``. That removes the second allocation and the second copy from every sample on the hot path.

## Test plan
- [x] Contiguous tensor copies produce matching output
- [x] Strided-after-permute copy matches ``Contiguous().AsSpan().CopyTo`` result
- [x] Strided-after-transpose and strided-after-slice paths exercised
- [x] Destination size validation (under / exact / over)
- [x] Generic dispatch works for both float and double
- [x] Zero-allocation on the fast path (net10.0)
- [x] Memory<T> overload correctness

### DIFF
--- a/src/AiDotNet.Tensors/LinearAlgebra/Tensor.cs
+++ b/src/AiDotNet.Tensors/LinearAlgebra/Tensor.cs
@@ -192,6 +192,91 @@ public class Tensor<T> : TensorBase<T>, IEnumerable<T>
     }
 
     /// <summary>
+    /// Copies the tensor's logical contents, in row-major order, into
+    /// <paramref name="destination"/>. Handles both contiguous and strided
+    /// (e.g., post-<see cref="Transpose()"/> / post-permute) layouts without
+    /// allocating an intermediate contiguous tensor.
+    /// </summary>
+    /// <param name="destination">Destination span. Must have length at least
+    /// <see cref="TensorBase{T}.Length"/>.</param>
+    /// <exception cref="ArgumentException"><paramref name="destination"/> is shorter than
+    /// <see cref="TensorBase{T}.Length"/>.</exception>
+    /// <exception cref="NotSupportedException">This tensor is a <see cref="SparseTensor{T}"/>.</exception>
+    /// <remarks>
+    /// <para>
+    /// Use this instead of <c>Contiguous().AsSpan().CopyTo(dst)</c> when the
+    /// goal is to spill a logical tensor into a flat buffer. The
+    /// materialize-then-copy idiom allocates a second full-size tensor and
+    /// does two passes over the data; <c>CopyTo</c> does one pass straight
+    /// into the destination. For a strided [3,32,32] permuted-to-HWC tensor
+    /// this roughly halves both allocation count and per-element work.
+    /// </para>
+    /// <para>
+    /// Bulk path: when this tensor is contiguous (with or without a storage
+    /// offset), the copy reduces to a single <see cref="Span{T}"/>-level
+    /// <c>CopyTo</c>, which on modern runtimes lowers to <c>memmove</c>.
+    /// Strided path: uses cached row-major strides to decompose the flat
+    /// destination index into per-axis indices, matching
+    /// <see cref="Contiguous"/>'s fallback arithmetic so behavior is
+    /// identical.
+    /// </para>
+    /// <para>
+    /// The destination need not match this tensor's shape rank; only length
+    /// matters. Writes exactly <see cref="TensorBase{T}.Length"/> elements
+    /// starting at <c>destination[0]</c>; any trailing elements of a larger
+    /// destination are left untouched.
+    /// </para>
+    /// </remarks>
+    public void CopyTo(Span<T> destination)
+    {
+        ThrowIfSparse();
+        if (destination.Length < Length)
+        {
+            throw new ArgumentException(
+                $"Destination span is too small: need {Length} elements, got {destination.Length}.",
+                nameof(destination));
+        }
+
+        if (IsContiguous)
+        {
+            // Fast path for contiguous tensors (with or without storage
+            // offset) — a single Span<T>.CopyTo delegates to Buffer.Memmove.
+            _data.AsSpan().Slice(_storageOffset, Length).CopyTo(destination);
+            return;
+        }
+
+        // Strided fallback. Identical decomposition to Contiguous() so the
+        // two methods stay behaviorally consistent — any future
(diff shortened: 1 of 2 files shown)


ooples:
## Independent audit against the original PR

Ran a wider review before requesting merge. Findings below.

### Covered edges (13 tests, all green on net10.0)

| Area | Test |
|---|---|
| Contiguous fast path | ``Contiguous_Copy_MatchesSource`` |
| Oversized destination | ``Contiguous_Copy_ToOversizedDestination_LeavesTrailUntouched`` |
| Undersized destination | ``Contiguous_Copy_ToUndersizedDestination_Throws`` |
| Strided post-permute rank-3 | ``StridedAfterPermute_Copy_MatchesContiguousThenCopy`` |
| Strided post-transpose | ``StridedAfterTranspose_Copy_MatchesLogicalOrder`` |
| Strided post-slice | ``StridedAfterSlice_Copy_MatchesLogicalOrder`` |
| ``Memory<T>`` overload | ``Memory_Overload_DelegatesToSpan`` |
| Generic dispatch (float vs double) | ``FloatTensor_CopyTo_WorksSymmetrically`` |
| Zero-alloc on fast path | ``Contiguous_Copy_DoesNotAllocate`` (net5+ only) |
| **Zero-length tensor (new)** | ``ZeroLength_Copy_IsNoOp`` |
| **Rank-0 scalar (new)** | ``Rank0Scalar_Copy_WritesSingleElement`` |
| **Rank-4 permute (new)** | ``Rank4AfterPermute_Copy_MatchesContiguousThenCopy`` |
| **Sparse rejection (new)** | ``SparseTensor_CopyTo_Throws`` |

### Audit items considered but left out

- **Matrix / Vector**: inherit from ``MatrixBase<T>`` / ``VectorBase<T>``, not ``TensorBase<T>`` — they're always dense/contiguous by construction, so a ``CopyTo(Span<T>)`` on them would just be ``AsSpan().CopyTo()``. Not worth a sibling API.
- **Broadcasting (stride-0)**: the stride-decomposition naturally handles ``stride == 0`` (same source bytes re-read per broadcast position). There's no public API in Tensors that returns a broadcast-to-larger-shape view today, so no concrete call site regresses if we leave that untested. Happy to add it if broadcast views get exposed later.
- **Storage-offset contiguous views**: covered implicitly by ``StridedAfterSlice`` (TensorSlice produces a view that may be contiguous-with-offset). The contiguous-fast-path in ``CopyTo`` correctly slices with ``_storageOffset`` — same code path as ``Contiguous()``.
- **``TensorView<T>``**: already has its own ``CopyTo(Span<T>)`` using the exact same contiguous / strided bifurcation pattern. Confirms the design is consistent with existing repo conventions; no changes needed there.

### Consistency check against ``Contiguous()``

The strided fallback copies byte-for-byte identically to ``Contiguous().AsSpan().CopyTo(dst)`` — both ``StridedAfterPermute_Copy_MatchesContiguousThenCopy`` and ``Rank4AfterPermute_Copy_MatchesContiguousThenCopy`` explicitly compare the two paths. Any future optimization applied to ``Contiguous()``'s stride decomp should be mirrored here (and vice versa).

Ready for review.

PR #280, opened 2026-05-01
fix(#279): GradientTape leak + single-thread backward across all linear ops

### LINKED ISSUE #279: GradientTape lifecycle leaks ~30% of model size per Train() — universal across consumers, blocks AiDotNet#1227 + #1228
## Summary

`AiDotNet`'s tape-based training path leaks managed memory at a rate proportional to the saved-for-backward intermediates per step. Tensors that should be released after `GradientTape.Dispose()` (or `Reset()`) survive Gen2 GC and accumulate. The leak is universal across consumers — confirmed on a tiny FFN AND on a 1-layer Transformer — but transformer/attention models hit the wall first because they save many intermediates per step.

This is the upstream root cause for two AiDotNet-side issues:
- ooples/AiDotNet#1227 — Transformer.Train: 52 GB RAM growth → host OS lockup
- ooples/AiDotNet#1228 — Transformer.Train runs single-threaded (Predict path is multi-threaded)

Both are downstream symptoms; this report is the upstream investigation that localizes the cause to this repo's `GradientTape<T>` / `CompiledDelegateChain<T>` lifecycle and the single-threaded `MatMulBackward` dispatch.

## Reproduction (50-line standalone)

`AiDotNet.Tensors` 0.69.0 + `AiDotNet` 0.171.0 + .NET 10, Windows 11, 16-core. Configuration consistent with #1227 / #1228.

### Transformer leak (matches #1227 profile)

```csharp
var arch = new TransformerArchitecture<float>(
    inputType: InputType.TwoDimensional,
    taskType: NeuralNetworkTaskType.SequenceClassification,
    numEncoderLayers: 1, numDecoderLayers: 0, numHeads: 4,
    modelDimension: 128, feedForwardDimension: 512,
    inputSize: 64, outputSize: 256,
    maxSequenceLength: 64, vocabularySize: 256);

var t = new Transformer<float>(arch, lossFunction: new CategoricalCrossEntropyLoss<float>());
t.SetTrainingMode(true);

var input = new Tensor<float>(new[] { 1, 64 });
var target = new Tensor<float>(new[] { 1, 256 });
var rng = new Random(42);

void Fill() {
    for (int s = 0; s < 64; s++) input[0, s] = rng.Next(256);
    for (int v = 0; v < 256; v++) target[0, v] = 0f;
    target[0, rng.Next(256)] = 1f;
}

Fill();
for (int i = 0; i < 5; i++) t.Train(input, target);   // warmup
GC.Collect(); GC.WaitForPendingFinalizers(); GC.Collect();

long start = GC.GetTotalMemory(true);
for (int i = 0; i < 200; i++) { Fill(); t.Train(input, target); }
GC.Collect(); GC.WaitForPendingFinalizers(); GC.Collect();
long end = GC.GetTotalMemory(true);

Console.WriteLine($"managed heap delta per call: {(end-start)/200.0:F0} bytes");
```

**Observed (verbatim run):**

```
call=  50 heap=207.9MB ws=347.4MB alloc_total=794.0MB
call= 100 heap=396.9MB ws=574.1MB alloc_total=1520.8MB
call= 150 heap=585.9MB ws=775.4MB alloc_total=2247.2MB
call= 200 heap=774.8MB ws=1073.7MB alloc_total=2973.5MB
=== Delta over 200 Train calls (2.95s, 14.8ms/call) ===
managed heap: 18.87MB -> 774.84MB (delta 755.97MB, 3963465 B/call)
working set : 92.25MB -> 1073.70MB (delta 981.45MB)
total alloc : 66.48MB -> 2973.52MB (delta 2907.05MB, 14884.1 KB/call)
```

**3.78 MiB/call** post-Gen2 retention. Linear, not asymptotic. Extrapolates to ~211 GiB at 56k calls (matches the #1227 reporter's 52 GiB at ~14k calls, where the host OS locked up).

### Universality test — tiny FFN

Same harness, replace the Transformer with a default `FeedForwardNeuralNetwork<float>` (input=8 -> output=4, default architecture, ~hundred params):

```
=== Tiny-FFN leak Delta over 500 Train calls (0.30s, 0.60ms/call) ===
managed heap: 0.71MB -> 3.40MB (delta 2754.0KB, 5640 B/call)
total alloc : 348.57MB total, 713.9 KB/call
```

**5.6 KiB/call** even on a model with no attention layers. Confirms the leak is not transformer-specific — it scales with the number of saved-for-backward intermediates per step.

## What was tried that did NOT mitigate

Adding explicit `tape.Reset()` after `opt.Step` and clearing the local gradient dictionaries right before scope exit, in `NeuralNetworkBase.TrainWithTape`:

```csharp
opt.Step(context);
tape.Reset();          // explicit reset before Dispose
allGrads.Clear();
grads.Clear();
```

**Result:** identical leak rate (3,963,137 B/call vs 3,963,465 B/call baseline — within noise). So the retained tensors are NOT rooted by the `GradientTape` instance the consumer holds, nor by the consumer's grads dictionaries. The roots are elsewhere inside this package.

`AiDotNet`-side `using var tape = new GradientTape<T>();` is already in place at `NeuralNetworkBase.TrainWithTape` — disposal is wired correctly at the consumer boundary.

## Probable internal roots (Tensors-side)

Reflection on `AiDotNet.Tensors.Engines.Autodiff.GradientTape<T>` (0.69.0) shows three interesting fields:

```
static TapeEntryArena<T>     _cachedArena          // process-wide static, suspicious
       CompiledDelegateChain  _cachedDelegateChain  // per-instance, but holds BackwardStep<T>[] _steps
       TapeEntryArena<T>      _entries
```

Combined with the #1227 reporter's `dotnet-trace` profile showing:

| Method | Per-Train calls |
|---|---|
| `TensorBase.Clone()` | ~15x (forward intermediates: Q, K, V, scores, FFN, residual) |
| `BackwardFunctions.MatMulBackward` | ~17x |
| `CompiledDelegateChain.Execute` | 1x |
| `GradientTape.ComputeGradientsViaGraph` | 1x |
| Gen2 GC | ~1 per 50 calls (constant promotion + retention) |

…the most likely roots are:

1. **`_cachedArena` (static) accumulates `TapeEntry<T>` instances across `GradientTape<T>` lifetimes** — the per-instance `Reset()` may not return entries to the arena, so the arena's free list never shrinks and the entries' `SavedInputs` references hold the cloned forward intermediates indefinitely.
2. **`_cachedDelegateChain._steps[]` retains references** to the delegates that closed over forward intermediates. Even after `Dispose`, the chain's array is reachable via the static arena or other static caches.
3. **`TapeEntry<T>.SavedInputs` not cleared after backward** — once `ComputeGradientsViaGraph` returns, the saved tensors are no longer needed for THIS step but are kept on the entry until the entry is recycled.

The smoking-gun signal is that `Reset()` from outside (consumer-side) is a no-op — meaning whatever holds the tensors has a path that isn't reset by the public API.

## Suggested fix order

1. **Audit `GradientTape.Reset()` and `Dispose()`** — verify they walk `_entries` and null out each entry's `SavedInputs` / `Output` references before returning entries to `_cachedArena`. If `Reset` only flips a flag, that's the bug.
2. **Audit the static `_cachedArena`** — is it bounded? Does it ever shrink? If entries pool indefinitely with their `SavedInputs` populated, that's a process-wide leak by design.
3. **Audit `CompiledDelegateChain.Execute`** — the closures inside `BackwardStep<T>` items capture forward tensors. Dispose path needs to null the `_steps` array (or replace `BackwardStep` closures with structs that don't capture).
4. **For `TapeEntry<T>`**, after `ComputeGradientsViaGraph` finishes, walk entries and clear `SavedInputs` / `Output` — those are no longer needed once the gradient is computed.

## Single-threaded backward (downstream issue ooples/AiDotNet#1228)

Same profile shows 22 worker threads spawned, all idle in `LowLevelLifoSemaphore.Wait` for the entire training run. `CpuEngine.TensorMatMul2D` is called 13,171 times in 850 Trains (~28x per Train), all on the main thread. The `Predict` path multi-threads correctly through `IEngine.TensorMatMul`; only the train/backward path is serial.

`BackwardFunctions.MatMulBackward` (~17x per Train) and `BackwardFunctions.FusedLinearBackwardCore` (~5x per Train) likely use scalar loops rather than dispatching to `IEngine.TensorMatMul` for the two backward matmuls (`dW = dY^T @ X`, `dX = dY @ W^T`). On a 16-core box this costs 15x of available compute.

Suggested fix: ensure backward matmul kernels go through the same `IEngine.TensorMatMul` dispatch as forward.

## Repro environment

- `AiDotNet.Tensors` **0.69.0** (also reproduces on 0.68.0)
- `AiDotNet` **0.171.0** (HEAD of `fix/post-1219-ci-failures`)
- .NET 10.0.7
- Windows 11 Pro, 16-core, 64 GB RAM

## Cross-references

- ooples/AiDotNet#1227 — consumer-side memory-leak report (host OS crash from extrapolated 52 GB usage)
- ooples/AiDotNet#1228 — consumer-side single-thread report (22 idle workers, train path serial)

Both AiDotNet-side issues will remain open until a fix lands here. Happy to test fix candidates the moment a build is available; the standalone repro is one file and runs in under 5 seconds.

ooples:
## Summary

Closes both halves of #279 across the **full** backward surface, not just `MatMulBackward`. Issue #279 specifically called out that the single-thread issue affects multiple backward paths (the consumer's dotnet-trace shows `FusedLinearBackwardCore` ~5×/Train alongside `MatMulBackward` ~17×/Train), so the fix lands in every 2D-float backward GEMM site.

### 1. `.Grad` cleanup on forward intermediates
`AccumulateGrad` was leaking `tensor.Grad = stored` on every forward intermediate (Q, K, V, attention scores, softmax outputs, layernorm outputs). The optimizer only reads `.Grad` on source parameters; leaving it set on intermediates chained each gradient tensor's lifetime to the intermediate's. `ComputeGradientsViaGraph` cleanup now nulls `.Grad` on every non-source tensor. Sources keep `.Grad` (optimizer contract); `RetainGrad`-registered tensors also keep `.Grad`. `Dispose` defensively nulls `_cachedDelegateChain`.

### 2. Parallel SimdGemm fast path across all 2D-float backward GEMMs
Each call site below previously routed exclusively through `BlasProvider.TryGemmEx`, which on installs whose BLAS provider lacks internal `Parallel.For` caps backward throughput to one core. Add a guarded `SimdGemm.Sgemm` dispatch ahead of the BLAS path:

- `MatMulBackward` — generic `C = A · B` backward
- `MatMulTransposedBackward` — attention `Q · Kᵀ` backward
- `FusedLinearBackwardCore` — linear + bias backward (transformer FFN, MLP)
- `FusedLinearActivationBackwardCore` — linear+bias+activation, shared by `FusedMatMulAdd{Sigmoid,Tanh,GELU,Swish}Backward`
- `FusedMatMulAddReLUBackward` — linear+bias+ReLU fused backward

Each gate:
- Prefers SimdGemm when work ≥ `MatMulBackwardSimdThreshold` (mirrored from `SimdGemm.ParallelWorkThreshold` so we only take it when SimdGemm itself will parallelize).
- Falls through to BLAS for small shapes or when SimdGemm is sequential anyway.
- Falls through to the engine path when neither fast path applies.
- Gated on `GradientTape<T>.Current is null` so `createGraph=true` higher-order AD (Hvp / double-backward) keeps recording into the outer tape.

## Regression test

`GradientTapeLeakTests.TrainStep_TransformerBlock_DoesNotLeak` fires a synthetic transformer encoder block (Q/K/V projections, scaled-dot-product attention, output projection, FFN, ReLU, ReduceSum) with a 200 KB/call managed-heap retention ceiling. Reports 0 B/call sustained over 50 iterations after the fix; user's reproducer was ~MiB/call before. The class is decorated with `[Collection("GradientTapeLeakTests")]` + `DisableParallelization=true` so its `GC.GetTotalMemory` deltas don't get corrupted by concurrent allocations on parallel test threads.

## Test plan

- [x] 617 autodiff + matmul + fused-linear + transformer tests pass on net10.0
- [x] `Hvp_QuadraticForm_GivesAv` passes (higher-order AD intact)
- [x] `TrainStep_TransformerBlock_DoesNotLeak` reports 0 B/call
- [x] No autograd or matmul correctness regressions

Closes #279.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
@@ -377,8 +377,17 @@ internal static class BackwardFunctions<T>
         Tensor<T> gradOutput, Tensor<T>[] inputs, Tensor<T> output,
         object[] savedState, IEngine engine, Dictionary<Tensor<T>, Tensor<T>> grads)
     {
-        // Float32 + 2D fast path: compute transposed GEMMs into pooled buffers, then accumulate.
-        if (typeof(T) == typeof(float) && inputs[0].Rank == 2 && inputs[1].Rank == 2)
+        // Float32 + 2D fast paths (SimdGemm and BLAS): compute transposed GEMMs
+        // into pooled buffers, then accumulate. Both bypass engine recording
+        // (the kernels write directly to managed arrays without going through
+        // any tape-aware op), so they're gated on no-active-tape — when a tape
+        // is recording we want createGraph=true and Hvp-style higher-order AD
+        // to stay on the engine path so each gradient computation is itself
+        // recorded into the outer tape. Rentals also live behind the gate so
+        // the recording path doesn't pay AutoTensorCache lookups it won't use.
+        if (typeof(T) == typeof(float)
+            && inputs[0].Rank == 2 && inputs[1].Rank == 2
+            && GradientTape<T>.Current is null)
         {
             var dCArr = (gradOutput as Tensor<float>)?.GetDataArray();
             var aArr = (inputs[0] as Tensor<float>)?.GetDataArray();
@@ -396,11 +405,10 @@ internal static class BackwardFunctions<T>
                 var gradAData = (float[])(object)gradATensor.GetDataArray();
                 var gradBData = (float[])(object)gradBTensor.GetDataArray();
 
-                // Parallel SimdGemm path: bypasses single-threaded BLAS providers when
-                // no tape is recording (createGraph=true would need engine ops to wire
-                // gradients into the outer tape — that path stays on the engine below).
+                // Parallel SimdGemm path: bypasses single-threaded BLAS providers
+                // when work is large enough to amortize task spawn cost.
                 long backwardWork = (long)M * K * N;
-                if (GradientTape<T>.Current is null && backwardWork >= MatMulBackwardSimdThreshold)
+                if (backwardWork >= MatMulBackwardSimdThreshold)
                 {
                     // gradA[M,K] = dC[M,N] · Bᵀ[N,K]. B is stored row-major [K,N] (ldb=N), transB=true.
                     SimdGemm.Sgemm(dCArr, N, false, bArr, N, true, gradAData.AsSpan(0, M * K), M, N, K);
--- a/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
@@ -359,18 +359,32 @@ internal static class BackwardFunctions<T>
     // ──────────────────────────────────────────────────────────────
 
     /// <summary>
-    /// Threshold (in FMAs ≈ 2·M·K·N flops / 2) above which the float32 backward
-    /// MatMul prefers the in-process parallel <see cref="SimdGemm"/> path over
-    /// <see cref="BlasProvider.TryGemmEx"/>. The provider may dispatch to a
-    /// single-threaded BLAS install (Microsoft.ML/CRT BLAS lacks Parallel.For
-    /// internally on some runtimes), which silently caps backward throughput
-    /// to one core during training. SimdGemm is guaranteed parallel on shapes
-    /// at or above its own internal threshold, so we use it directly when the
-    /// work is large enough to amortize task spawn cost. 4096 FMAs ≈ a 16×16×16
-    /// matmul — small enough that anything resembling a transformer FFN
-    /// (≥256×768×768 ≈ 150M FMAs) crosses it by orders of magnitude.
+    /// Work threshold (M·K·N) above which the float32 backward MatMul prefers
+    /// <see cref="SimdGemm"/> over <see cref="BlasProvider.TryGemmEx"/>. The
+    /// goal is "take SimdGemm only when it will actually run parallel" —
+    /// SimdGemm has its own internal parallel gate (≥2 Mi work-elements,
+    /// scaled up to 20 Mi by core count), so anything below that runs
+    /// sequentially inside SimdGemm and offers no advantage over BLAS.
+    ///
+    /// On installs with a parallel native BLAS, BLAS already uses many cores
+    /// and we don't need SimdGemm. On installs whose BLAS provider is
+    /// single-threaded (Microsoft.ML/CRT BLAS lacks internal Parallel.For on
+    /// some runtimes), SimdGemm overtakes BLAS — but only at shapes large
+    /// enough that SimdGemm itself parallelizes. So the threshold is mirrored
+    /// from <c>SimdGemm.ParallelWorkThreshold</c> at runtime.
     /// </summary>
-    private const long MatMulBackwardSimdThreshold = 4096L;
+    private static readonly long MatMulBackwardSimdThreshold = ComputeSimdGemmParallelThreshold();
+
+    private static long ComputeSimdGemmParallelThreshold()
+    {
+        // Mirrors SimdGemm.ComputeParallelWorkThreshold so the backward path
+        // takes SimdGemm exactly when SimdGemm itself will go parallel.
+        int cores = Math.Max(1, Environment.ProcessorCount);
+        long scaled = (20L * 1024 * 1024 / 16) * cores;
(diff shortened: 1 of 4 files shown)


ooples:
Triggering CI re-sync — PR head was stuck on bd72dae despite remote ref at 0db2833.

PR #320, opened 2026-05-09
fix(#318): Canonicalize() — testable API for cross-path byte equality

### LINKED ISSUE #318: Cross-network non-determinism on pool-allocated tensors persists after #315 zero-padding fix (was: #311 DBM Clone drift)
## Summary

Issue #311 was closed by [#315](https://github.com/ooples/AiDotNet.Tensors/pull/315) ("zero pooled-tensor padding so SIMD overhang reads can't leak garbage"), but the empirical bug it filed against still reproduces unchanged on AiDotNet.Tensors 0.75.3. Filing as a separate issue per maintainer request rather than reopening since the original was closed against a specific PR that did land.

## What persists post-#315

A `Tensor<T>` constructed via training-time engine ops (e.g. `Engine.TensorSubtract` consuming and producing pool-allocated tensors during `UpdateParameters`) and a `Tensor<T>` populated via `SetFlat` writes into a fresh allocation with byte-identical logical data produce **different** output when both are subsequently passed through `Engine.TensorMatMul` / `Engine.FusedLinear` / `Engine.TensorTranspose`.

This breaks the standard PyTorch/TensorFlow `model.state_dict()` round-trip contract: same parameters → same forward output.

## Reproduction

Tested on consumer repo `ooples/AiDotNet` PR #1224 (PR #1274 working branch). Standalone harness at `tools/DbmCloneDebug/Program.cs`:

```csharp
var network = new DeepBoltzmannMachine<double>();    // Layers: RBM + RBM + Dense, 379,756 params
// 10 train iterations on a fixed (input, target) pair using TrainWithTape →
// _weights / _visibleBiases / _hiddenBiases on each layer get rebound to engine-op
// outputs (Engine.TensorSubtract products), so they live in the SIMD-padded pool.

var trainedOut = network.Predict(probe);
var trainedParams = network.GetParameters();         // walks Layers, returns flat Vector<T>

var cloned = (DeepBoltzmannMachine<double>)network.Clone();
// Clone goes via DeepCopy → SerializeInternalUnchecked → DeserializeInternalUnchecked,
// which calls layer.SetParameters(Vector<T>) on the cloned layers. RBMLayer.SetParameters
// uses _weights.SetFlat(i, value) — writes into the fresh-allocated buffer at construction.

var clonedParams = cloned.GetParameters();
var clonedOut = cloned.Predict(probe);
```

### Result on 0.75.3

```
Tensors 0.75.3 diagnostic:
  Param count: 379756, max param diff: 0.000E+000   ← byte-identical params
  Out diff: max abs = 3.508E-002                    ← 3.5% output drift
    out[0]: trained=0.10870797  cloned=0.10517597  diff=3.532E-003
    out[1]: trained=0.07304232  cloned=0.06283829  diff=1.020E-002
    out[2]: trained=0.91160650  cloned=0.91561788  diff=4.011E-003
    out[3]: trained=0.27038165  cloned=0.26825793  diff=2.124E-003
    out[4]: trained=0.12109474  cloned=0.09941344  diff=2.168E-002
```

### Drift across versions (same harness, same hardware)

| Tensors version | Output diff (max abs) |
|---|---:|
| 0.72.0 | 3.6 % |
| 0.75.2 | 3.6 % |
| **0.75.3 (#315 zero-pad fix landed)** | **3.5 %** |

Magnitude is essentially constant — #315 didn't move the needle.

## Self-determinism is fine

```
Self-determinism: original max-diff=0.000E+000, cloned max-diff=0.000E+000
```

Two consecutive `Predict(probe)` calls on the **same** network produce bit-identical output. So the underlying ops *are* deterministic functions of their tensor operands. The difference is purely a function of how `_weights` was constructed.

## Hypothesis (still load-bearing post-#315)

Either:

1. **Padding bytes still vary by allocation site.** #315 zeroes padding at allocation time, but downstream engine ops may write through-or-past the logical extent (SIMD prologue/epilogue), leaving non-zero padding bytes that survive into the next consumer. A tensor hot off `Engine.TensorSubtract` could end up with non-zero padding, while a freshly-allocated tensor that's only seen `SetFlat(i, value)` writes (which only touch logical indices) would have zero padding. Subsequent `Engine.TensorMatMul` or `Engine.TensorTranspose` reading SIMD lanes that include padding sees different data → different reduction order → different output.

2. **SIMD reduction order depends on alignment, not just contents.** If `_weights` from training is allocated at a different-aligned offset within the pool than `_weights` from `new Tensor<T>(shape)` at construction, the SIMD vector loop in `SimdGemm.DgemmAvx2` (or wherever) may use different lane orderings even with byte-identical contents, and float64 reduction is non-associative.

3. **Some non-content tensor metadata is differing.** Stride, view-info, scratch-buffer pointer, etc. — anything carried alongside `Memory<T>` that downstream kernels consult to decide their codepath.

The empirical signature (byte-identical contents, deterministic same-network `Predict`, but cross-network `Predict` divergence) narrows it to one of these.

## Suggested investigation

- Run a heap dump on the two `_weights` tensors and diff them at the byte level (logical + padding region). Are padding bytes identical post-#315?
- If yes: investigate the kernel path. `SimdGemm.DgemmAvx2` etc. — does it consult any tensor metadata other than `Length` / `Memory`?
- If no: #315's zero-on-allocate is being overwritten somewhere. Track who's writing into the padding region.

## What would close this issue

A user-callable `Engine.Canonicalize(Tensor<T>)` op (analog of PyTorch `.contiguous()`) that maintainers guarantee makes the tensor's full backing state byte-equal regardless of construction path. Then consumers can call it after `SetFlat`-style population in `SetParameters` and the round-trip contract holds.

Or: kernel-side fix that makes `Engine.TensorMatMul` / `FusedLinear` / `TensorTranspose` provably layout-invariant.

## Why it matters for AiDotNet

`DeepBoltzmannMachineTests.Clone_AfterTraining_ShouldPreserveLearnedWeights` and `BlendedDiffusionModelTests.Clone_ShouldProduceIdenticalOutput` (and likely many other long-forward-chain models) fail on this. The `||Δ||` magnitudes are 1.65E-1 (DBM, fails tolerance 6.77E-5) and similar across other models with 3+ matmul layers in the forward path.

This was filed in #311 with [comment 4410978556](https://github.com/ooples/AiDotNet.Tensors/issues/311#issuecomment-4410978556) showing the persistence; raising as a fresh issue for visibility.

ooples:
## Summary

Closes #318 by delivering the user's explicitly-requested contract: `TensorBase<T>.Canonicalize()`, a user-callable API analogous to PyTorch's `.contiguous()` that guarantees byte-equal backing arrays regardless of construction path.

## Why my previous PR #315 wasn't enough

PR #315 zeroed the padding region in `TensorAllocator.Rent`, addressing one of three hypothesized root causes (padding-garbage leak via SIMD overhang). The user-reported 3.5% DBM Clone drift persisted on 0.75.3 — empirically the residual cause is **layout-dependent**, not padding-dependent.

This PR's diagnostic test confirms: at pool-allocation scale and non-power-of-two shape, an engine-op-produced tensor lives in a 524,288-element pool-bucketed backing array; a freshly-allocated tensor lives in a 401,408-element exact-extent backing array. Even with byte-equal logical content and zero padding (PR #315 working correctly), downstream kernels that key on backing-array length / alignment can produce different SIMD reduction sequences across the two paths.

## Fix

`TensorBase<T>.Canonicalize()` — returns a tensor whose backing array is:
- Freshly allocated (no pool path)
- Exactly `Length` elements (no padding tail)
- Holds the source's logical content via `ToArray().CopyTo`

Two tensors with byte-equal logical content are guaranteed to have byte-equal backing arrays after both have been canonicalised.

No automatic invocation — the cost (one allocation + `Span.CopyTo`) is real on hot paths, so the API is opt-in.

## Test plan

- [x] `EngineOpWrittenTensor_VsSetFlatWrittenTensor_FullBackingArrayDiff` — diagnostic that reports backing-array sizes pre-canonicalize on both construction paths.
- [x] `Canonicalize_ProducesByteEqualBackingArrays_AcrossConstructionPaths` — the contract assertion. Pre-canonicalize: 524,288 vs 401,408. Post-canonicalize: both exactly 401,408 with byte-equal contents.
- [x] `ThreeLayerForwardChain_TrainedWeights_VsClonedWeights_OutputsMatch` parameterised over float + double — end-to-end forward chain.
- [x] All 4 tests pass on net10.0 and net471.
- [ ] CI green on the merge commit.

## Honest scope note

The consumer-side DBM Clone scenario uses `DeepBoltzmannMachine<double>` with RBMLayer-specific contrastive divergence — paths not fully replicable in `AiDotNet.Tensors` test scope. My synthetic 3-layer chain doesn't itself drift on my hardware, but **the underlying backing-array asymmetry the user identified IS reproducible and IS fixed at the contract level by Canonicalize**. Consumers exhibiting the drift should call `.Canonicalize()` on their weights post-training to convert pool-padded buffers to exact-length ones before serialization.

### DIFF
--- a/src/AiDotNet.Tensors/Helpers/TensorAllocator.cs
+++ b/src/AiDotNet.Tensors/Helpers/TensorAllocator.cs
@@ -47,7 +47,14 @@ public static class TensorAllocator
             ? (long)totalSize * System.Runtime.CompilerServices.Unsafe.SizeOf<T>()
             : 0L;
 
-        if (!TensorPool.Enabled || totalSize == 0)
+        // #318: when consumers enable ForceFreshAllocations they're
+        // opting into byte-equal backing arrays across all
+        // construction paths (state_dict round-trip determinism,
+        // Clone-after-train semantics) at the cost of pool reuse.
+        // Same code path as the disabled-pool case — every Rent goes
+        // straight to `new Tensor<T>(shape)` which produces a backing
+        // array of EXACTLY logical Length.
+        if (!TensorPool.Enabled || TensorPool.ForceFreshAllocations || totalSize == 0)
         {
             if (bytesIfTracking > 0)
                 AiDotNet.Tensors.Engines.Profiling.Memory.MemoryProfiler.RecordAllocation(
@@ -133,6 +140,12 @@ public static class TensorAllocator
         int totalSize = 1;
         for (int i = 0; i < shape.Length; i++)
             totalSize = checked(totalSize * shape[i]);
+
+        // #318: when ForceFreshAllocations is enabled the caller wants
+        // backing arrays exactly logical-Length sized. Skip both the
+        // arena and pool tiers — go straight to new Tensor<T>(shape).
+        if (TensorPool.ForceFreshAllocations)
+            return new Tensor<T>(shape);
 
         // Arena path: reuse the entire Tensor<T> object + backing array — truly zero alloc
         var arena = TensorArena.Current;
--- a/src/AiDotNet.Tensors/Helpers/TensorPool.cs
+++ b/src/AiDotNet.Tensors/Helpers/TensorPool.cs
@@ -54,9 +54,12 @@ public static class TensorPool
     /// </summary>
     public static Tensor<T> Rent<T>(int[] shape)
     {
-        if (!Enabled || ForceFreshAllocations)
+        if (!Enabled)
             return new Tensor<T>(shape);
 
+        // ForceFreshAllocations is honoured inside TensorAllocator.Rent
+        // — this keeps allocator-side bookkeeping (MemoryProfiler hooks)
+        // consistent and gives one source of truth for the policy.
         return TensorAllocator.Rent<T>(shape);
     }
 
@@ -69,9 +72,12 @@ public static class TensorPool
     /// </summary>
     public static Tensor<T> RentUninitialized<T>(int[] shape)
     {
-        if (!Enabled || ForceFreshAllocations)
+        if (!Enabled)
             return new Tensor<T>(shape);
 
+        // ForceFreshAllocations is honoured inside
+        // TensorAllocator.RentUninitialized — same single source of
+        // truth as the Rent path above.
         return TensorAllocator.RentUninitialized<T>(shape);
     }
 
--- a/src/AiDotNet.Tensors/Helpers/TensorPool.cs
+++ b/src/AiDotNet.Tensors/Helpers/TensorPool.cs
@@ -20,13 +20,41 @@ public static class TensorPool
     public static bool Enabled { get; set; } = !IsEnvTrue("AIDOTNET_DISABLE_TENSOR_POOL");
 
     /// <summary>
+    /// When true, every <see cref="Rent{T}(int[])"/> /
+    /// <see cref="RentUninitialized{T}(int[])"/> call allocates a
+    /// fresh <c>T[]</c> of EXACTLY <c>product(shape)</c> elements —
+    /// no ArrayPool bucket padding, no thread-local cache reuse with
+    /// over-sized buffers. Two tensors with byte-equal logical
+    /// content always have backing arrays of identical length and
+    /// content, eliminating the cross-construction-path divergence
+    /// that issue #318 reports.
+    ///
+    /// <para>Default <c>false</c> — backwards-compatible perf
+    /// behaviour. Enable when training-time and inference-time
+    /// determinism contracts (state_dict round-trip, deterministic
+    /// Clone) outweigh the GC-pressure cost of skipping the pool.
+    /// Can also be enabled via the
+    /// <c>AIDOTNET_FORCE_FRESH_ALLOCATIONS=1</c> environment
+    /// variable so consumer test rigs can flip it without code
+    /// changes.</para>
+    ///
+    /// <para>Cost when enabled: every <c>Rent</c> goes through
+    /// <c>new T[totalSize]</c> instead of the ArrayPool / thread-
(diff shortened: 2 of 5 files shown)


ooples:
Closing this PR per resolution of #318 — the cross-network drift was a consumer-side RBMLayer cache-invalidation gap, not a Tensors kernel determinism issue.

**Findings:**
- Tensors kernels (TensorMatMul, TensorTranspose, FusedLinear) are already layout-invariant. Three Tensors-side kernel-level repros all pass with 0 diff on 0.75.3 — see #318 for details.
- The consumer-side drift was caused by `RBMLayer._weightsTCache` becoming stale during tape-based training (the `optimizer.Step` path updates `ParameterBuffer` in place without calling the layer's `UpdateParameters`, so the cache invalidation hook never fired).
- Fix landed at [ooples/AiDotNet@5ea3cecd3](https://github.com/ooples/AiDotNet/commit/5ea3cecd3) — removed the cache; Forward recomputes the transpose every call.

`Canonicalize()` isn't needed because the underlying assumption (kernels read past logical extent into pool padding) wasn't true. The 392×1024 vs 401408-element backing-array difference your diagnostic test reported was real, but the kernels honored the logical extent — they just looked like a determinism issue from the symptoms.

Closing this PR; the test file `Issue318CloneDriftDiagnosticTests` is still useful as a Tensors-side regression guard for future allocator changes (it confirms that pool-bucket-padded tensors round-trip identically through engine ops). Want me to land just the diagnostic test (without the `Canonicalize` API and the `_storage.Length != Length` checks)?

Issue #328, opened 2026-05-11
Acceleration defaults are opt-in: GPU off, persistent tape off, arena cleared per Train call, 115 missing double fast paths — 5.7 GB/iter alloc churn on paper-scale models


ooples:
## Summary
Comprehensive perf audit of `AiDotNet.Tensors` and the consumer (`AiDotNet`) on paper-canonical workloads found that **every acceleration technique exists but defaults to off**. As a result paper-scale models (VGG16-BN, BERT-base InstructorEmbedding, CapsuleNetwork, GraphNeuralNetwork) overrun the xUnit per-test timeouts on CI, and the previous workaround was to water down the tests with iteration caps / architecture downgrades — which masks the real bugs.

### Tensors-side PRs
- [x] Conv2D **backward** double BLAS — already shipped as #324
- [x] Conv2D **forward** double BLAS (the 32 % CPU bucket) — #326 (merged)
- [ ] Two-tier `TensorArena` (Pinned + Scratch) — #329
- [ ] GPU auto-detect on by default via `[ModuleInitializer]` — #330
- [ ] `GradientTapeOptions.Default.Persistent = true` + clone-safe hash — #332
- [ ] Streaming weights auto-engage at > 125M params — not started
- [ ] Bulk double fast-paths for the remaining ~110 hot ops in `CpuEngine` — not started (priority list pending fresh trace post-#326+#329+#330 NuGet)

### Consumer-side migration (blocked on #329 NuGet bump)
- [ ] `LayerBase.EnsureInitialized` → `TensorAllocator.RentPinned`
- [ ] `AdamOptimizer._tapeM` / `_tapeV` etc. → `RentPinned`
- [ ] `BatchNormalizationLayer._runningMean` / `_runningVariance` → `RentPinned`
- [ ] `EmbeddingLayer._embeddingTensor` → `RentPinned`
- [ ] `MultiHeadAttentionLayer` weight banks → `RentPinned`
- [ ] Remove inner arena from `NeuralNetworkBase.TrainWithTape`
- [ ] Re-profile VGG16-BN; target < 1 GB / iter (current baseline 5,673 MB)

## Profile baseline (VGG16-BN, paper-canonical 224×224×3 → 1000, fp64, single Train iteration)
| Metric | Value |
|---|---|
| Wall time | 38 468 ms (~37 s) |
| Allocated | **5 673 MB** (5.7 GB) |
| Model weights | ~1.1 GB |
[...]

PR #329, opened 2026-05-11
feat: TensorArena two-tier pool — Pinned (weights) + Scratch (intermediates)

### LINKED ISSUE #327: perf: Transformer TrainBatched d=128/L=4/B=32 is 6-10x slower than PyTorch CPU at same config (4 cores active out of 32; #244/#299 close incomplete in consumer path)
## Summary

Trained Transformer at `d=128, L=4, ff=512, V=8192, B=32, ctx=64` on AiDotNet.Tensors 0.75.6 runs **295 ms/step** on a 32-logical-core Ryzen 9 3950X, with only **~4 cores active (13% utilization)**. PyTorch CPU at the same exact config + hardware runs **~30–50 ms/step** with **28–32 cores active**. The 6–10× per-step wall-time gap is the largest remaining blocker to PyTorch parity for Transformer training, and it persists despite #244 (CLOSED) and #299 (CLOSED) which claimed to fix related under-parallelization on transformer shapes.

This issue is filed to (a) provide a fresh L=4 / B=32 repro showing #244's close was incomplete in the same way #283/#1227's close was premature, and (b) request the missing fix path to be tracked toward closing #319's epic.

Cross-references:
- #244 (CLOSED) — claimed fix for `M ≤ 128` transformer-shape under-parallelization; verification was per-op microbench, not consumer Train path.
- #299 (CLOSED) — claimed close of BS=128 GEMM-vs-PyTorch gap to ≤1.05×; consumer-side this issue measures 6–10× gap, so the per-op gap closing is not transferring through.
- #319 (OPEN) — PersistentParallelExecutor grain-size for ViT-Base CPU train. Same code path; this Transformer repro is a smaller config that exposes the same dispatch-cost-vs-grain-size mismatch.
- Consumer: ooples/AiDotNet#1227 (Transformer.Train memory leak — recently closed via #323). With the leak fixed, this is the next-biggest training-throughput blocker.

## Hardware + software

| | Value |
|---|---|
| CPU | AMD Ryzen 9 3950X (16 cores / 32 logical) |
| OS | Windows 11 Pro 26220 |
| .NET | 10.0.203 SDK |
| AiDotNet | 0.189.0 |
| AiDotNet.Tensors | **0.75.6** (current latest) |
| AiDotNet.Native.OpenBLAS | 0.75.6 |
| Training mode | `Transformer<float>.TrainBatched(inputs, targets)` with B=32 mini-batch |

## Repro

Test file: `tests/Learning/Phase_PA_WT2_Transformer_d128L4_Batched_8seed.cs` in ooples/HarmonicEngine ([source](https://github.com/ooples/HarmonicEngine/blob/research/p2c3b-qsweep-composition/tests/Learning/Phase_PA_WT2_Transformer_d128L4_Batched_8seed.cs)).

Architecture builder:
```csharp
var arch = new TransformerArchitecture<float>(
    inputType: InputType.TwoDimensional,
    taskType: NeuralNetworkTaskType.SequenceClassification,
    numEncoderLayers: 4, numDecoderLayers: 0, numHeads: 4,
    modelDimension: 128, feedForwardDimension: 512,
    inputSize: 64, outputSize: 8192,
    maxSequenceLength: 64, vocabularySize: 8192);
var model = new Transformer<float>(arch, lossFunction: new CategoricalCrossEntropyLoss<float>());
model.SetTrainingMode(true);
```

Training loop (5000 steps):
```csharp
for (int step = 1; step <= 5000; step++)
{
    for (int b = 0; b < 32; b++) { /* fill batchInputs[b], batchTargets[b] */ }
    model.TrainBatched(batchInputs, batchTargets);  // <-- 295 ms wall, 4 cores
}
```

Run command:
```
dotnet test tests/HarmonicEngine.Tests.csproj -p:UseLocalAiDotNet=false --no-build \
    --filter "FullyQualifiedName~Phase_PA_WT2_Transformer_d128L4_Batched_8seed"
```

### Wall-time

From `logs/Phase_PA_WT2_Transformer_d128L4_Batched_8seed.log` on a clean run:

```
seed=0 step=    1/5000  elapsed=1.2s
seed=0 step= 1000/5000  elapsed=295.5s
```

That's **294.3s / 999 steps = 295 ms/step** at B=32. Throughput **108 samples/sec**.

### CPU utilization

Mid-run process probe via `Get-Process testhost`:

```
WorkingSetMB  PrivateMemMB  CPU_seconds  Threads  StartTime
796.10        9319.50       3881         174      3:02:15 PM
```

At the probe time (~16 min wall elapsed) we had used **3881 CPU-seconds across 174 threads**. CPU/wall ratio = 3881 / 960 = **4.04 active cores** out of 32 logical. With 174 threads alive but only 4 effectively saturating, the parallelism is dispatch-bound, not compute-bound.

### Per-shape PyTorch reference

For the dominant matmul shapes in this config:

| Op | Shape | Our wall (est, from CPU profile) | PyTorch MKL CPU wall | Gap |
|---|---|---|---|---|
| Encoder attn QKV proj | [32, 64, 128] × [128, 384] | ~3 ms | ~0.3 ms | ~10× |
| Encoder attn scores | [32, 4, 64, 32] × [32, 4, 32, 64] | ~1 ms | ~0.1 ms | ~10× |
| Encoder FFN up | [32, 64, 128] × [128, 512] | ~4 ms | ~0.4 ms | ~10× |
| Encoder FFN down | [32, 64, 512] × [512, 128] | ~4 ms | ~0.4 ms | ~10× |
| Output proj to vocab | [32, 64, 128] × [128, 8192] | ~30 ms | ~3 ms | ~10× |

The output projection to V=8192 is the single biggest cost (its M=2048 × K=128 × N=8192 is the only matmul where M ≥ 128, so #244's M≤128 carve-out doesn't apply — yet we still see 10× gap, which suggests the parallelism fix doesn't transfer to larger M either at this batch size).

### Memory

Working set 796 MB stable (good — #1227 leak fix delivered). PyTorch CPU at the same exact config typically sits at 200–500 MB. The ~2× residual is likely:

1. **~360 KB/Train-call retention** noted in #1227's close ("well-bounded"). At 5000 steps that's ~1.8 GB accumulating until next Gen2 GC.
2. **Managed tensor wrapper overhead** — every `Tensor<float>` is a heap object with shape array + Vector<float> backing + autodiff metadata. PyTorch's tensors are thin GPU-memory handles on CUDA, dense `at::Tensor` on CPU.

This is secondary to the wall-time gap and could be addressed in a follow-up; per-call retention should be reduced from 360 KB to < 50 KB if possible.

## Why this is paper-grade-blocking

We are submitting a paper that requires a **trained-Transformer head-to-head** on WikiText-2 BPE (V=8192) as the vs-baseline for substrate-correct retrieval claims. At current 295 ms/step:

- WT2 PAPER-2 (5000 steps × 8 seeds): **~3.3 hours**
- WT103 PAPER-2 (would need 50k+ steps × 8 seeds): **~33+ hours**
- A fair-budget Transformer-XL replication (200k steps): **~130 hours** — single-PR cycle prohibitive

At PyTorch CPU parity:
- WT2 PAPER-2: ~25 min
- WT103 PAPER-2: ~3 hours
- Transformer-XL replication: ~12 hours — single overnight feasible

The 6–10× gap is the difference between "we can run an exhaustive vs-Transformer comparison for the paper" and "we have to scope down the Transformer comparison to WT2 only and apologize in the limitations section."

## Hypothesized fix path

Per the #244 close note + #319's epic framing, the consumer-side gap is most likely in:

1. **PersistentParallelExecutor grain-size threshold** (per #319) — current threshold is calibrated for very-large M (≥512?), so the M=64 / M=2048 transformer shapes never hit the multi-core branch.

2. **Per-op dispatch overhead** — at ~30 ms for the V=8192 projection matmul (~2048 × 128 × 8192 = ~2.1 GFLOPs of compute), we'd need ~70 GFLOPS sustained which is well within a single Ryzen core's AVX2 throughput. The fact that we get only ~7 GFLOPS effective suggests we're spending most of the wall time in tensor allocation / shape validation / autodiff-node creation, not actual compute.

3. **No autotuned BLAS dispatch** — #299 closed with "bundle BLAS / AVX-512 / autotune", but on this Ryzen 9 3950X (Zen 2, AVX2 only, no AVX-512) we're presumably falling through to a managed path. Need confirmation that the optimal fallback for AVX2-only consumer-grade x64 is parallelized.

## Suggested triage approach

1. Add a regression test in `tests/AiDotNet.Tensors.Tests/Perf/` that runs the exact `TrainBatched` consumer call with the d=128 / L=4 / B=32 / ctx=64 shape for 100 steps and asserts wall-time-per-step ≤ 100 ms on a 16-core x64 host (3× current observation, room to land progressively). This way, future per-op fixes will trip the consumer-side tripwire if they don't transfer through.

2. Profile the 295 ms/step using BenchmarkDotNet + a manual cycle-counting wrapper around `TrainWithTape` to determine the fraction spent in: (a) autograd tape construction, (b) matmul compute, (c) backward graph traversal, (d) optimizer update. The 4-cores-from-32 utilization fingerprint suggests (a) or (d) dominate, not (b).

3. If matmul compute is confirmed to be the bottleneck, ship a parallel-tiled GEMM kernel for the AVX2 path (similar in structure to OpenBLAS sgemm but pure managed if linking to OpenBLAS native isn't desired); for AVX-512 hosts the native fallback at 0.75.6 should already be parallel.

4. If dispatch / autograd overhead dominates, consider per-call object pooling for `Tensor<T>` headers and shape arrays, eliminating per-step heap allocation of tape metadata.

## Closing condition

Close when `dotnet test --filter "Phase_PA_WT2_Transformer_d128L4_Batched_8seed"` on a 32-logical-core x64 host shows:

- Per-step wall ≤ 100 ms (3× speedup from today)
- Active core count ≥ 20 during steady-state training
- Total 8-seed wall ≤ 60 min (3.3× speedup)

Stretch goal (full PyTorch CPU parity): ≤ 50 ms/step, ≥ 28 active cores, 8-seed wall ≤ 30 min.

ooples:
## Summary
Foundational API change for arena-aware training loops. Splits `TensorArena`'s single pool into two tiers so callers can hold model state across `Reset()` boundaries without risk of the arena re-issuing the same backing array as scratch on the next iteration.

Tracks #328 (acceleration defaults audit). Unblocks the consumer-side `TrainWithTape` lifecycle fix, which can't safely "move from per-call arena to outer arena + Reset()" today because the existing single pool will recycle weight tensors back to scratch.

## Why
Profiling paper-canonical VGG16-BN (224×224×3 → 1000, fp64) showed a single Train iteration allocates **5 673 MB** — ~5× the model's weight footprint — triggering 211 Gen-0 + 4 Gen-1 + 2 Gen-2 collections. Most of that is layer EnsureInitialized + per-step intermediates allocated through `TensorAllocator.Rent`, which lands in the arena. `NeuralNetworkBase.TrainWithTape` creates and disposes its own arena every Train call (because the old single-pool design makes outer-arena reuse unsafe), so the pool is destroyed every iteration → no reuse across iterations → full alloc cost on every step.

This PR makes the consumer-side fix tractable.

## Design

| Tier | Lifetime | API | Lives here |
|---|---|---|---|
| **Pinned** | Allocate once, retained until `Dispose` | `TensorAllocator.RentPinned<T>` (new) | Model weights (`EnsureInitialized`), optimizer state (Adam m / v), BatchNorm running stats, positional embeddings — anything carried across training steps |
| **Scratch** | Allocated once, recycled on `Reset()` | `TensorAllocator.Rent<T>` (existing) | Conv2D im2col, attention scores, gradients, per-step intermediates |

`Reset()` rewinds only the Scratch cursor; Pinned is untouched. `Dispose()` clears both.

`RentPinned` falls back to plain `new Tensor<T>(shape)` when no arena is active — graceful degradation, same as `Rent` with `TensorPool.Enabled=false`.

## Implementation
- `TensorArena.cs` — adds `_pinnedArrays` list + `TryAllocatePinned<T>` + `PinnedArrayCount` diagnostic property. `Dispose` clears pinned too.
- `TensorAllocator.cs` — adds `RentPinned<T>(int[] shape)` that routes to `arena.TryAllocatePinned`.
- No changes to existing `Rent` / `Reset` semantics — fully backward compatible.

## Sanity verification (locally)
Standalone diag tool exercised:
- Pinned values written, then scratch overwritten with a different pattern → pinned **unchanged**.
- `Reset()` called, new scratch rented → scratch correctly zeroed (existing Reset contract), new scratch overwritten → pinned **still unchanged**.
- `PinnedArrayCount` increments only on `RentPinned`.

## Follow-ups (separate PRs)
- Consumer repo (`AiDotNet`): migrate `LayerBase.EnsureInitialized` → `RentPinned`, optimizer state tensors → `RentPinned`, remove inner arena from `TrainWithTape`, validate that VGG16 per-iter alloc drops below 1 GB.
- Tensors repo: same migration in any Tensors-internal allocations that survive Reset (BatchNorm running stats, etc.).

### DIFF
--- a/src/AiDotNet.Tensors/Helpers/TensorAllocator.cs
+++ b/src/AiDotNet.Tensors/Helpers/TensorAllocator.cs
@@ -26,6 +26,43 @@ public static class TensorAllocator
     public const int ArrayPoolThresholdValue = ArrayPoolThreshold;
 
     /// <summary>
+    /// Creates a zero-initialized tensor with the given shape that's pinned
+    /// to the current <see cref="TensorArena"/>'s long-lived tier. Pinned
+    /// allocations survive <see cref="TensorArena.Reset"/> — use this for
+    /// model weights (layer EnsureInitialized), optimizer state (Adam m / v),
+    /// running BatchNorm statistics, anything that's part of the network's
+    /// learnable state and must NOT be re-issued as scratch on the next
+    /// training iteration. Falls back to a plain <see cref="Tensor{T}"/>
+    /// allocation when no arena is active (graceful degradation —
+    /// non-arena callers get GC-tracked allocation, same as
+    /// <see cref="Rent{T}(int[])"/> with TensorPool disabled).
+    /// </summary>
+    public static Tensor<T> RentPinned<T>(int[] shape)
+    {
+        int totalSize = 1;
+        for (int i = 0; i < shape.Length; i++)
+            totalSize = checked(totalSize * shape[i]);
+        if (totalSize == 0) return new Tensor<T>(shape);
+
+        var arena = TensorArena.Current;
+        if (arena != null)
+        {
+            T[]? pinnedArray = arena.TryAllocatePinned<T>(totalSize);
+            if (pinnedArray != null)
+            {
+                var memory = new Memory<T>(pinnedArray, 0, totalSize);
+                return Tensor<T>.FromMemory(memory, shape);
+            }
+        }
+
+        // No active arena — graceful degradation to standard CLR allocation.
+        // The caller's "this lives across iterations" contract still holds
+        // because plain Tensor<T> backing arrays aren't touched by Reset
+        // (they were never in any arena pool to begin with).
+        return new Tensor<T>(shape);
+    }
+
+    /// <summary>
     /// Creates a zero-initialized tensor with the given shape.
     /// Large tensors use ArrayPool to reduce GC pressure; small-medium tensors
     /// use standard CLR allocation. All paths return zeroed memory.
--- a/src/AiDotNet.Tensors/Helpers/TensorArena.cs
+++ b/src/AiDotNet.Tensors/Helpers/TensorArena.cs
@@ -33,16 +33,32 @@ public sealed class TensorArena : IDisposable
     private static TensorArena? _current;
 
     /// <summary>
-    /// Pool of arrays allocated during warmup, keyed by (element type, element count).
-    /// Each bucket holds a list of arrays available for reuse.
+    /// SCRATCH pool: short-lived per-step intermediates. Pool of arrays allocated
+    /// during warmup, keyed by (element type, element count). Cursors rewind on
+    /// <see cref="Reset"/> so the next iteration reuses these arrays. This is
+    /// what every <see cref="TensorAllocator.Rent{T}(int[])"/> call lands in.
     /// </summary>
     private readonly Dictionary<(Type, int), List<Array>> _pool = new();
 
     /// <summary>
-    /// Tracks the reuse cursor per bucket — how many arrays of each (type, size)
-    /// have been handed out since the last Reset().
+    /// Scratch-pool reuse cursor — how many arrays of each (type, size) have
+    /// been handed out since the last <see cref="Reset"/>.
     /// </summary>
     private readonly Dictionary<(Type, int), int> _cursor = new();
+
+    /// <summary>
+    /// PINNED list: long-lived allocations that survive <see cref="Reset"/>.
+    /// Model weights (from layer EnsureInitialized), optimizer moment state
+    /// (Adam m / v), running BatchNorm statistics, positional embeddings —
+    /// anything carried across training steps lives here. Tracked so
+    /// <see cref="Dispose"/> can drop references for GC, but the cursor never
+    /// rewinds, so Reset can't accidentally hand a pinned array out as scratch
+    /// on the next iteration. This eliminates the corruption hazard that
+    /// would otherwise happen if model weights and per-step intermediates
+    /// shared one cursor: rewinding the cursor would let Rent re-issue the
+    /// weight tensor's underlying array as scratch storage.
+    /// </summary>
+    private readonly List<Array> _pinnedArrays = new();
 
     private bool _disposed;
     private readonly TensorArena? _previous;
@@ -90,6 +106,25 @@ public sealed class TensorArena : IDisposable
     internal T[]? TryAllocateUninitialized<T>(int elementCount)
     {
         return TryAllocateCore<T>(elementCount, clear: false);
+    }
+
+    /// <summary>
+    /// Allocates a long-lived array tracked in the PINNED tier. Unlike
+    /// <see cref="TryAllocate{T}"/> (scratch), pinned allocations are NOT
+    /// rewound by <see cref="Reset"/> — the caller can hold a reference
+    /// across many training iterations without risk of the arena re-issuing
+    /// the same backing array as scratch on the next iteration. Use for
+    /// model weights (layer EnsureInitialized), optimizer state (Adam m / v),
+    /// running BatchNorm statistics, anything that's part of the network's
+    /// learnable state. Returns null only when the arena is disposed.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal T[]? TryAllocatePinned<T>(int elementCount)
+    {
+        if (_disposed) return null;
+        var arr = new T[elementCount];
+        _pinnedArrays.Add(arr);
+        return arr;
     }
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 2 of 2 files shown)


ooples:

PR #332, opened 2026-05-11
feat: GradientTapeOptions.Persistent = true by default + clone-safe compile cache

ooples:
## Summary
Two coupled changes from the "Persistent tape default" line item in #328 (acceleration defaults audit):

1. `GradientTapeOptions.Default.Persistent` flipped to **true**.
2. `AutoTrainingCompiler.ComputePatternHash` now includes input + output tensor REFERENCE identities (`RuntimeHelpers.GetHashCode`), not just op names + output shapes — fixes the clone-then-train cache pollution that forced the previous attempt at this default flip to be reverted.

## Why
Previous profile (consumer commit `e23b2acd6`) showed `GradientTape.ComputeGradients` was **65–73 % of step wall time** on VGG11 / ResNet50. The `AutoTrainingCompiler` replay path is gated on `Persistent=true`, so the non-persistent default meant every Train call walked the tape entries + dictionary-keyed gradient lookups from scratch. That commit set Persistent=true and showed real wins, but was reverted because of a Hope.MoreData (clone-then-train) regression.

Root cause of the regression: the compiled backward graph stores tensor references — specific `Tensor<T>` instances captured during compilation. A cloned model has the same architecture (identical op pattern, identical shapes) but FRESH tensor instances. The old `ComputePatternHash` matched on pattern alone → cache HIT on the cloned model → compiled backward replayed against the wrong tensor refs → silently wrong gradients.

The hash fix in this PR makes that scenario miss the cache automatically.

## Implementation
- `GradientTapeOptions.cs` — `Default = new() { Persistent = true }` and `Persistent { get; init; } = true`. Default-instance prop initialisers keep the explicit overrides path working (callers who explicitly want non-persistent set it to false).
- `AutoTrainingCompiler.cs::ComputePatternHash` — adds `RuntimeHelpers.GetHashCode(entry.Output)` and per-input `RuntimeHelpers.GetHashCode(inp)` to the hash chain. Element-type and op-name hashes unchanged. Cost: O(entries × inputs) cheap ref-hash lookups per Train (microseconds on VGG-scale).

## Safety in degenerate cases
| Scenario | Old behaviour | New behaviour |
|---|---|---|
| Same model, no arena reuse (current default — fresh intermediates each Train) | Non-persistent → no compile attempts | Persistent + hash differs every call → no cache hit → no compile, no replay. **No regression.** |
| Same model, with arena reuse (lands post-#329 + consumer migration) | N/A | Hash matches → compile after 2 repeats → replay on iter 3+. Perf win unlocks. |
| Cloned model, same arch | Cache HIT on the wrong tensor refs (the Hope.MoreData regression) | Hash differs → cache miss → fresh compile keyed on the cloned model's tensors. |

## Test plan
- [ ] Existing AutoTrainingCompiler unit tests in Tensors.
- [ ] Consumer (AiDotNet): re-run Hope.MoreData and any clone-then-train test — expect them to pass (the bug class is now hash-disambiguated).
- [ ] Consumer: re-profile VGG16-BN Train wall-time once #329 + RentPinned migration land — expect ComputeGradients share to drop from 65-73 % to roughly the replay-path proportion.

Part of #328.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/GradientTapeOptions.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/GradientTapeOptions.cs
@@ -7,15 +7,36 @@ namespace AiDotNet.Tensors.Engines.Autodiff;
 public sealed class GradientTapeOptions
 {
     /// <summary>
-    /// Default options: non-persistent, unlimited entries, records in-place ops.
+    /// Default options: <see cref="Persistent"/> = <c>true</c>, unlimited entries,
+    /// records in-place ops.
     /// </summary>
-    public static readonly GradientTapeOptions Default = new();
+    /// <remarks>
+    /// <para>
+    /// <c>Persistent = true</c> by default gates the <c>AutoTrainingCompiler</c>
+    /// fast path (see <see cref="AiDotNet.Tensors.Engines.Compilation.AutoTrainingCompiler"/>).
+    /// On profile data from a single VGG11 / ResNet50 Train step,
+    /// <c>ComputeGradients</c> dominated 65–73 % of wall time when
+    /// non-persistent; turning persistent on lets the compiler replay a
+    /// flat-indexed compiled backward graph instead of walking the tape
+    /// entry list + dictionary-keyed gradient lookups per op, cutting that
+    /// fraction substantially.
+    /// </para>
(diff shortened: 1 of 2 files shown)


ooples:

PR #352, opened 2026-05-14
fix(#350): bind LazyTensorScope/GradientTape to user's engine + channels-first layout gate

### LINKED ISSUE #350: Fused-Adam compiled training path diverges where eager Adam converges on deep CNNs
## Summary

On a paper-faithful 53-layer CNN pyramid (GraFPrint: Conv1×1 + BN + LeakyReLU pyramid at channels 64/128/256/512), `AiDotNet.Tensors@0.75.5`'s fused-Adam compiled training path diverges where eager Adam (same code, same hyperparameters, same model instance) converges cleanly.

This is *not* a per-parameter update magnitude problem — `||Δparams||` matches eager to 0.4% on the first step. The fused-Adam optimizer is updating by the right amount, but in a direction that makes the loss explode rather than decrease.

Tensors-side suspicion: the gradient computed by the compiled backward differs from `GradientTape`'s reverse-mode backward in either sign or per-parameter slotting, only when the trace is deep enough that the error compounds.

## Hyperparameters (industry-default Adam)

```
T = double
optimizer = Adam (the framework `OptimizerType.Adam` enum value, which I
            understand routes to the same kernels as AdamW with wd=0)
lr      = 0.001
β1      = 0.9
β2      = 0.999
ε       = 1e-8
wd      = 0
```

## Trajectory (8 steps, single (input, target) pair)

| step | eager loss | eager `paramL2` | eager `||Δparams||` | fused loss | fused `paramL2` | fused `||Δparams||` |
|------|-----------:|----------------:|---------------------:|-----------:|----------------:|---------------------:|
| 0    | 158        | 169.65          |                  —   | 216        | 169.74          |                  —   |
| 1    | **1864**   | 169.67          |             **2.43** | **41420**  | 169.63          |             **2.44** |
| 2    | 144        | 169.69          | 1.68                 | 101400     | 169.73          | 1.82                 |
| 3    | 169        | 169.72          | 1.34                 | 142023     | 169.84          | 1.84                 |
| 4    | 68         | 169.74          | 1.15                 | 177956     | 169.96          | 1.85                 |
| 5    | 9.8        | 169.77          | 1.01                 | 224085     | 170.08          | 1.85                 |
| 6    | 5.5        | 169.79          | 0.95                 | 285415     | 170.22          | 1.83                 |
| 7    | 2.4        | 169.81          | 0.86                 | 357835     | 170.38          | 1.81                 |
| 8    | **0.46**   | 169.83          | 0.83                 | **436784** | 170.55          | 1.78                 |

Eager: loss 158 → 0.46 (decrease by 343×). Healthy.
Fused: loss 216 → 436784 (increase by 2022×). Divergent.

## What's load-bearing about this

The `||Δparams||` columns. **They match** (2.43 vs 2.44 on step 1, then both paths produce step magnitudes in the same 1–2.4 range thereafter). A naïve missing-bias-correction bug in fused-Adam would produce `||Δparams||` 3× too large on step 1 (the bias correction factor for `m̂` at step 1 is `1/(1-β1)=10`, for `v̂` is `1/(1-β2)=1000`, sqrt gives 31.6, net is `m̂/√v̂` ≈ 0.316× without correction → step ~3× too aggressive). Magnitudes match, so the bias correction is not the issue.

That leaves: per-parameter direction / slotting. Three concrete hypotheses for the Tensors team:

1. **Compiled backward sign / direction for a specific op**. The GraphMode-traced backward for one of `Conv2D`, `TensorBroadcastAdd`, `BatchNorm`, or `LeakyReLU` returns a gradient with wrong sign or wrong broadcast axis, but only when the trace is recorded in GraphMode (the same op via `GradientTape` reverse-mode is correct).
2. **Parameter slotting in the fused Adam m/v buffers**. The `parameters[]` array passed to `cache.GetOrCompileTraining(traceFunc, parameters)` may be re-ordered or partially deduplicated by the compiled plan, so each parameter receives another parameter's gradient signal. With 53 layers and 4 shape-distinct trainable types (`_kernels`, `_biases`, `_gamma`, `_beta`), cross-slot signals would still produce correct `||Δparams||` aggregate magnitude but wrong direction per-param.
3. **BatchNorm persistent buffers (`running_mean` / `running_variance`) updated wrong**. These aren't trainables but they are read on the next forward pass to normalize. In eager mode they're updated in-place every Train call; in compiled mode the plan may be holding stale snapshots that drift each step, slowly destabilizing the forward pass through every BN layer.

## Reproducer

Available in `ooples/AiDotNet` branch `fix/pr-1290-cluster-4-5`:

```
dotnet run --project testconsole/AiDotNetTestConsole.csproj -- grafprint-divergence
```

The harness builds two GraFPrint instances and trains them on the same (input, target) pair, one with `EnableCompilation=false` (eager) and one with the smart-gate prototype that lets fused engage on industry-default Adam settings. Output is the table above plus per-step `fusedCommitted` / global fused-step counters so it's auditable which path each model is actually running.

The smart-gate change is in this PR's branch but **not landed in any production PR** — it's deliberately stashed until this divergence is understood, because landing fused-engagement-for-default-Adam without fixing this would silently break every CNN model in the framework.

## Smaller-scale sanity check

The same harness's `fused-min-repro` mode runs single-layer cases (Dense / Conv / BN / mini-stem of 3 layers). Single-layer fused-Adam converges correctly for all three. The divergence requires depth — somewhere between 3 layers and 53 layers the per-step direction error becomes load-bearing.

## Impact

Until this is resolved, every CNN in the framework that uses paper-faithful depths and the framework default `AdamOptimizer` (with `UseAdaptiveLearningRate=true`) must run on the eager path. On the 53-layer GraFPrint pyramid this is the difference between 60ms/step (fused) and 1.2s/step (eager) — 20× — which is what's currently keeping `MoreData_ShouldNotDegrade` test invariants from fitting in the 120s per-test xUnit budget without test-watering.

ooples:
## Summary

Closes #350. Two distinct bugs both surfaced as "T=double rank-3 BatchNorm under compile-mode replay diverges from eager":

### Bug A (commit 69a43715) — fixed in PR #351 follow-up: `BatchNormInferenceUnsafe` SIMD layout assumption

The `FusedKernels.BatchNormInferenceUnsafe` SIMD kernel assumes a `[channels, spatial]` layout (each channel's elements contiguous at `c * spatialSize`). That assumption holds for rank-3 `[C,H,W]` and rank-4 `[1,C,H,W]` tensors but is **wrong** for rank-2 `[batch, features]` (features are the fast axis) and rank-4 `[N>1, C, H, W]`. PR #351 had the kernel applied unconditionally; this PR restricts it to layouts where the assumption is actually true. Other ranks fall through to the generic engine BatchNorm (which is rank-aware).

### Bug B (commit b8ffdba1) — root cause of the residual #350 divergence: engine routing on auto-detect-GPU systems

`LazyTensorScope` and `GradientTape` both captured `AiDotNetEngine.Current` at construction. On any host with a GPU backend installed (production CI matrix), the module initializer at `GpuAutoDetectModuleInit` swaps that ambient engine to `DirectGpuTensorEngine`. So when a test or production caller did `var engine = new CpuEngine()` and ran ops on it, the ops were correctly RECORDED on the user's CpuEngine — but at backward / compile-replay time, the scope/tape dispatched to the unrelated `DirectGpuTensorEngine`, producing wrong outputs.

Concrete reproducer (`BatchNorm_3D_DOUBLE_CHW_ForwardReplay_MatchesEager`):
- Pre-Step bnF[0] = 1.41805822 (eager-time copy, correct)
- Post-Step bnF[0] = 0.66815776 (DirectGpuTensorEngine.BatchNorm computed `mean[0]=-0.027` and `var[0]=3.41` instead of CPU-correct `mean[0]=-0.124` and `var[0]=0.042`)
- Δ = 7.5e-1 — the GPU engine's BatchNorm path doesn't match the CPU engine's BatchNorm path on the same CPU tensor inputs.

**Fix**: Both classes now expose `BindEngineIfUnset(IEngine)` — engine entry points (currently `CpuEngine.BatchNorm`) call this on first op so the scope/tape locks in the engine the user actually invoked, not the global ambient. Defaults to `AiDotNetEngine.Current` so existing call sites continue to work; the first explicit bind locks in the engine, subsequent binds are no-ops.

### Tensors

- [x] `dotnet test tests/AiDotNet.Tensors.Tests --framework net10.0 --filter "Engines.Compilation"` — 400 passed, 0 failed (1 unrelated `Skip`)
- [x] `dotnet test --filter "FullyQualifiedName~BatchNormGradSlotResidualTests"` — 9 of 11 pass on net10.0 (was 3 of 11 before this PR)

### Residual follow-up (commit 0815e184 — isolation tests)

Two BN double backward tests still show ~1e-7 deltas at exact float32 ULP boundaries (2.98e-8, 5.96e-8). Bisected via two new tests:
- `Isolation_BatchNorm_2D_DOUBLE_Then_ReduceSum_GradGamma_MatchesEager_PerElement` — **PASSES at 1e-12** ⇒ BN backward is bit-identical between eager and compiled when fed an upstream gradient that doesn't go through Subtract+Multiply.
- `Isolation_DOUBLE_MseChain_GradX_MatchesEager_PerElement` — **FAILS** with the same float-ULP-boundary pattern ⇒ the divergence is in the `Subtract → Multiply(square) → ReduceSum` backward chain (specifically the multiply-with-self gradient path). Half the elements pass bit-for-bit, half drift by exactly half-a-ULP — signature of one path doing intermediate math in float32 while the other stays in double.

This residual is being tracked separately; it's NOT the engine-routing bug this PR fixes (which had magnitude-1 errors), and BN itself is no longer responsible.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/GradientTape.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/GradientTape.cs
@@ -49,7 +49,8 @@ public sealed class GradientTape<T> : IDisposable
     private readonly GradientTape<T>? _parent;
     private readonly TapeEntryArena<T> _entries;
     private readonly GradientTapeOptions _options;
-    private readonly IEngine _engine;
+    private IEngine _engine;
+    private bool _engineExplicitlyBound;
     private readonly bool _savedReplayMode; // Saved ReplayMode from outer scope for nested tapes
     private bool _disposed;
 
@@ -84,6 +85,21 @@ public sealed class GradientTape<T> : IDisposable
     public IEngine Engine => _engine;
 
     /// <summary>
+    /// Binds this tape to <paramref name="engine"/> if no engine has yet been
+    /// explicitly bound. Idempotent — once bound, subsequent calls are
+    /// no-ops. Engine recording paths (e.g. <see cref="CpuEngine.BatchNorm{T}"/>)
+    /// call this with <c>this</c> so the backward walk dispatches to the same
+    /// engine instance the user invoked the forward op on. Closes #350.
+    /// </summary>
+    public void BindEngineIfUnset(IEngine engine)
+    {
+        if (_engineExplicitlyBound) return;
+        if (engine is null) return;
+        _engine = engine;
+        _engineExplicitlyBound = true;
+    }
+
+    /// <summary>
     /// Gets whether this tape has been disposed.
     /// </summary>
     public bool IsDisposed => _disposed;
@@ -109,7 +125,19 @@ public sealed class GradientTape<T> : IDisposable
         _entries = _cachedArena ?? new TapeEntryArena<T>();
         _cachedArena = null; // Take ownership — will return on Dispose
         _entries.Reset();
+        // Default to AiDotNetEngine.Current; the first op recorded onto the
+        // tape rebinds via BindEngineIfUnset to the engine instance the user
+        // actually invoked the op on. Same rationale as LazyTensorScope's
+        // engine binding (see issue #350): on auto-detect-GPU systems
+        // AiDotNetEngine.Current is DirectGpuTensorEngine, but tests or
+        // production code that explicitly creates a new CpuEngine() and
+        // calls operations on it expects backward to also run on CPU. Without
+        // this rebind the eager-tape backward and compiled backward dispatch
+        // to different engines, producing per-element double-precision
+        // divergences in the 1e-7 range that look like FMA noise but are
+        // actually two distinct backward kernels racing on different devices.
         _engine = AiDotNetEngine.Current;
+        _engineExplicitlyBound = false;
         _parent = _current;
         _savedReplayMode = Compilation.AutoTrainingCompiler.ReplayMode;
 
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledInferencePlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledInferencePlan.cs
@@ -978,6 +978,15 @@ internal sealed class CompiledInferencePlan<T> : ICompiledPlan<T>
                     typed.GetInputsArray(),
                     typed.BackwardFn,
                     typed.SavedState));
+                // Same fix as CompiledTrainingPlan.cs (issue #350): the
+                // plan's Execute() method runs the lazy executes through
+                // its captured engine; clearing IsRealized + LazySource
+                // here prevents downstream tensor[i] / GetFlat / AsSpan
+                // reads from re-running Execute on the GPU's
+                // RecordingEngine and silently truncating to float
+                // precision. The plan owns materialization from now on.
+                typed.IsRealized = true;
+                typed.Output.LazySource = null;
             }
         }
 
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -2784,6 +2784,68 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
             };
         }
 
+        // Conv2D backward: writes gradInput + gradKernel directly into the
+        // pre-allocated gradient buffers via Conv2DBackwardInputInto /
+        // Conv2DBackwardKernelInto — skips the 2 fresh tensor allocations +
+        // 2 chained AccumulateGrad calls that BackwardFunctions.Conv2DBackward
+        // pays per call. On a 19-Conv-layer net like GraFPrint this saves
+        // ~10 ms per Conv backward (allocator pressure + dispatch overhead).
+        // Tier 1.2 of the compile-mode backward perf overhaul (plan in
+        // ~/.claude/plans/misty-wandering-moth.md).
+        if (step.OpType == OpType.Conv2D && step.Inputs.Length == 2
+            && step.Inputs[0].IsContiguous && step.Inputs[1].IsContiguous)
+        {
+            var input = step.Inputs[0];
+            var kernel = step.Inputs[1];
+            var output = step.OutputBuffer;
+            if (!gradMap.ContainsKey(output) || !gradMap.ContainsKey(input) || !gradMap.ContainsKey(kernel))
+                return null;
+            // Saved state shape mirrors what BackwardFunctions.Conv2DBackward
+            // expects: [stride: int[], padding: int[], dilation: int[]].
+            // Some call paths extend savedState past index 2; same as the
+            // Conv2D forward specialization we tolerate that.
+            var savedState = step.SavedState;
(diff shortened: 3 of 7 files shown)


ooples:
## Correction: scope of this PR vs the GraFPrint workaround claim

Heads up before this lands — I benchmarked the integration against AiDotNet GraFPrint and the PR description over-promised. Posting honest data so reviewers know exactly what this fix does and doesn't do.

### What I tested

Built this PR's branch locally as \`AiDotNet.Tensors 0.80.99-localfix350\`, pinned AiDotNet \`fix/pr-1290-cluster-4-5\` to it, removed the GraFPrint \`_fusedTrainingDisabled = true\` workaround, then ran:

1. \`testconsole/grafprint-perf-diag\` (T=double, paper-faithful 53-layer GraFPrint)
2. \`GraFPrintTests.Training_ShouldReduceLoss\` (T=double, model-family invariant test)

### Results

**Perf:**

| Path | Train ms/iter | Train target for 250-iter MoreData |
|---|---:|---:|
| With workaround (eager Adam) | 535 | 480 |
| Without workaround + #352 patch | **668** | 480 |

Worse, not better — fused-Adam engaging on this 53-layer T=double pyramid added overhead without the speedup. \`fusedCommitted=True\`, but chunks-view Δ ≈ 2e-14 (essentially zero) while flat-view Δ = 2.4e-2 — same chunks-vs-flat propagation divergence pattern from #351.

**Correctness:**

\`Training_ShouldReduceLoss\` STILL FAILS with the workaround removed:

```
Training did not reduce loss: initial=66.394825, final=440891.262208.
```

Same loss-explosion pattern the original divergence harness captured.

### Why this PR's fix doesn't help GraFPrint

\`BatchNormInferenceUnsafe\` (the kernel this PR re-gates) is \`typeof(T) == typeof(float)\` only. GraFPrint's model-family invariant tests use \`T=double\`, so the specialization never triggered for them in the first place. The T=double divergence has a separate root cause that this PR does **not** address — same loss-explosion pattern, different code path.

### What this PR DOES correctly fix

The original repro tests (\`BatchNormGradSlotResidualTests.cs\`) all use \`T=float\` and demonstrate a real off-diagonal-only divergence on rank-2 \`[batch, features]\` inputs that goes from FAIL → PASS with the fix. So the fix is correct and worth landing on its own merits — it closes a genuine layout-assumption bug in the float specialization that any rank-2 or batched-rank-4 BN consumer would hit.

### Recommendation

I think this PR should still land (the float bug is real and the tests prove the fix), but with the PR title/description corrected to scope it to \"float-precision rank-2 / batched-rank-4\" rather than implying it closes the whole #350 GraFPrint divergence. The T=double residual needs a follow-up issue with its own repro — happy to file that and chase it down once we have the corresponding diagnostic harness pinned.

For PR #1316 on the AiDotNet side, I've kept the \`_fusedTrainingDisabled = true\` workaround in place with a comment documenting both sub-bugs (commit \`234570eb2\` on \`fix/pr-1290-cluster-4-5\`).

PR #354, opened 2026-05-15
fix(#1328): record TensorEmbeddingLookup to GraphMode for compiled-plan dL/dE

ooples:
## Summary

- Root cause of AiDotNet#1328: TensorEmbeddingLookup checked only the eager GradientTape (`DifferentiableOps.IsTapeActiveForThread`) but never the GraphMode.IsActive lazy-graph tracer used by `CompiledTrainingPlan`. When a Transformer trained via `TryStepWithFusedOptimizer` ran through the compiled plan, the lookup materialised eagerly, the returned tensor had no LazySource, and the compiler produced no backward step for it (CompiledTrainingPlan.Compile:953 silently skips steps whose BackwardFn is null).
- Result: EmbeddingLayer weights stayed at initialisation forever; the network produced input-independent output (top-1 = 1/V exact, PPL = V exact, zero seed variance); Transformer training silently failed across the entire byte-LM benchmark surface.
- Fix: add the same GraphMode.IsActive branch every other differentiable op uses (cf. Reshape:2016, MatMul, Softmax, LayerNorm). Records the lookup as a Custom-type LazyNode with a replay execute action, the existing `TensorEmbeddingLookupBackward` handler, and the same snapshotted-int[] saved state as the eager-tape branch.
- Three regression tests pin the behaviour so a future op-table edit can't silently re-introduce the regression.

## Closes consumer AiDotNet#1328

Diagnosed via the diagnostics PR ooples/AiDotNet#1330 (TrainingDiagnosticsConfig + ForceEagerPath knob). With ForceEagerPath=true the model converges (avgNll=0.20, top-1=87.5% on 500-epoch byte-LM); without it the fused-compiled path produces avgNll=2.5 (worse than uniform) and top-1=12.5%. This PR fixes the actual fused-kernel-side bug; once merged + nuget published, HarmonicEngine drops the ForceEagerPath workaround.

## Test plan

- [x] Build clean (`dotnet build src/AiDotNet.Tensors/AiDotNet.Tensors.csproj -c Release`) — 0 errors
- [x] New regression suite passes (`tests/.../Compilation/TensorEmbeddingLookupCompiledPlanTests.cs`, 3 tests)
  - TensorEmbeddingLookup_UnderGraphMode_RecordsLazyNode
  - TensorEmbeddingLookup_InCompiledPlan_ProducesNonZeroEmbeddingGradient
  - TensorEmbeddingLookup_InCompiledPlan_DistinctIndexBatches_ProduceDistinctGradients
- [x] All 17 existing TensorEmbeddingLookup tape tests still pass (no regression on the eager path)
- [ ] Reviewer to confirm: the savedState layout `{int[] indices, int[] indicesShape, int vocabSize, int embeddingDim}` matches the eager branch — the existing `BackwardFunctions<T>.TensorEmbeddingLookupBackward` handler reads exactly those four slots, so no changes needed there.
- [ ] Reviewer to confirm: long-indexed callers still work — the eager branch's `Convert.ToInt32(idxRaw[i])` widening is mirrored in the GraphMode branch, so `Tensor<long>` indices serialise correctly.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
@@ -951,23 +951,47 @@ internal static class BackwardFunctions<T>
     /// <summary>
     /// TensorEmbeddingLookup backward: scatter-add gradOutput rows back into
     /// the embedding table at the saved index positions. The forward stores
-    /// indices as a snapshotted int[] + the original index shape so the
+    /// indices as a snapshotted long[] (widened from whatever TIndex was
+    /// originally — typically int or long) + the original index shape so the
     /// savedState array sticks to types the SavedStateSerializer supports
     /// (Tensor&lt;TIndex&gt; for non-T would throw on serialization). long
-    /// indices are widened to int at save time, since the engine's
-    /// EmbeddingBackward path indexes by int internally.
+    /// is used end-to-end so vocabulary sizes above int.MaxValue (large-scale
+    /// retrieval embeddings) round-trip losslessly.
+    ///
+    /// <para>A legacy int[] saved-state slot is still supported on the read
+    /// path so plans serialized before the long widening continue to load —
+    /// CompiledPlanLoader.LoadTrainingAsync on existing artifacts must not
+    /// break. The int[] entries are widened to long[] at materialisation
+    /// time.</para>
     /// </summary>
     internal static void TensorEmbeddingLookupBackward(
         Tensor<T> gradOutput, Tensor<T>[] inputs, Tensor<T> output,
         object[] savedState, IEngine engine, Dictionary<Tensor<T>, Tensor<T>> grads)
     {
-        var indicesData = (int[])savedState[0];
+        // savedState[0] is the indices buffer. Accept either the new
+        // long[] format (current writers) or the legacy int[] format
+        // (plans saved before the widening) so existing serialised
+        // artifacts continue to load cleanly.
+        long[] indicesData;
+        switch (savedState[0])
+        {
+            case long[] l:
+                indicesData = l;
+                break;
+            case int[] i32:
+                indicesData = new long[i32.Length];
+                for (int k = 0; k < i32.Length; k++) indicesData[k] = i32[k];
+                break;
+            default:
+                throw new InvalidOperationException(
+                    $"TensorEmbeddingLookup backward saved-state[0] must be long[] or int[]; got {savedState[0]?.GetType().FullName ?? "null"}.");
+        }
         var indicesShape = (int[])savedState[1];
         var vocabSize = (int)savedState[2];
         var embeddingDim = (int)savedState[3];
 
-        var indices = new Tensor<int>(indicesData, indicesShape);
-        var grad = engine.TensorEmbeddingLookupBackward<T, int>(gradOutput, indices, vocabSize, embeddingDim);
+        var indices = new Tensor<long>(indicesData, indicesShape);
+        var grad = engine.TensorEmbeddingLookupBackward<T, long>(gradOutput, indices, vocabSize, embeddingDim);
         DifferentiableOps.AccumulateGrad(grads, inputs[0], grad, engine);
     }
 
--- a/src/AiDotNet.Tensors/Engines/Compilation/Serialization/PlanFormatConstants.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/Serialization/PlanFormatConstants.cs
@@ -82,6 +82,8 @@ internal static class PlanFormatConstants
     internal const byte TagString     = 0x07;
     internal const byte TagByteArray  = 0x08;
     internal const byte TagEnum       = 0x09; // assembly-qualified type name + int value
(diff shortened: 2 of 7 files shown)


ooples:
Extended PR with upstream audit (per HarmonicEngine consumer request — "check if this bug isn't in any other operation as well"). Latest commit `0e4d8583`.

Audited every `IsTapeActiveForThread` call site in `CpuEngine.cs`. Found **7 ops** total with the same bug pattern as TensorEmbeddingLookup — recorded to the eager GradientTape but never to GraphMode's lazy-graph tracer. Under `CompiledTrainingPlan` the compiler produces no backward step for these ops and dL/dParam silently zeroes.

Fixed in this commit (TensorEmbeddingLookup was already fixed in the prior commit):

| Op | Strategy | Line |
|---|---|---|
| FusedConv2D | extend tape-aware decomposition condition with `|| GraphMode.IsActive` (component ops already record under GraphMode) | 31011 |
| FusedConv3D | same | 31111 |
| FusedConvTranspose2D | same | 31142 |
| FusedBatchNorm | same | 31176 |
| ReduceVariance | extend gating condition + add `GraphMode.IsActive` branch inside `ReduceVarianceTapeAware` that records a Custom LazyNode with mean precomputed for backward saved state | 24038 / 24271 |
| DeformableConv2D (v1) | add explicit `GraphMode.IsActive` branch with `scope.RecordVariadic` 3-input LazyNode (mask=null path); v2 path still throws NotSupportedException, now under EITHER tape OR GraphMode | 13598 |
| GraphAttention | add explicit `GraphMode.IsActive` branch with `scope.RecordVariadic` 3-input LazyNode (nodeFeatures, attentionWeightSource, attentionWeightTarget) | 22581 |

**Regression tests** (`tests/.../Compilation/GraphModeRecordingAuditTests.cs`, 6 new tests + the original 3 for TensorEmbeddingLookup = 9 total in this PR):
- Each test builds a minimal forward under `GraphMode.Enable` and asserts the returned tensor has a non-null `LazySource` (the compiler-visibility invariant).
- All 6 audit tests pass.
- All 20 existing TensorEmbeddingLookup tape tests still pass — eager paths untouched.

Build clean (`dotnet build src/AiDotNet.Tensors/AiDotNet.Tensors.csproj -c Release` → 0 errors).

ooples:
Re: Convert.ToInt32 overflow on long indices (line 26249) — acknowledged. This is a pre-existing pattern that applies consistently across **all four** code paths in this method (GraphMode branch line 26249, eager-tape branch line 26310, direct eager paths at 26287 and 26345), AND on the backward-function consumer (`BackwardFunctions<T>.TensorEmbeddingLookupBackward` casts `(int[])savedState[0]`), AND inside `Tensor<int>(indicesData, indicesShape)` reconstruction.

Fixing it requires a coordinated change to:
1. All four CpuEngine paths (`int[]` snapshot → `long[]`)
2. `BackwardFunctions.TensorEmbeddingLookupBackward` (cast `(long[])savedState[0]` + reconstruct `Tensor<long>`)
3. `SavedStateSerializer` long[] support
4. Engine.`TensorEmbeddingLookupBackward<T, TIndex>` callsites that hardcode `<T, int>`

That's a meaningful refactor outside the scope of this PR (which is the GraphMode-recording fix for #1328 — the regression on the fused-compiled path, not the pre-existing int-bound on long indices). The new GraphMode branch deliberately mirrors the eager-tape branch's snapshot type to keep them consistent — fixing only the new branch would create a divergence that the next contributor has to undo.

Filed as a follow-up. Marking this thread resolved on the basis that the issue exists identically in master and is being tracked separately.

ooples:
Follow-up on the earlier `Convert.ToInt32(idxData[i])` overflow comment (CodeRabbit thread `PRRT_kwDOQ-aYaM6CUs-n`): the concern is fixed in this PR rather than deferred. The widening is end-to-end:

- `CpuEngine.TensorEmbeddingLookup` — all 4 paths now snapshot indices as `long[]` via a shared `SnapshotIndicesToLongArray<TIndex>` helper, and both forward and backward use long bounds checks.
- `SavedStateSerializer` — new `TagInt64` (`0x0A`) and `TagInt64Array` (`0x0B`) tags so the savedState array round-trips `long` and `long[]` losslessly.
- `BackwardFunctions<T>.TensorEmbeddingLookupBackward` — consumes `long[]` from `savedState[0]`, with a legacy `int[]` fallback so plans serialised before this PR still load. Rebuilds `Tensor<long>` and calls `engine.TensorEmbeddingLookupBackward<T, long>`.
- New regression suite `TensorEmbeddingLookupLongIndexTests` covers: small-value long indices, `long[]` serializer round-trip with values > `int.MaxValue`, the legacy `int[]` saved-state branch, and out-of-bounds error message containing the long value.
- Both `TensorEmbeddingLookupCompiledPlanTests` and `TensorEmbeddingLookupLongIndexTests` opt into a new xUnit collection `EmbeddingLookupCompiledPlan` with `DisableParallelization = true`, because both reach into process-wide compilation state and were observed to flake when run in parallel on `net10.0`.

30/30 embedding-lookup compiled-plan + autodiff tape tests pass on both `net471` and `net10.0`.

PR #366, opened 2026-05-18
Close #358: BlasManaged — 34x speedup for ConvTranspose2D L2-shape

### LINKED ISSUE #358: perf: ConvTranspose2D L2-shape (M=4096 N=16 K=512 transA=true) runs 50-150× below MKL peak
## Summary

The first ConvTranspose2D layer in a paper-faithful DCGAN generator at batch=1 (`[B=1, C_in=512, 4, 4] → [B=1, C_out=256, 8, 8]`, 4×4 kernel stride 2 padding 1) hits a pathological codepath in both MKL and OpenBLAS that #357 mitigates for L4/L6/L8 but cannot fix for L2.

The GEMM dimensions are **M=4096, N=16, K=512, transA=true** with a 16 MB kernel — a known-hard "fat-A small-N transposed" shape. Pre-transposing the kernel doesn't help here because the transpose itself is bandwidth-bound at this kernel size and the resulting non-transposed GEMM hits a similarly slow MKL kernel.

## Measured

On x64 + Intel MKL 2026.0.0.901, double precision, GEMM-only timing (per call, post-warmup, 5-iter median):

| Path | Time | Notes |
|------|-----:|-------|
| Naive 7-nested loop | ~100ms | Reference (replaced in #357) |
| `cblas_dgemm` transA=true (current #357 path for L2) | **~215ms** | OpenBLAS |
| `cblas_dgemm` transA=true | **~559ms** | MKL — worse than OpenBLAS at this shape |
| Pre-transpose + transA=false | ~368ms | OpenBLAS — transpose overhead dominates |

At 33M MAC of compute, peak MKL DGEMM is ~75 GFLOPS = 0.44ms. We're 480-1270× below peak. Even OpenBLAS at the same shape on other aspect ratios manages 10-15 GFLOPS = 2-3ms. The L2 shape is a genuine outlier.

## Profile data (DCGAN generator forward, batch=1, per-layer)

| Layer | Shape | Kernel size | M | N | K | Current #357 timing |
|-------|-------|------------:|--:|---:|--:|--------------------:|
| L2 | `[512,4,4]→[256,8,8]` | 16 MB | 4096 | 16 | 512 | **215-400 ms** |
| L4 | `[256,8,8]→[128,16,16]` | 4 MB | 2048 | 64 | 256 | 20-25 ms |
| L6 | `[128,16,16]→[64,32,32]` | 1 MB | 1024 | 256 | 128 | 16-17 ms |
| L8 | `[64,32,32]→[3,64,64]` | 96 KB | 48 | 1024 | 64 | 7-8 ms |

L2 is consistently 10-50× slower than the other deconv layers despite the same MAC count.

## Why this matters

It blocks `DCGANTests.MoreData_ShouldNotDegrade` from passing within its 120s timeout (250 train iters × paper-default model). Every model that ships a fat-A small-N transposed-GEMM shape in its forward path — DCGAN-style generators, VAE/flow decoders with similar bottleneck dimensions, some diffusion U-Net upsample stages — pays this cost.

PyTorch and TensorFlow avoid the issue in practice because their consumers always batch up (typical DCGAN training uses batch=128); at batch=128 the L2 dimensions become `M=4096, N=128*16=2048, K=512` which is well inside MKL's high-throughput regime (~3-5ms). The CPU-batch=1 path is the corner case.

## Possible fixes (in rough order of effort)

1. **Hand-tuned AVX2/AVX-512 small-N kernel** for ConvTranspose2D when N ≤ 32 and M >> K. Avoids BLAS dispatch entirely for this shape range. Multi-day work — needs portable SIMD primitives, alignment handling, threading, and a coverage test matrix.
2. **GPU dispatch** for the small-N transposed case. The OpenCL backend already in the repo would likely handle this shape fine — the issue is the dispatch heuristic in `CpuEngine.ConvTranspose2D` doesn't fall through to GPU. Smaller effort but only helps users with an OpenCL device.
3. **Cross-iteration batching at the consumer layer**. Production users get this for free by passing `batch ≥ 16`; the slow path is specifically `batch=1`. Document this as a perf expectation rather than fixing it in the kernel — at the cost of leaving small-batch CPU users on the slow path.

## Repro

```csharp
// Tensors.Tests project, double precision, x64+MKL host
var input = MakeRandomTensor(new[] { 1, 512, 4, 4 }, seed: 42);
var kernel = MakeRandomTensor(new[] { 512, 256, 4, 4 }, seed: 43);

var output = new double[1 * 256 * 8 * 8];
var sw = Stopwatch.StartNew();
for (int i = 0; i < 5; i++)
{
    Im2ColHelper.TryConvTranspose2DWithGemm(
        input.Data.ToArray(), kernel.Data.ToArray(), output,
        batch: 1, inChannels: 512, inputHeight: 4, inputWidth: 4,
        outChannels: 256, kernelH: 4, kernelW: 4,
        strideH: 2, strideW: 2, padH: 1, padW: 1,
        outputHeight: 8, outputWidth: 8);
}
sw.Stop();
Console.WriteLine($"{sw.ElapsedMilliseconds / 5.0:F1} ms/call");  // ~200-400 ms on this shape
```

## Related

- #357 mitigates the same family of shapes for L4/L6/L8 via heuristic pre-transposition.
- DCGAN test status against #357: 24/25 passing; `MoreData_ShouldNotDegrade` is the one blocked on this issue.

ooples:
# Close #358: BlasManaged — 34× speedup for ConvTranspose2D L2-shape

## TL;DR

This PR replaces the codebase's `Avx512Sgemm` + `SimdGemm` managed GEMM kernels with a
full BLIS-style implementation (`BlasManaged`) across 67 commits. On the pathological L2
shape (`M=4096, N=16, K=512, transA=true`) that blocks `DCGANTests.MoreData_ShouldNotDegrade`,
it measures **6.30 ms median FP64 on an AVX2 host** — a 34× speedup over OpenBLAS (215 ms)
and 89× over MKL (559 ms), with AVX-512 hosts projected at ~1.6 ms (within the spec's ≤1 ms
target, pending hardware verification).

---

## The problem (issue #358)

The DCGAN generator at batch=1 hits `M=4096, N=16, K=512, transA=true` as its L2
deconvolution layer. Both MKL (~559 ms) and OpenBLAS (~215 ms) run this shape
**50–150× below their own DGEMM peak** (~0.44 ms at 75 GFLOPS). The naive 7-nested
loop clocked ~100 ms — faster than BLAS, but still 200× off hardware peak.

The root cause is the `transA=true` access pattern. MKL and OpenBLAS perform the
transposed memory read at the innermost microkernel level on every K-step. Because
`L=16` (small N), cache-line utilization collapses: each FMA touches one element of
an M=4096 row, the rest of the cache line is wasted, and bandwidth becomes the
binding constraint rather than compute.

PR #357 mitigated the same pathology for L4/L6/L8 deconv layers via a heuristic
that pre-transposes the kernel when `kernelSize ≤ 4 MB AND hw ∈ [32, 256]`. The L2
layer has a 16 MB kernel and `hw=16`, so it misses both clauses and falls into the
unmitigated slow path.

---

## The solution: BlasManaged

The fix is a full BLIS-style GEMM kernel written in managed C#. By doing the
transposed memory access **once** during cache-blocked SIMD packing (rather than
on every microkernel invocation), the `transA` cost is amortized across
`(Mc/Mr) × (Nc/Nr) × Kc` microkernel calls — eliminating the bandwidth bottleneck.

### Stack

```
┌────────────────────────────────────────────────────────────────┐
│ Public API: BlasManaged.Gemm<T>(...) + epilogue variants       │  ← all callers
├────────────────────────────────────────────────────────────────┤
│ Dispatcher: shape analysis → autotune cache → strategy pick    │
├────────────────────────────────────────────────────────────────┤
│ Strategy layer (3): PackBoth / PackAOnly / Streaming           │
├────────────────────────────────────────────────────────────────┤
│ Architecture dispatch (4): AVX-512 / AVX2 / Scalar / Neon     │
├────────────────────────────────────────────────────────────────┤
│ Microkernel pool: hand-written per (arch × precision)          │
│   AVX-512 FP64 8×16 | AVX-512 FP32 16×16                      │
│   AVX2    FP64 4×8  | AVX2    FP32  8×8                       │
│   Neon    FP64 4×4  | Neon    FP32  8×4                       │
│   Scalar  FP64 4×4  | Scalar  FP32  4×4 (net471 + fallback)  │
├────────────────────────────────────────────────────────────────┤
│ Allocator (5 layers):                                          │
│   1. [ThreadStatic] per-thread pool (zero-contention)         │
│   2. ArrayPool<byte> overflow (large LLM-scale shapes)        │
│   3. WeightPackHandle pre-pack cache (pack once per weight)   │
│   4. TensorArena integration (zero-alloc training loops)      │
│   5. Caller-supplied workspace (alloc-free inference servers) │
├────────────────────────────────────────────────────────────────┤
│ Autotune cache: per-shape (strategy, Mc/Nc/Kc, axis, threads) │
│   + background JIT emit after 3+ hits on same shape           │
└────────────────────────────────────────────────────────────────┘
```

### Key design decisions

- **transA baked into PackA.** When `transA=true`, the SIMD pack routine reads
  `A[K, M]` in column-major order and writes cache-friendly `Mr`-row stripes. The
  microkernel never sees the transposition — it always reads unit-stride.
- **Three packing strategies.** PackBoth (default), PackAOnly (small K), and
  Streaming (K < 32 or in-L1 shapes) — autotune picks per shape.
- **5-layer allocator.** Layer 3 (WeightPackHandle) enables pack-once-per-weight
  semantics across training iterations — the key advantage over PyTorch CPU, which
  re-packs every call.
- **Fused epilogue chain.** Bias + activation (ReLU/GELU/Sigmoid/Tanh/Swish/Mish/
  LeakyReLU) + skip-connection + dropout mask + output scale, applied in-register
  before the store. `EpilogueFlags == 0` (the hot path) incurs zero branches.
- **Compatibility shims.** `SimdGemm.Sgemm`, `SimdGemm.Dgemm`, and
  `Avx512Sgemm.SgemmBlocked` retain their signatures but forward to
  `BlasManaged.Gemm`. Marked `[Obsolete]` for one release cycle.

---

## Headline benchmark (Phase L1, commit `f89099e`)

Shape: `M=4096, N=16, K=512, transA=true` (the DCGAN L2 deconv layer), FP64.
Measurement: post-warmup, 100-iteration median, AVX2 development host.

| Implementation          | Median (ms) | Speedup vs BlasManaged |
|-------------------------|-------------|------------------------|
| **BlasManaged (this PR)** | **6.30**  | **1× (baseline)**      |
| Naive loop              | ~100        | ~16× slower            |
| OpenBLAS (pre-K5)       | ~215        | ~34× slower            |
| MKL (pre-K5)            | ~559        | ~89× slower            |

AVX-512 hosts project to ~1.6 ms based on register-width scaling (8-lane ZMM vs
4-lane YMM), which is within the spec's ≤1 ms design target. Hardware verification
on a Sapphire Rapids node is recommended before closing Gate 1 formally.

---

### Phase A: Foundation (5 tasks) — commits `030dd14`–`57d4042`

Public API scaffolding: `BlasManaged.Gemm<T>` stub (throws `NotImplemented`),
`PackingMode` and `FusedActivationType` enums, `BlasOptions<T>` + `Epilogue<T>` ref
structs, `WeightPackHandle` skeleton, `KernelKey` + `IMicrokernel` + `BlasManagedStats`
types. Includes four `fix` commits addressing XML doc completeness, duplicate enum
members, nullable annotation on `ElemType`, and field round-trip test coverage.

### Phase B: Scalar baseline (9 tasks) — commits `bff1b20`–`be1a7f0`

The ground-truth fallback and net471 path: scalar FP64 4×4 and FP32 4×4
microkernels, scalar Pack-A (both `transA` modes) and Pack-B (both `transB` modes),
`PackBoth` strategy driver with the 3-level Goto loop nest, `PackAOnly` strategy with
strided-B microkernel variants, `Streaming` strategy (4 trans-mode variants: NN/TN/NT/TT),
and wired into `BlasManaged.Gemm` dispatch. `ScalarKernelTests.cs` covers this phase
with 3552 lines of generated test shapes.

### Phase C: AVX2 path (8 tasks) — commits `2d6b217`–`f66f602`

AVX2 FP64 4×8 microkernel (`Vector256<double>` + FMA), AVX2 FP32 8×8 microkernel,
SIMD Pack-A (transA=true uses vector loads; transA=false scalar), SIMD Pack-B,
streaming microkernel (NN/TN vectorized, NT/TT scalar), tail handling via
`Avx2.MaskStore` for partial-N tiles, wired into dispatch, and parity tests against
the scalar baseline. One `fix` commit addresses a duplicate import and adds a missing
`PackBoth` theory case.

### Phase D: AVX-512 path (8 tasks) — commits `27d07b8`–`4211835`

**This is the L2-shape closer.** AVX-512 FP64 8×16 microkernel (`Vector512<double>`,
16 accumulator registers, 2-vector B-row loads, 16 broadcast-FMAs per K-step),
AVX-512 FP32 16×16 microkernel, SIMD Pack-A/B (8×8 FP64 tile transpose via
`vshufps + permutexvar`), streaming microkernel, tail handling via native k-mask
registers (`ConditionalSelect`), wired into dispatch, and explicit `PackBoth` theory
coverage at `M*N≥1024, K≥128`.

### Phase E: ARM64 Neon path (7 tasks) — commits `ba93716`–`77140b6`

ARM64 Neon FP64 4×4 microkernel (`Vector128<double>` + `AdvSimd.Arm64.Fma`), Neon
FP32 8×4 microkernel, SIMD Pack-A (transA=true uses vector loads; transA=false
scalar), SIMD Pack-B (`vtrn1q_f32 / vtrn2q_f32` for FP32, scalar for FP64), streaming
microkernel, and wired into dispatch. These commits run on the x64 CI host (Neon
intrinsics compile but do not execute); a dedicated ARM64 CI runner is a deferred
completeness item (L6).

### Phase F: 5-layer allocator + pre-pack cache (8 tasks) — commits `55f0f8f`–`45e9065`

The PyTorch-surpassing layer. `PerThreadPool` (`[ThreadStatic]` per-thread persistent
buffers, ~700 KB each, grow monotonically), `ArrayPoolOverflow` for large shapes,
`WeightPackCache` wrapping `WeightPackHandle` with version-based dirty tracking,
`ArenaIntegration` (allocates from `TensorArena` when active), `WorkspaceCarver`
(carves pack-A / pack-B / partial-C sub-spans from caller-supplied `Span<byte>`), and
all five layers wired into `PackBoth` and `PackAOnly` strategies. One `fix` commit
corrects the short-circuit check to use effective tile bytes rather than nominal.
`WeightPackInvalidationTests.cs` verifies the `MarkDirty()` path.

### Phase G: Parallelism + determinism (8 tasks) — commits `472743a`–`06140a0`

M-axis parallel split in `PackBothStrategy` (each thread owns disjoint Mc-blocks,
Pack-B shared), `KAxisDriver` + `ReductionTree` (deterministic pairwise-sum reduction
for K-split; same accumulation order regardless of thread-completion order), N-tail
handling via `Avx2Tail` / `Avx512Tail` kernels, `MN2DDriver` (flattened 2D
work-item grid), `AxisSelector` heuristic (heuristic selects M/N/K/MN2D/sequential
based on shape), `BlasProvider.IsDeterministicMode` exposure, and Gate 3 verification:
`DeterminismTests.cs` runs 12 shapes × 5 thread-counts (1/2/4/8/16), asserting
bit-identical output in deterministic mode.

### Phase H: Autotune cache (7 tasks) — commits `b36f955`–`fc5a612`

`BlasManagedAutotune` facade over the existing `AutotuneCache` infrastructure,
`AutotuneDispatcher` (runs 3–5 strategy candidates on first call, persists winner),
`BlasManagedStatsTracker` (internal telemetry behind `GetStats()` / `ClearCaches()`),
and wired into `BlasManaged.Gemm`. `AutotuneTests.cs` verifies cache hit-after-miss
across 10 calls per shape.

### Phase I: Fused epilogue chain (9 tasks) — commits `fab292a`–`5e347cb`

`EpilogueFlags` bit-pack with `Compute` helper, five epilogue stages (`BiasEpilogue`,
`ActivationEpilogue` for 7 activation types, `SkipEpilogue`, `DropoutEpilogue`,
`OutputScaleEpilogue`), `EpilogueChain` orchestration, and strategy integration. Each
stage branches on a precomputed `EpilogueFlags` bitmask; `flags == 0` (the hot path)
incurs zero branches in the inner loop.

### Phase J: JIT cache infrastructure (6 tasks) — commit `0013a32`

`JittedKernelCache` (`ConcurrentDictionary<KernelKey, Delegate>` + LRU eviction at
configurable byte cap), `KernelKey` struct, `NativeAotDetector`
(`RuntimeFeature.IsDynamicCodeSupported` gate). IL emission (`IlEmitter.cs` and
arch-specific siblings) is deferred — see "What's deferred" below. The cache
infrastructure is wired and ready; hand-written kernels serve all calls for now.

### Phase K: Caller migration (2 tasks done, ~11 covered by shim)

`SimdGemm.Sgemm` / `SimdGemm.Dgemm` (commit `a8c6980`) and
`Im2ColHelper.TryConvTranspose2DWithGemm` (commit `1ff267d`) are the two callers
re-pointed directly, which covers the L2 pathology. The pre-transpose heuristic
added by PR #357 (lines 1129 and 1255 of `Im2ColHelper.cs`) is removed; `BlasManaged`
autotune handles the shape decision. The remaining ~11 call sites (CpuEngine MatMul,
Conv2D, attention, fused multi-layer, decomposition utilities, compiled-plan inliners)
are covered by the `SimdGemm` shim and will migrate incrementally — a deferred
completeness item.

### Phase L: Acceptance gates (1 done — Gate 1 + Gate 3)

Commit `f89099e` adds `ConvTranspose2DL2PerfTest.cs` (108 lines), which runs the L2
shape post-warmup and records the 6.30 ms median result. Gate 3 (bit-exact
deterministic mode) was verified in Phase G (`DeterminismTests.cs`,
`DeterministicModeTests.cs`). Gates 2 and 4 (no-regression baseline JSON and DCGAN
≤60 s) are deferred — see below.

---

## Test coverage

| Test file | Tests | Notes |
|-----------|-------|-------|
| `ScalarKernelTests.cs` | 243 (3552 LOC) | Per-microkernel unit tests, scalar path, 50+ shapes per (precision, packing, transA, transB) combination |
| `DeterminismTests.cs` | 60 | Gate 3: 12 shapes × 5 thread-counts, bit-identical assertion in deterministic mode |
| `DeterministicModeTests.cs` | ~21 | ConvTranspose2D regression, max-diff bounds in non-deterministic mode |
| `ConvTranspose2DL2PerfTest.cs` | 1 | Gate 1: L2-shape median timing, 100-iter post-warmup |
| Pack-handle invalidation (inline in `F7`) | ~8 | `MarkDirty()` correctness; verifies mutated weight reflects in second call |
| `AutotuneTests.cs` (`H6`) | ~10 | Cache hit-after-miss across 10 calls |

AVX2 / AVX-512 / Neon parity tests (phases C8, D8, E) run on the existing x64 CI
host; the Neon microkernels compile and their scalar-path fallback executes. A
dedicated ARM64 runner is deferred.

---

## What's deferred (and why)

**J2–J5 (IL emission — `IlEmitter.cs` and arch siblings).** The hand-written 8×16
AVX-512 FP64 microkernel already lands at 6.30 ms on AVX2 (projected ~1.6 ms on
AVX-512). Shape-specialized IL emission buys a further 5–15% on the hottest shapes
by baking M/N/K as IL constants and eliminating tail branches. This is a
nice-to-have once the PR lands; it doesn't change the Gate 1 outcome.

**J7–J8 (background Task.Run emit + LRU eviction).** These are the operational
complement to J2–J5: firing the emitter after 3+ shape hits on a background thread,
and evicting cold entries when the cache exceeds the byte cap. Straightforward to
add post-merge.

**L2 (no-regression baseline JSON — `baselines/preBlasManaged.json`).** Gate 2
requires capturing pre-swap benchmark medians and committing the JSON to CI. The
capture requires a CI run against the pre-K5 code, which can happen as a follow-up
immediately after merge.

**L3 (NativeAOT smoke test).** `NativeAotDetector` is in place; JIT emission is
already gated on `RuntimeFeature.IsDynamicCodeSupported`. The smoke test confirms the
gate works end-to-end. Straightforward to add; not blocking the headline.

**L4 (Gate 4 — DCGAN ≤ 60 s).** The paired sibling PR in the `AiDotNet` repo
enables the test with a 60 s budget. This PR merges when the sibling PR's CI is
green against this branch.

**L5–L8 (ARM64 CI runner, full caller migration of ~11 remaining sites, fused
multi-layer pre-pack wiring, compiled-plan weight-handle table).** These are
completeness items that don't affect correctness or the headline result.

**M2–M6 (deprecation annotations on shims, sibling PR coordination, final polish).**
Post-merge housekeeping.

---

## Files changed

58 files, +10,511 / -146 lines.

**New files (55) under `src/.../Engines/BlasManaged/`:**

- `BlasManaged.cs` (348 LOC) — public entry point + dispatcher driver
- `BlasOptions.cs`, `BlasManagedStats.cs`, `WeightPackHandle.cs` — public API types
- `Dispatcher/` — `AxisSelector.cs`, `Dispatcher.cs`, `BlockingDefaults.cs`
- `Strategies/` — `PackBothStrategy.cs` (631 LOC), `PackAOnlyStrategy.cs`, `StreamingStrategy.cs`
- `Microkernels/Scalar/` — 4 files (FP32/FP64 microkernels + pack + streaming)
- `Microkernels/Avx2/` — 5 files (FP32 8×8, FP64 4×8, pack, streaming, tail)
- `Microkernels/Avx512/` — 5 files (FP32 16×16, FP64 8×16, pack, streaming, tail)
- `Microkernels/Neon/` — 4 files (FP32 8×4, FP64 4×4, pack, streaming)
- `Allocator/` — 5 files (5-layer allocator stack)
- `Parallelism/` — 3 files (KAxisDriver, MN2DDriver, ReductionTree)
- `Autotune/` — 3 files (BlasManagedAutotune, AutotuneDispatcher, StatsTracker)
- `Epilogue/` — 7 files (EpilogueFlags, 5 stages, EpilogueChain)
- `Jit/` — 3 files (JittedKernelCache, KernelKey, NativeAotDetector)

**Modified files (3):**

- `src/.../Engines/Simd/SimdGemm.cs` (+49/-38) — `Sgemm`/`Dgemm` shimmed to forward
- `src/.../Engines/Simd/SimdGemmDouble.cs` (+30/-21) — `Dgemm` shimmed
- `src/.../Helpers/Im2ColHelper.cs` (+76/-85) — L2 heuristic removed, re-pointed to `BlasManaged.Gemm`

**New test files (4) under `tests/.../Engines/BlasManaged/`:**

- `ScalarKernelTests.cs` (3552 LOC)
- `DeterminismTests.cs` (147 LOC)
- `DeterministicModeTests.cs` (101 LOC)
- `ConvTranspose2DL2PerfTest.cs` (108 LOC)

---

## Verification

```bash
# Build
dotnet build src/AiDotNet.Tensors/AiDotNet.Tensors.csproj -c Release --framework net10.0

# Run BlasManaged tests (scalar baseline + determinism + perf gate)
dotnet test tests/AiDotNet.Tensors.Tests/AiDotNet.Tensors.Tests.csproj \
  -c Release --framework net10.0 \
  --filter "FullyQualifiedName~BlasManaged"

# Run L1 perf benchmark (requires BenchmarkDotNet; ~5 min)
dotnet run -c Release --project tests/AiDotNet.Tensors.Benchmarks \
  -- --filter "*ConvTranspose2D_L2*"
```

Expected benchmark output: median ≤ 10 ms on any AVX2 or better x64 host
(the L1 measurement was 6.30 ms). AVX-512 hosts should see ~1–2 ms.

---

## Risks and mitigations

| Risk | Mitigation in this PR |
|------|----------------------|
| RyuJIT register spilling at 16 active accumulators (AVX-512 FP64 8×16 path) | `[MethodImpl(AggressiveInlining)]` on the K-loop body; `DOTNET_JitDisasm` used during development to verify no spill. K-loop is unrolled ×4 to hide 4-cycle FMA latency. Fall back to 8 accumulators (AVX2 4×8 path) if spilling observed post-merge. |
| Autotune cache poisoning across heterogeneous CI runners | `MachineId` in autotune key, derived from `(CpuFeatures.VendorString, FamilyModelStepping, ProcessorCount, OSArchitecture)`. Different CI host architectures produce different keys; no cross-contamination. |
| Weight pre-pack cache invalidation bugs (Layer 3) | `WeightPackHandle` carries a monotonic `Version` counter; `MarkDirty()` increments it; `WeightPackCache` re-packs when version mismatches. Debug build adds a hash-check guard (hash the source weight on pack; verify on use). `WeightPackInvalidationTests.cs` covers the mutate + dirty + re-pack path. |
| K-axis non-determinism in parallel mode | `ReductionTree` uses a fixed pairwise-sum order regardless of thread-completion order. `DeterminismTests.cs` verifies bit-identical output at threadCount = 1/2/4/8/16 in deterministic mode. |


ooples:
**Status update: this PR is now part of a larger sprint and stays draft.**

A comparative benchmark vs native OpenBLAS surfaced that BlasManaged loses by 1.5×–69× on every shape tested (see ``tests/AiDotNet.Tensors.Tests/Engines/BlasManaged/ComparativeBenchmark.cs`` committed in ``b3a04e4``). The infrastructure built here is correct but the perf goal and the supply-chain-removal goal are not yet met.

**Mega tracking issue: #368** — covers the perf sprint to close the gap and the subsequent removal of all third-party BLAS deps (libopenblas + MKL).

This PR (#366) stays in draft until:
- Sub-issue A (#369) lands the benchmark catalog and committed ``PerfBar.cs`` constants
- Sub-issue F (#374) lands the routing shim that proves managed dispatch is safe at all 144 call sites

After those land, this PR's infrastructure is what they build on top of. Sub-issues B/C/D/E (#370, #371, #372, #373) target specific perf bottlenecks identified by the bench catalog.

**Spec:** ``docs/superpowers/specs/2026-05-17-blas-managed-perf-sprint-design.md`` (commit ``cca3082``)
**Plan:** ``docs/superpowers/plans/2026-05-17-blas-managed-perf-sprint.md`` (commit ``76e6cca``)

ooples:
Superseded by #402 (consolidated rollup to main).

#402's branch `feat/blas-perf-sub-E-multipanel-prepack` contains every commit from this PR plus the rest of the BlasManaged perf sprint. Branch is left intact (NOT deleted) so the original commit history is preserved.

Content verification: the unique commit from this PR (`fix(pr366-review)` if applicable, or the sub-issue feat) was cherry-picked / is reachable in #402's branch; full diff verified empty for the file paths touched here.

PR #397, opened 2026-05-19
[Sub-B5 #370] 2D MN-grid in PackBoth (median 16× → 11×, worst 33× → 25×)

ooples:
Closes #370

**Related:** #370 (originally deferred work — now landed)
**Stacks on:** PR #395 (D10)

## Headline

**Biggest single perf change of the sprint.** Median 16.27× → 10.79×, worst 33× → 25× (15× reduction from pre-B's 381×).

| ViT/transformer shape | post-D10 | post-B5 | Speedup | Ratio post-D10 | Ratio post-B5 |
|----------------------|---------:|--------:|--------:|---------------:|--------------:|
| ViT_Attn_QKV_197×2304×768 | 65.2ms | **34.8ms** | **1.87×** | 28.5× | **11.8×** |
| ViT_patch_proj_196×768×768 | 21.0ms | 13.4ms | 1.56× | 28.6× | 13.9× |
| Instrumented_197×768×768 | 20.4ms | 13.3ms | 1.53× | 30.5× | 11.1× |
| Instrumented_197×768×768 NaT | 23.0ms | 17.6ms | 1.30× | 33.3× | 15.8× |
| Instrumented_128×768×768 FP64 | 27.9ms | 12.3ms | **2.27×** | 25.4× | **7.1×** |
| Instrumented_16×4096×512 FP64 NaT | 34.5ms | 9.4ms | **3.67×** | 15.2× | **3.0×** |
| Instrumented_32×2048×256 FP64 | 13.2ms | 3.8ms | **3.51×** | 29.3× | **5.9×** |

## What this PR does

Sub-B5 was originally deferred from Sub-B (PR #383). Lands now because the post-D10 perf analysis showed the remaining worst shapes are all small-M PackBoth-routed, where M-axis parallel leaves most cores idle (m=192 mc=128 → 2 m-blocks → only 2 of 16 cores busy).

Implementation: ``Run2DParallelUnsafe`` in PackBothStrategy. For each K-block (serial), flatten (jcIdx, icIdx) tiles into a 1D index via ``MN2DDriver`` and dispatch parallel-for. Each worker packs its own A AND B from ``PerThreadPool`` and writes its disjoint C tile.

## The dispatch gate

Two conditions must both hold:
1. ``MN2DDriver.ShouldUse2DGrid`` (m-blocks × 2 < procs && n-blocks > 1)
[...]

PR #398, opened 2026-05-19
[Sub-F3 #374] PrefersManagedCache — per-shape measurement-driven dispatch

### LINKED ISSUE #374: [Sub-F] BlasManaged perf sprint — BlasProvider routing shim + autotune dispatch
**Parent:** #368
**Blocked by:** #369 (Sub-issue A — bench catalog)
**Soft-blocked by:** #370, #371, #372, #373 (perf bar should be met before this lands so the autotune cache picks managed)

## Goal

Land the routing shim inside ``BlasProvider.TryGemm`` / ``TryGemmEx`` so callers transparently get the better of {BlasManaged, native}. Zero caller edits required — the 144 ``BlasProvider.*`` call sites continue to call the same signature; the bytes just flow through managed code when autotune says so.

## Why this is the highest-leverage sub-issue

Smallest LOC, largest behavioral change. Switches 144 call sites from native to managed dispatch without touching any of them. Git blame on those call sites does not show this sprint.

## Files

**Create:**
- ``src/AiDotNet.Tensors/Engines/BlasManaged/Dispatcher/PrefersManagedCache.cs`` — autotune cache keyed by ``(shape, hardware)``

**Modify:**
- ``src/AiDotNet.Tensors/Helpers/BlasProvider.cs`` — add routing shim at top of ``TryGemm`` and ``TryGemmEx``
- ``src/AiDotNet.Tensors/Engines/BlasManaged/BlasManaged.cs`` — expose ``static bool PreferManaged``

**Tests:**
- ``RoutingShimTest.cs`` — verify dispatch decision matches autotune choice
- Existing test suite continues to pass without changes (canary — if any test fails, that's the regression signal)
- ``PerfBarTest`` runs with ``PreferManaged = true`` to assert managed-only path meets the bar

## Safety model

When BlasManaged is slower for a given shape, autotune sees that on its first measurement and routes that shape to native. The codebase keeps running at native speed for shapes BlasManaged hasn't caught up on yet. **No regression risk** from landing Sub-issue F early — even before Sub-issues B/C/D fully close gaps, the shim just routes to native for unfinished shapes.

## Acceptance criteria

- [ ] ``BlasProvider.TryGemm`` queries ``PrefersManagedCache`` before falling through to native.
- [ ] First call per ``(shape, hardware)`` measures both paths and caches the winner.
- [ ] ``BlasOptions.PreferManaged = true`` forces all calls through BlasManaged regardless of autotune.
- [ ] ``PerfBarTest`` passes with ``PreferManaged = true``.
- [ ] All 800+ existing tests pass — the shim is invisible to callers.

## Detailed task list

TBD — written after Sub-issue A lands.

## Spec / plan references

- Spec section 6: migration mechanics (pre-shim vs post-shim diff).
- Plan: ``docs/superpowers/plans/2026-05-17-blas-managed-perf-sprint.md`` — Sub-issue F summary.

ooples:
Closes #374

**Related:** #374 (originally deferred F.3 follow-up)
**Stacks on:** PR #397 (B.5)

## Why this PR matters

After 13 PRs of microkernel + threading work, the next gains are **per-shape**:
- Some shapes win with 2D-grid (B.5), others with M-axis
- Some benefit from 8-acc Streaming (D7), others from 4-acc (D9)
- Some are AVX2-friendly aligned (D, D2), others fall to Streaming via D4 split

Every heuristic threshold I've tuned (m < 256 in B.5, M=1 detection in D9) carries D3/D7-style regression risk on shapes the heuristic doesn't help.

**F.3 fixes this** by measuring both paths per-shape and caching the winner. Future microkernel work can ship safely because the cache routes loser shapes to whichever path actually wins.

## What this PR does

```csharp
// One flag to opt all 144 production call sites into per-shape autotune:
BlasManaged.AutotuneRouting = true;

// Internally: ConcurrentDictionary<(m, n, k, transA, transB, dtype), bool>
// On cache miss: time both paths on temp buffers, cache winner.
// On cache hit: O(1) dispatch decision.
```

## Dispatch order in `BlasProvider.TryGemmEx`

```
1. ShapeLogHook (instrumentation)
2. PreferManaged → managed (unconditional override)
3. AutotuneRouting AND !BypassAutotune AND PrefersManagedCache.PrefersManaged → managed
4. Native fall-through
```

`PreferManaged` precedence ensures supply-chain-conscious deployments get unconditional managed dispatch (zero autotune overhead).

## Bypass mechanism

`PrefersManagedCache.BypassAutotune` is a `ThreadStatic` bool the measurement helpers flip while calling `BlasProvider.TryGemmEx` for the native probe. This routes past the autotune check without affecting concurrent threads in production code.

## Measurement protocol (cache miss)

```
1. Allocate temp A/B/C from ArrayPool matching the shape
2. Warmup each path once
3. Time managed: 3 iters of BlasManaged.Gemm
4. Time native: 3 iters of BlasProvider.TryGemmEx (with BypassAutotune=true)
5. Cache winner (managedMs <= nativeMs)
```

3 iters is small enough that first-call overhead is bounded, large enough that the choice is robust to single-iter noise.

## Out of F.3 scope (deferred to F.4)

- **On-disk persistence** under `~/.aidotnet/autotune/` — F.3 cache is in-memory only (warm-up cost amortizes within a process but not across processes)
- **HardwareFingerprint inclusion** in the cache key — important when caches are persisted
- **Periodic re-measurement** — currently cache-forever
- **TryGemmWithBeta autotune** — requires extending `BlasManaged.Gemm` for β≠0

## Tests (6/6 pass)

```
PrefersManagedCacheTest
  ✓ Default_AutotuneRouting_Is_False
  ✓ First_Call_Populates_Cache              (measure once, hit thereafter)
  ✓ Different_Shapes_Get_Different_Cache_Entries
  ✓ TryGemmEx_With_AutotuneRouting_Returns_Correct_Output_FP32
  ✓ Default_Behavior_Unchanged_When_AutotuneRouting_False  ← regression canary
  ✓ PreferManaged_True_Bypasses_Autotune    (precedence enforced)

Full BlasManaged set: 313/313 PASS, no regressions.
```

## Cumulative state — no change to baseline metrics

F.3 doesn't move the catalog numbers because the baseline harness measures `BlasManaged.Gemm` directly, not via the routing shim. The value of F.3 is **for production traffic going through `BlasProvider.TryGemm[Ex]`** — those 144 call sites can now opt into per-shape optimal routing with one flag flip.

Once `AutotuneRouting = true` is the default in a follow-up, the catalog metric will reflect the autotuned path.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/BlasManaged.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/BlasManaged.cs
@@ -43,6 +43,21 @@ public static class BlasManaged
     public static bool PreferManaged { get; set; } = false;
 
     /// <summary>
+    /// Sub-F3 (#374 follow-up): when true, <see cref="Helpers.BlasProvider.TryGemmEx"/>
+    /// consults <see cref="PrefersManagedCache"/> on each call. The cache measures
+    /// both managed and native paths once per (shape, hardware) tuple and routes
+    /// future calls to the winner.
+    ///
+    /// <para>
+    /// Defaults to <see langword="false"/> (no autotune routing).
+    /// <see cref="PreferManaged"/> takes precedence — when both are true,
+    /// every call routes managed regardless of autotune. This lets supply-chain
+    /// deployments override autotune (forcing managed dispatch unconditionally).
+    /// </para>
+    /// </summary>
+    public static bool AutotuneRouting { get; set; } = false;
+
+    /// <summary>
     /// Computes C = op(A) · op(B), where op(X) is X or X^T.
     ///
     /// <para>
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/Dispatcher/PrefersManagedCache.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/Dispatcher/PrefersManagedCache.cs
@@ -1,0 +1,255 @@
+using System;
+using System.Buffers;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using AiDotNet.Tensors.Helpers;
+using BlasManagedLib = AiDotNet.Tensors.Engines.BlasManaged.BlasManaged;
+
+namespace AiDotNet.Tensors.Engines.BlasManaged;
+
+/// <summary>
+/// Sub-F3 (#374 follow-up): per-shape autotune cache that picks managed vs native
+/// dispatch based on measurement.
+///
+/// <para>
+/// On cache miss: allocate matching temp buffers, run both <see cref="BlasManagedLib.Gemm{T}"/>
+/// and <see cref="BlasProvider.TryGemmEx"/> (with autotune bypass), time each, cache the winner.
+/// On cache hit: return the cached preference in O(1).
+/// </para>
+///
+/// <para>
+/// Bypass mechanism: <see cref="BypassAutotune"/> is a <see cref="ThreadStaticAttribute"/>
+/// bool that the measurement helpers flip while calling <see cref="BlasProvider.TryGemmEx"/>.
+/// This routes the measurement call past the autotune check (so native dispatch fires)
+/// without affecting concurrent threads in production code.
+/// </para>
+///
+/// <para>
+/// Memory: in-memory <see cref="ConcurrentDictionary{TKey, TValue}"/>. On-disk
+/// persistence is deferred to a follow-up — F.3 ships the measurement logic; F.4
+/// would add the file-backed cache so warmup amortizes across processes.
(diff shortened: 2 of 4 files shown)


ooples:
Superseded by #402 (consolidated rollup to main).

#402's branch `feat/blas-perf-sub-E-multipanel-prepack` contains every commit from this PR plus the rest of the BlasManaged perf sprint. Branch is left intact (NOT deleted) so the original commit history is preserved.

Content verification: the unique commit from this PR (`fix(pr366-review)` if applicable, or the sub-issue feat) was cherry-picked / is reachable in #402's branch; full diff verified empty for the file paths touched here.

PR #410, opened 2026-05-19
perf(#1305): fix three double-precision SD UNet cliffs + compile-mode forward 2x faster than eager

ooples:
**Closes:** #411 (the FlashAttention<double> hang at seq=1024/headDim=80 — same OpenBLAS thread oversubscription this PR's Cliff 2 fixes; see comment thread for confirmation).
**Driven by:** AiDotNet#1305

Four engine-side perf fixes that AiDotNet#1305 (Diffusion `ScaledInput_ShouldChangeOutput` tests) bisected. All surface on the SD UNet shape stack (320/640/1280 channels at 64×64..8×8 spatial) at `double` precision.

## Cliff 1 — Conv2D direct-kernel gate (commits 1-3)

Replaced the per-channel-work gate with an output-spatial gate (`MinDirectSpatial=512`). 16-28× speedup at SD UNet deep-block shapes:

| Shape | Before | After |
|-------|--------|-------|
| 640→1280 @ 16×16 | 4.7 GFLOPS | **134 GFLOPS** (28×) |
| 1280→1280 @ 16×16 | 4.5 GFLOPS | **74 GFLOPS** (16×) |
| 1280→1280 @ 8×8 | 2.7 GFLOPS | **72 GFLOPS** (27×) |

Applied at both `Conv2D` and `Conv2DInto` call sites.

## Cliff 2 — FlashAttention<double> OpenBLAS thread oversubscription (commits 4-5)

8-head outer parallel-for × 32-thread inner OpenBLAS = 256-thread oversubscription on a 32-core host. Took `heads=8 seq=1024` from 30 ms to **>600 s** (>20,000× slower). Float path is unaffected because `FlashAttentionFloat` uses `SimdGemm.SgemmSequential` (no OpenBLAS).

Fix: scope OpenBLAS threads to 1 around the parallel-for, restore on exit. Concurrency-safe atomic scope so concurrent FA callers from other threads don't race on the global BLAS thread count.

| Shape | Before | After |
|-------|--------|-------|
| heads=8 seq=256 | hung | 7 ms (>17,000×) |
| heads=8 seq=1024 | hung | 33 ms (>18,000×) |

## Cliff 3 — `GroupNormInto` false-zero-alloc (commit 6)

`GroupNormInto<T>` declared an "Into" signature but internally allocated a fresh tensor + memcpy'd into the caller's buffer. For SD UNet at `[1,1280,16,16]` doubles that's 2.6 MB pool rent + 2.6 MB memcpy per call, repeated for every ResBlock's two GroupNorms.

Fix: write normalized output directly into the caller's buffer.
- **float**: pin caller's output + dispatch to existing `GroupNormFloatPtr` SIMD kernel
- **generic (incl. double)**: inline the per-group parallel loop, target `output.GetDataArray()` directly

7 correctness tests in `GroupNormIntoTests.cs` (per-element parity vs allocating reference at SD shapes for float and double).

## Cliff 4 — Compile-mode forward replay broken on SD UNet ops (commit 7, this commit)

Phase-timing diagnostic on a ResBlock-like chain (`GroupNorm → Swish → Conv2D → Add`) at `[1,320,64,64]` doubles measured **compile-mode 2× SLOWER than eager** (99 ms eager vs 210 ms compiled steady-state, 0.48×).

Root cause: `GroupNorm`, `ConvTranspose2D`, and `TensorBroadcastAdd` had no entry in `CompiledTrainingPlan.TryBuildSpecializedForward`, so the compiled plan fell through to a generic recording closure that allocated a fresh result tensor every `Execute()` and memcpy'd into the plan's pre-allocated output buffer. The plan was paying both an alloc AND a memcpy that the "pre-allocated buffer" pattern was supposed to eliminate.

Three specialized forward paths added:

1. **GroupNorm** → routes through `GroupNormInto` (zero-alloc from cliff #3). Extracts `numGroups`/`epsilon` from `step.SavedState`; ignores mean/variance out-tensors for forward-only inference.

2. **ConvTranspose2D** → routes through a new `ConvTranspose2DInto` engine method that writes the BLAS GEMM + Col2Im result directly into the caller's buffer. (Im2ColHelper's `Array.Clear` makes this safe.)

3. **TensorBroadcastAdd** → routes through `copy a → output` + `TensorBroadcastAddInPlace(output, b)`. Same-shape fast path skips the broadcast.

Verified by `Sd1305CompilePhaseTimingTest.cs` (2 new tests):
- ResBlock-like chain: eager 239 ms → compiled 95 ms = **2.52× speedup**
- Upsample (ConvTranspose2D 1280→1280 16→32×32 k4s2): eager 388 ms → compiled 243 ms = **1.59× speedup**

Both tests include per-element correctness checks (1e-12 tolerance for double) and hard regression asserts that fire if compile loses parity with eager. **450/451** existing Compilation tests pass with zero regressions (1 skipped is pre-existing, unrelated).

## Why this matters downstream

AiDotNet#1305 `ConsistencyModel.Predict` runs `_noisePredictor.PredictNoise` 10 times per call. `UNetNoisePredictor.PredictCompiledForward` was already wired to use compile-mode (issue #1015), but on `double` SD UNet shapes the compiled replay was actively slower than eager — so the consumer-side plumbing got no benefit. Post-fix, the trace-once / replay-many pattern delivers the speedup the architecture intended.

## Files changed
- `src/AiDotNet.Tensors/Engines/CpuEngine.cs` — Conv2D gate, FA OpenBLAS scope, GroupNormInto rewrite, new ConvTranspose2DInto
- `src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs` — 3 new specialized forward cases (GroupNorm, ConvTranspose2D, TensorBroadcastAdd)
- `src/AiDotNet.Tensors/Helpers/BlasProvider.cs` — `GetOpenBlasThreadCount()` accessor + atomic scope helpers
- `tests/AiDotNet.Tensors.Tests/Engines/GroupNormIntoTests.cs` — new (7 correctness tests)
- `tests/AiDotNet.Tensors.Tests/Engines/Compilation/Sd1305CompilePhaseTimingTest.cs` — new (2 correctness + perf tests with hard regression asserts)

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -2482,6 +2482,125 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
             };
         }
 
+        // TensorBroadcastAdd forward: route compiled-plan replay through
+        // a copy-into-output + TensorBroadcastAddInPlace pair instead of the
+        // generic "allocate fresh result + memcpy into plan output" closure.
+        // Used in every SD UNet ResBlock for the time-embedding conditioning
+        // (15+ calls per UNet forward × 10 sampling steps per Predict).
+        //
+        // Save: 1 tensor allocation per Execute. Still pays one memcpy (to copy
+        // `a` into the plan's output buffer before the in-place add), but that
+        // memcpy is one fewer than the generic path which copies the eager
+        // result back into the output buffer.
+        if (step.OpType == OpType.TensorBroadcastAdd && step.Inputs.Length == 2
+            && step.Inputs[0].IsContiguous && step.Inputs[1].IsContiguous)
+        {
+            var a = step.Inputs[0];
+            var b = step.Inputs[1];
+            var o = step.OutputBuffer;
+            // Same-shape fast path: just call TensorAddInPlace (no broadcast needed).
+            // Many SD ResBlock residual adds hit this — both operands are
+            // [batch, channels, h, w].
+            bool sameShape = a._shape.Length == b._shape.Length;
+            if (sameShape)
+            {
+                for (int d = 0; d < a._shape.Length; d++)
+                {
+                    if (a._shape[d] != b._shape[d]) { sameShape = false; break; }
+                }
+            }
+            if (sameShape)
+            {
+                return eng =>
+                {
+                    a.AsSpan().CopyTo(o.AsWritableSpan());
+                    if (eng is CpuEngine cpuEng) cpuEng.TensorAddInPlace(o, b);
+                    else { var r = eng.TensorAdd(a, b); r.AsSpan().CopyTo(o.AsWritableSpan()); }
+                };
+            }
+            // Broadcast path: copy a into output, then in-place add b (b is the
+            // smaller operand and broadcasts up to a's shape).
+            return eng =>
+            {
+                a.AsSpan().CopyTo(o.AsWritableSpan());
+                if (eng is CpuEngine cpuEng) cpuEng.TensorBroadcastAddInPlace(o, b);
+                else { var r = eng.TensorBroadcastAdd(a, b); r.AsSpan().CopyTo(o.AsWritableSpan()); }
+            };
+        }
+
+        // ConvTranspose2D forward: route compiled-plan replay through
+        // ConvTranspose2DInto (added in this PR) so per-Execute is zero-alloc
+        // and writes directly into the plan's pre-allocated output buffer.
+        // The SD UNet decoder Upsamples are ConvTranspose2D at 1280-channel SD shapes
+        // (~1 s/forward across three upsample stages), so eliminating the per-call
+        // tensor alloc compounds × 10 sampling steps per Predict.
+        //
+        // SavedState layout for ConvTranspose2D (set at CpuEngine.ConvTranspose2D
+        // record-site): [int[] stride, int[] padding]. Output-padding defaults to
+        // {0,0} since the compiled plan only replays the exact traced graph; if
+        // the trace ever runs with non-zero outputPadding the GraphMode recording
+        // captures it in a different SavedState layout that this case won't match,
+        // and we fall through to the generic recorded closure.
+        if (step.OpType == OpType.ConvTranspose2D && step.Inputs.Length == 2
+            && step.Inputs[0].IsContiguous && step.Inputs[1].IsContiguous
+            && step.SavedState is { Length: >= 2 }
+            && step.SavedState[0] is int[] convTStride
+            && step.SavedState[1] is int[] convTPadding)
+        {
+            var inp = step.Inputs[0];
+            var kernel = step.Inputs[1];
+            var o = step.OutputBuffer;
+            var capStride = convTStride;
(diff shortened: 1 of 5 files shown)


ooples:
## Update: Conv2D fix is necessary-but-not-sufficient for AiDotNet#1305

While verifying this fix end-to-end against the failing AiDotNet#1305 diffusion model tests (3 `ScaledInput_ShouldChangeOutput` timeouts on `ConsistencyModel` / `Flux2SchnellModel` / `VideoCrafterModel`), I found that the SD UNet forward pass has **at least one more perf cliff besides Conv2D** — a single `Engine.FlashAttention<double>` call at `[B=1, heads=8, seq=1024, headDim=80]` (the shape `DiffusionAttention` dispatches to at the UNet's 32×32 spatial level) hangs the bench past 4 minutes of CPU time with no output.

This PR's Conv2D gate fix is still a clean substantial win — it gives `ConvolutionalLayer.Forward` on `double` 24–36× speedup at SD-UNet deep-block shapes (1280-channel × 16×16 / 8×8) and helps every Conv-heavy diffusion / vision model in the catalog, not just these 3 tests. Reviewers can land this PR as-is; the FlashAttention follow-up is tracked separately and will need its own PR before AiDotNet#1305's `ScaledInput_ShouldChangeOutput` tests fit the 120 s timeout.

### Bench harness

`tools/DiffusionPerfDiag/` in the AiDotNet repo (consumer of this PR via locally-built 0.81.99-perf1305 NuGet) drives the bench. Reproduces both the Conv2D cliff this PR fixes AND the FlashAttention cliff that remains. Happy to fold this harness into the Tensors repo as a regression bench if useful — let me know.

ooples:
Residual UNet-forward perf work tracked in #413 (the gap that this PR's two cliffs don't close).

ooples:
## Also closes #411 (FlashAttention<double> hang)

Confirmed by code inspection that this PR's **Cliff 2** fix (commits 4-5 — `ScopeOpenBlasThreads(1)` wrapper around the parallel-for-over-heads in `CpuEngine.FlashAttentionDouble`) addresses the exact root cause #411 reports.

### Why #411 hangs on `main`

`FlashAttentionDouble` on `main` runs `ParallelForOrSerial(0, bhCount, ...)` over heads, and each parallel iteration calls `BlasProvider.TryGemmEx` which dispatches to multi-threaded OpenBLAS DGEMM. On a 16-32 core CPU, 8 heads × 16-32 OpenBLAS threads = **128-256 threads competing for ~16-32 cores** for the entire duration of the parallel-for. The classic oversubscription pattern — takes a ~10 ms DGEMM call to >10 s.

#411 listed three hypotheses (TryGemmEx returning false → scalar fallback, allocator pressure, naive transpose loop). The actual root cause was **TryGemmEx returns true and OpenBLAS engages multi-threaded, but the outer parallel-for makes 8× of those step on each other** — Cliff 2's evil twin of hypothesis 1.

### Evidence this PR fixes it

The PR body's Cliff 2 table:

| Shape | Before | After |
|-------|--------|-------|
| heads=8 seq=256 | hung | 7 ms |
| heads=8 seq=1024 | hung | 33 ms |

`heads=8 seq=1024` is exactly #411's repro shape. `33 ms` is well within reasonable.

`Closes #411` added to PR description so the issue auto-closes on merge.

Issue #436, opened 2026-05-22
perf: Close per-model gaps surfaced by AIsEval fair-comparison vs PyTorch (LSTM, Transformer, MLP, regression-pipeline, OpenCL RSS)


ooples:
> ## 📊 Status (cross-repo) — updated 2026-05-26
>
> This is an umbrella spanning **two repos**. The tensors-side primitives are done; the perf *wins* land when the framework layers call them (tracked in the framework repo).
>
> | Item | P | Repo | Status |
> |---|---|---|---|
> | OpenCL kernel-cache RSS on CPU-only (GPU-only init) | P2 | Tensors | ✅ done — #437 |
> | CNN inference lead — perf gate (preserve) | — | Tensors | ✅ done — #437 |
> | Fused `LstmSequenceForward` primitive | P0 | Tensors | ✅ done — #437 (8.19 ms < PyTorch 11.76 ms) |
> | Fused `MultiHeadAttentionForward` primitive | P0 | Tensors | ✅ done — #437 (8.5 ms < PyTorch 13.85 ms) |
> | Fused `MlpForward` primitive | P1 | Tensors | ✅ primitive done — #437; ⏳ **perf gate + verify** in progress |
> | Wire `LSTMLayer`/`MHALayer`/`DenseLayer.Forward` → primitives | P0/P1 | **ooples/AiDotNet** | ⏳ tracked in ooples/AiDotNet#1447 |
> | `AiModelBuilder<MultipleRegression>` 48k overhead | P2 | **ooples/AiDotNet** | ⏳ tracked in ooples/AiDotNet#1447 |
> | Fused **backward** for LSTM/MHA (training path) | P1 | Tensors | ❌ forward-only today; follow-up |
> | Gradient-step (backward+optimizer) perf | P1 | both | ❌ gated on #433 |
>
> **Why #437 didn't auto-close this:** #436 is a multi-repo umbrella; #437 intentionally closed only the tensors-side primitives (see its "What this PR does not close" section). The framework-side wiring is the remaining blocker for the AIsEval rerun to show the gaps closed.
>
> ---
>
## Context

[...]

Issue #433, opened 2026-05-22
perf: tape backward + optimizer is 15x forward on 25M-param Paraformer (Persistent-tape cached-chain fast-path under-delivers)


ooples:
## Summary

Profiled a paper-faithful SenseVoice-Small (Du et al. 2024 — 12 Conformer encoder layers + 6 non-AR decoder layers, 25.4 M params, 176 layers total) running a single training step at `[1, 64, 32]` input through `[1, 64, 25000]` output on the current main branch. The backward + optimizer phase consumes **94% of the training step wall time** at steady state. Net: 100 memorization iters (the default for `LossStrictlyDecreasesOnMemorizationTask` invariant) project to ~460 s vs the 180 s xUnit timeout budget.

## Profile numbers

CPU-only (`AiDotNetEngine.ResetToCpu()`), `float`, Adam optimizer, MSE loss, fresh master + AiDotNet#1421's `BN→LN` fix:

```
Ctor + InitializeLayers:        49 ms
Predict #1 (lazy init):       2269 ms
Predict #2 (warm):             304 ms
Predict avg (5 iters):         292 ms
ForwardForTraining avg (5):    273 ms (= forward inside Train)

Train #1 (warm-up):           5616 ms
Train #2:                     4596 ms
Train #3:                     4909 ms
Train #4:                     4373 ms

=> Backward + optimizer ≈ Train − ForwardForTraining ≈ 4500 − 273 ≈ 4227 ms (94% of step)
```

For reference, PyTorch's typical backward/forward ratio on CPU is 1.5×–2×. The observed 15×–18× ratio here suggests a structural overhead in the tape backward dispatch.

## Suspected gaps

Per the `NeuralNetworkBase.TrainWithTape` comments at line 5553:
[...]

PR #437, opened 2026-05-22
perf(#436): tensors-side fixes for AIsEval gaps — OpenCL GPU-only init, CNN perf gate, fused LstmSequence + MHA primitives

ooples:
## Summary

Closes the tensors-side items from the AIsEval-surfaced perf gap list in #436. Four stages, each independently testable:

| Stage | Item | Status |
|---|---|---|
| 1 | OpenCL backend skips init on CPU-only systems | done |
| 2 | CNN bs=128 inference perf gate (preserves the lead) | done |
| 3 | Fused \`LstmSequenceForward\` primitive | done |
| 4 | Fused \`MultiHeadAttentionForward\` primitive | done |

## Stage 1 — OpenCL only initializes when a real GPU is present

The OpenCL backend's constructor eagerly compiles ~591 kernels (Spectral, spatial transformer, locally-connected, deformable conv, LSTM sequence, GRU sequence, capsule, broadcast, gated activation, fused linear, fused-advanced, IoU loss, softmax variant, complex, parity-210, detection, geometry, RoI, audio, linalg, CLBlast direct + packing). On the AIsEval CPU-only Windows 11 rig this added 2-3 GB to process RSS for no benefit — the legacy availability gate (\`OpenClNativeBindings.IsAvailable\`) returns true whenever any OpenCL ICD is installed, including CPU runtimes (Intel CPU OpenCL, AMD APP CPU, POCL).

- New \`OpenClNativeBindings.IsGpuAvailable\` enumerates platforms, calls \`clGetDeviceIDs(p, CL_DEVICE_TYPE_GPU, ...)\` per platform, returns true only when at least one GPU exists. Honors \`AIDOTNET_DISABLE_OPENCL=1\` env var for manual opt-out.
- \`DirectOpenClContext.IsGpuAvailable\` + \`OpenClBackend.IsGpuAvailable\` forwarders.
- \`OpenClBackend\` ctor gate switched from \`IsAvailable\` to \`IsGpuAvailable\` — kernel-compile path skipped on CPU-only systems.
- \`DirectGpuEngine.TryCreateOpenClBackend\` short-circuits before \`new OpenClBackend()\` to skip native-binding probe.
- \`OpenClNativeBindings.IsAvailable\` diagnostics: \`Console.WriteLine\` → \`Trace.WriteLine\` (stops spamming test stdout).

\`IsAvailable\` semantics unchanged for callers that legitimately want \"any OpenCL ICD\".

## Stage 2 — AIsEval CNN perf regression gate

The fair-comparison rerun showed AiDotNet currently wins CNN inference at bs=128: **6.21 ms vs PyTorch's 41.42 ms** (6.7× faster). That lead is the only \"AiDotNet faster than PyTorch\" datapoint in the AIsEval numbers and the only \"preserve\" item in #436.

Two manual perf gates in \`PerformanceRegressionTests\` ([Fact(Skip = ...)] like the existing ones):
- \`Conv2D_AiseValCnnL1_Forward_PreservesPyTorchLead\` — guards \`[128, 1, 28, 28]\` @ \`[16, 1, 3, 3]\` pad=1 at <= 30 ms.
- \`Conv2D_AiseValCnnL2_Forward_PreservesPyTorchLead\` — guards \`[128, 16, 14, 14]\` @ \`[32, 16, 3, 3]\` pad=1 at <= 35 ms.

Run via \`dotnet test --filter PerformanceRegression\`.

## Stage 3 — Fused \`LstmSequenceForward<T>\` (P0)

The AIsEval LSTM \`Predict()\` did NOT finish in 3+ minutes at PyTorch-default workload (\`[B=128, seq=32, in=32, hidden=64]\`). PyTorch's \`nn.LSTM\` finished the same workload in 11.76 ms. Root cause was not the GEMM compute — it was the per-timestep loop in \`AiDotNet.NeuralNetworks.Layers.LSTMLayer.Forward\` spawning 32 separate \`Tensor<T>.MatMul + Sigmoid + Tanh\` dispatches per call, each recording on the autograd tape and allocating scratch tensors.

New \`CpuEngine.LstmSequenceForward<T>\`:
- Pre-computes \`Wx = input @ W_ih^T\` for ALL timesteps as one [B*seq, 4H] GEMM via \`TensorMatMulTransposed\` (routes through \`SimdGemm.Sgemm\` with transB=true for float).
- Folds \`b_ih\` into Wx in-place.
- Per-step tight inner loop over raw \`Span<T>\` for \`h_prev @ W_hh^T\` + 4-way gate split + sigmoid/tanh/c-update/h-update.
- Forward-only (throws on active GraphMode — training paths keep the decomposed \`LSTMLayer.Forward\` until backward fuses too).
- PyTorch gate order (i, f, g, o) so the framework can adopt without weight permutation.

**Result:** At the AIsEval shape, fused primitive finishes in **44 ms/iter (net10.0) / 74 ms/iter (net471)** vs the per-step path that didn't finish in 3+ minutes. Still ~4× behind cuDNN/MKL nn.LSTM, but the \"doesn't finish\" state is closed.

Correctness verified against a from-scratch scalar-loop reference in 4 tests (small shapes, returnSequences true/false, custom h0/c0, shape validation). Tolerance 1e-4.

## Stage 4 — Fused \`MultiHeadAttentionForward<T>\` (P0)

Transformer \`Predict()\` at bs=128 was 16.8× slower than PyTorch (233 ms vs 13.85 ms) — 5+ separate \`Tensor<T>.MatMul\` + reshape + \`Transpose\` + \`ScaledDotProductAttention\` calls in the framework's MHA layer, each entering GraphMode-aware recording paths and allocating intermediates.

New \`CpuEngine.MultiHeadAttentionForward<T>\` wraps the standard MHA chain:
- Q/K/V projections via 3 GEMMs on flattened \`[B*seq, dModel]\` input.
- Reshape + transpose (0,2,1,3) to \`[B, heads, seq, dHead]\`.
- Existing \`ScaledDotProductAttention<T>\` (no new attention math).
- Inverse-transpose + reshape + output projection.
- Throws on active GraphMode.

Same compute as the explicit chain — the win is 5 dispatches → 1 (no per-call autograd-tape entry, no per-op scratch).

Correctness verified against the explicit chain inline in the test (the chain documented in the wrapper's XML doc). Shape validation tested.

## What this PR does **not** close (and why)

- **MLP \`Predict()\` @ bs=128 (P1 — 4.7× slower)**: The underlying primitives needed to close this (\`FusedLinear\`, \`Avx512Sgemm\`) already exist in this repo. The fix is for \`AiDotNet.NeuralNetworks.Layers.DenseLayer.Forward\` to call \`FusedLinear\` instead of \`MatMul + Add + Activation\`. **That's a framework PR**, not tensors.

- **Gradient-step time (P1 — 3-9× slower)**: Overlaps with #433 (tape backward + optimizer perf on 25M-param Paraformer) which is in flight. Not duplicating effort here.

- **\`AiModelBuilder<MultipleRegression>\` at 48k rows (P2 — 4.3× slower)**: Lives entirely in \`ooples/AiDotNet\` (the framework repo); the underlying tensor ops aren't the bottleneck. Needs a separate framework-side PR to reduce builder overhead / split-phase compute.

- **MLP/Transformer/LSTM inference perf wins from the new primitives**: This PR ships the primitives but does NOT yet wire them into \`LSTMLayer.Forward\` / \`MultiHeadAttentionLayer.Forward\` / \`DenseLayer.Forward\` — those live in the framework repo. A follow-up framework PR will switch the inference paths to call the new primitives, at which point an AIsEval rerun should show the gaps closed.

- **Fused backward for LstmSequence + MHA**: Forward-only primitives in this PR. Backward is a follow-up — training paths keep the existing decomposed (tape-recorded) chain in the meantime.

## Test plan

- [x] \`dotnet build src/AiDotNet.Tensors -c Release\` succeeds.
- [x] \`dotnet test --no-build -c Release --filter \"FullyQualifiedName~LstmSequenceForward|FullyQualifiedName~MultiHeadAttentionForward|FullyQualifiedName~Conv2D_Aiseval\"\` — 6 passed, 3 skipped (manual perf gates), 0 failed on both net10.0 and net471.
- [ ] After framework PR lands: re-run AIsEval scaffold (\`dotnet run -- --models lstm,transformer\`) and confirm LSTM finishes and Transformer bs=128 latency is within 2× PyTorch.

## Stages as commits

Each stage is a single commit so they can be reviewed independently:

\`\`\`
9f636bc4 perf(#436): Stage 1 — OpenCL backend only initializes when a real GPU is present
337e8e24 perf(#436): Stage 2 — AIsEval CNN inference perf gate
bac4ee77 perf(#436): Stage 3 — fused LstmSequenceForward primitive (closes P0 LSTM gap)
e474fe76 perf(#436): Stage 4 — fused MultiHeadAttentionForward primitive (P0 Transformer gap)
978585c6 test(#436): match repo convention for perf-gate tests (skip-by-default)
\`\`\`


ooples:
## Stages 5-7 added (additional perf wins beyond the original PR scope)

Per follow-up review that we were still substantially behind PyTorch after Stages 1-4 — three more commits land actual compute speedups on top of the new primitives:

### Stage 5: float fast path for `LstmSequenceForward`

The Stage 3 primitive used a generic-T inner loop with `INumericOperations<T>` for sigmoid/tanh per element. Replaced with a float-specialized path that:
- Calls `SimdGemm.Sgemm` directly for the Wx pre-compute AND per-step h_prev @ wHh^T
- Pulls all scratch (wx, h_curr, c_curr, h_prev, c_prev, hh, gate buffers) from `ArrayPool<float>.Shared` once at the top — zero per-step allocation
- Reuses scratch via tuple-swap of array references at end of each step
- Uses `SimdKernels.SigmoidUnsafe` / `TanhUnsafe` for vectorized AVX2/AVX-512 gate activations at `[B, hidden]` granularity per gate per batch

**Result on AIsEval shape `[B=128, seq=32, in=32, hidden=64]`:**

| Stage | net10.0 | net471 | vs PyTorch |
|---|---|---|---|
| Pre-PR (per-step loop in framework) | >3 min (didn't finish) | n/a | infinite |
| Stage 3 (generic-T fused) | 44.44 ms | 73.76 ms | 4x slower |
| **Stage 5 (float fast path)** | **8.19 ms** | 56.58 ms | **30% FASTER than PyTorch's 11.76 ms** |

### Stage 6: float fast path for `MultiHeadAttentionForward`

The Stage 4 wrapper went through 4 `TensorMatMul` dispatches and 4 strided-transpose materializations. Replaced with a path that:
- Calls `SimdGemm.Sgemm` directly for all four projections (Q/K/V + output)
- Writes the `[B*seq, dModel] -> [B, heads, seq, dHead]` transposes as explicit nested-loop block copies of the `dHead`-element innermost stride (one O(B*seq*dModel) pass per matrix via `Span<T>.CopyTo` -> Buffer.MemoryCopy/SIMD)
- All scratch is pooled, lifetimes bounded by try/finally
- Existing `ScaledDotProductAttention<float>` primitive is reused (already optimized) but called against pooled-buffer-backed `Tensor<float>` wrappers so no Tensor allocation happens for the intermediates

**AIsEval MHA shape `[B=128, seq=32, dModel=64, heads=4]`: 22.31 ms → 10.13 ms on net10.0 (2.2x).**

### Stage 7: parallelize transposes + lock in the wins

- `Parallel.For` across batch in `TransposeQkv` / `InverseTransposeQkv`. Each batch element writes disjoint memory regions so no synchronization needed.
- Two new perf regression gates in `PerformanceRegressionTests`:
  - `LstmSequenceForward_Aiseval_FloatFastPath_BeatsPyTorch` (budget 30 ms, measured 8.19 ms)
  - `MultiHeadAttentionForward_Aiseval_FloatFastPath` (budget 30 ms, measured 8.50 ms)

**Final MHA: 8.50 ms — 2.6x total speedup vs Stage 4.**

## Cumulative AIsEval status (after this PR + framework adoption)

| Family | PyTorch | AiDotNet primitive | Verdict |
|---|---|---|---|
| MLP @ bs=128 | 1.91 ms | (existing FusedLinear — needs framework adoption) | unchanged in this PR |
| **CNN @ bs=128** | 41.42 ms | 6.21 ms | **AiDotNet wins (existing, regression-gated)** |
| **LSTM @ bs=128** | 11.76 ms | **8.19 ms** | **AiDotNet wins** (was: didn't finish) |
| Transformer MHA @ bs=128 | ~7 ms est | 8.50 ms | within 1.2x (was: 233 ms full layer) |
| OpenCL RSS on CPU-only | 282 MB | now skipped entirely | closed |

The "we are still substantially behind" gap is now: LSTM and MHA are competitive or winning at the primitive level. The remaining work is the framework PR (LSTMLayer / MultiHeadAttentionLayer / TransformerEncoderLayer / DenseLayer adopting these primitives in their inference paths).

## Test status

- Correctness: 6 of 6 functional tests pass on both net10.0 and net471 (LSTM 4, MHA 2).
- Perf gates: 5 of 5 `[Fact(Skip = ...)]` (CNN L1+L2, LSTM, MHA, plus the original `LstmSequenceForward_AisevalShape_FinishesWithinBudget`) — skipped in default CI sweep, run manually via `--filter PerformanceRegression` or `--filter Aiseval`.

PR #447, opened 2026-05-26
fix(#364): correct 6 gfx1012 GPU kernel bugs + auto GPU-vs-CPU correctness harness

### LINKED ISSUE #364: DirectGpu OpenCL gemm_small + CLBlast-direct kernels produce numerically wrong results on AMD gfx1012 (RDNA1 Navi 14)
## Summary

`AiDotNet.Tensors.Engines.DirectGpuTensorEngine.TensorMatMul<float>` produces **numerically incorrect** results on AMD `gfx1012:xnack-` (Radeon RX 5500 / RDNA1 Navi 14) on Windows. The bug reproduces on both kernel paths the dispatcher selects: `gemm_small` for `M<128` (or `N<128` or `K<64`), and CLBlast-direct / `gemm_double_buffered` for `M,N>=128, K>=64`.

This silently corrupts `model.Train` / `model.Predict` for any HarmonicEngine consumer who turns on `AiModelBuilder.ConfigureGpuAcceleration()` — the marketing copy promises 10-100x speedups with no code change, but on this hardware the GPU returns garbage. Filed at upstream P0 against the `ConfigureGpuAcceleration` user contract.

## Environment

| field | value |
|---|---|
| AiDotNet | 0.198.0 |
| AiDotNet.Tensors | 0.81.0 |
| AiDotNet.Native.OpenBLAS | 0.81.0 |
| OS | Windows 10.0.26220 |
| Framework | .NET 10.0.8 |
| GPU | AMD Radeon RX 5500 (`gfx1012:xnack-`, RDNA1 Navi 14) |
| OpenCL runtime | Available (Native GPU libs: OpenCL=True) |
| CLBlast native | NOT loaded (`HasClBlast=False`) — so the suspect path is the in-tree `CLBlast direct kernels` (compiled from `.cl` source at startup, NOT the native CLBlast library) |
| CPU | x64 32-thread, AVX2+FMA, no AVX-512 |
| Autotune fingerprint | `x64-amd-avx2-cpu32` |
| Default deterministic mode | True (set by `AiModelBuilder.BuildAsync`; doesn't gate the DirectGpu path) |

`AiDotNet.Diagnostics.AccelerationDiagnostics.GetReport()` at startup:

```
=== AiDotNet Acceleration Report ===
  Engine:           Direct GPU Engine (OpenCL gfx1012:xnack-)
  Deterministic:    True
  ...
  GPU backends:     CUDA=False OpenCL=False HIP=False
  Native BLAS:      OpenBLAS=False CLBlast=False MKL=False CpuBLAS=False
  Native GPU libs:  CUDA=False HIP=False OpenCL=True
```

(Diagnostic divergence: `PlatformCapabilities.HasOpenCLSupport=False` while `NativeLibraryDetector.Status.HasOpenCl=True` — both come from `AccelerationDiagnostics`. Probably a separate filed issue, but worth flagging: the two probes disagree about whether OpenCL is available even though the engine is sitting on a fully-loaded `DirectGpuTensorEngine`.)

## Reproducer

```csharp
using AiDotNet.Tensors.Engines;
using AiDotNet.Tensors.LinearAlgebra;

// Build deterministic [64,64] x [64,64] float tensors.
var rng = new Random(42);
var a = new Tensor<float>([64, 64]);
var b = new Tensor<float>([64, 64]);
for (int i = 0; i < 64; i++)
    for (int j = 0; j < 64; j++) {
        a[i, j] = (float)(rng.NextDouble() - 0.5);
        b[i, j] = (float)(rng.NextDouble() - 0.5);
    }

// CPU baseline.
AiDotNetEngine.ResetToCpu();
var cpuResult = AiDotNetEngine.Current.TensorMatMul(a, b);
// cpuResult[0,0] ~ 0.458285

// DirectGpu.
using var gpuEngine = new DirectGpuTensorEngine();
Console.WriteLine($"isAvailable={gpuEngine.IsGpuAvailable}, name={gpuEngine.Name}");
// isAvailable=True, name=Direct GPU Engine (OpenCL gfx1012:xnack-)

AiDotNetEngine.Current = gpuEngine;
var gpuResult = AiDotNetEngine.Current.TensorMatMul(a, b);
_ = gpuResult.AsSpan();  // force deferred download to fire

Console.WriteLine($"cpu[0,0]={cpuResult[0,0]:G6}  gpu[0,0]={gpuResult[0,0]:G6}");
// cpu[0,0]=0.458285  gpu[0,0]=0.160251  (varies seed-to-seed but consistent within a run)

double maxErr = 0;
for (int i = 0; i < 64; i++)
    for (int j = 0; j < 64; j++) {
        double e = Math.Abs((double)cpuResult[i,j] - (double)gpuResult[i,j]);
        if (e > maxErr) maxErr = e;
    }
Console.WriteLine($"max_abs_err = {maxErr:G6}");  // ~ 3.18
```

### Phase 2b — `[64,64] x [64,64]` (`gemm_small` kernel path)

```
Pre-materialize gpuResult[0,0] = 0.160251
Post-AsSpan() gpuResult[0,0]   = 0.160251
GPU TensorMatMul: 89.11ms, max_abs_err vs CPU = 3.17851

Top 5 abs-error positions (i, j, gpu, cpu, |err|):
  [31,32]: gpu=0.0984512 cpu=-3.08006 err=3.17851
  [31,15]: gpu=0.402722  cpu=-2.12435 err=2.52707
  [30,53]: gpu=-0.126621 cpu=2.35923  err=2.48586
  [15,12]: gpu=-0.0684334 cpu=2.28911 err=2.35754
  [44,10]: gpu=-0.13121  cpu=2.21485  err=2.34606

GPU result first row (8 entries): 0.160251, 0.0951236, -0.208234, 0.221463, -0.103071, -0.0288194, 0.142358, -0.0555886
CPU result first row (8 entries): 0.458285, -0.4876, -0.107915, 0.139458, -0.683643, -0.10098, -0.843341, 0.527958
```

GPU values are roughly 3x smaller in magnitude than CPU — not random noise, not uninitialized memory, but a **systematic mis-accumulation**.

### Phase 2c — `[128,128] x [128,128]` (CLBlast-direct path)

```
Phase 2c: 128x128 GPU matmul max_abs_err vs CPU = inf
GPU [0,0] = -2.37012E+37  CPU [0,0] = -1.87171
```

Different kernel path, same broken outcome. The values overflow to ~1e+37 here because the broken accumulation now runs over 128 K elements instead of 64.

## What is NOT the cause

- **Not uninitialized memory** — values are deterministic across runs (same `0.160251` at `[0,0]` on rerun with same seed).
- **Not a missing deferred-download** — the `DeferredArrayMaterializer` registers correctly and fires on first indexer access (verified by reading `gpuResult[0,0]` both before and after explicit `AsSpan()` — identical result).
- **Not a `DownloadBuffer` sync issue** — `clEnqueueReadBuffer` is blocking, so any kernel-completion race would surface as garbage that varies in magnitude, which is not what we see.
- **Not a CPU-side bug** — Phase 2a confirms `CpuEngine.TensorMatMul` agrees with the hand-written reference to `max_abs_err = 5.96E-07`.

## Likely root cause hypotheses (for triage)

1. **AMD RDNA1 driver bug specific to `gfx1012`** — this is an underserved consumer GPU; the OpenCL kernel JIT-compile may produce a buggy SPIR-V/AMDGCN binary on this exact arch. Could be tested by another `gfx10xx` GPU (5700XT, 6700XT) — if those work, narrow to gfx1012.
2. **Kernel arg packing for `__kernel void gemm_small`** — the kernel source `for (k=0; k<K; k++) sum = fma(A[row*K+k], B[k*N+col], sum);` is correct, so the bug is upstream of the kernel (host-side arg setup). Possible: `kernel.SetArg(3, M)` / `(4, N)` / `(5, K)` are being passed in the wrong order, or `alpha`/`beta` are being uploaded as wrong types (int vs float). The values being roughly the right magnitude (just 3x smaller) is suspicious — would happen if `K=21` got passed instead of `K=64` (which would explain 1/3 accumulation, 1/3 magnitude).
3. **Buffer upload row-major vs column-major mismatch** — the upload via `backend.AllocateBuffer(floatData)` may be using a layout the kernel doesn't expect. CPU uses row-major; if the kernel expects column-major (e.g., FORTRAN convention) and reads `A[row*K+k]`, the resulting matrix is computed against a transposed input. But the values being smaller in magnitude (not just permuted) argues against this.

The 3x magnitude difference makes hypothesis 2 the most likely culprit (`K` being passed as a fraction of its true value).

## Severity

P0 against the `AiModelBuilder.ConfigureGpuAcceleration()` user contract. Any HE-consumer who turns on GPU acceleration on RDNA1 hardware will get a model that trains on incorrect gradients (silently). The `Aggressive` usage level explicitly promises maximum speedup with no fallback — that fails very loudly here.

## Workaround (HE side, today)

- Default `useGpu=false` on this host.
- `AIDOTNET_DIRECTGPU_BACKENDS=cpu` env var to force CPU through the module-init auto-detect.
- Call `AiDotNetEngine.ResetToCpu()` in test setup before any tensor op.

## Reference

- Test: `tests/Learning/Phase_PAPER_A_PathB_GpuAcceleration_Sanity.cs` in `ooples/HarmonicEngine` on branch `research/ledger-overhaul-and-l4-reification`
- Mechanism analysis: `docs/ledger/mechanism_failures.md` ("GPU acceleration diagnostic audit + DirectGpu OpenCL kernel correctness failure — Phase PAPER-A Path B (2026-05-17)")

ooples:
## Summary

Fixes #364. Three distinct numerical-correctness bugs in the DirectGpu (OpenCL) backend on AMD gfx1012 (RDNA1, RX 5500 XT), plus a comprehensive GPU-vs-CPU correctness test suite so these can't silently regress for any core kernel again.

All three were caught/verified by pinning `DirectGpuTensorEngine` output to the trusted `CpuEngine` reference across a thorough shape sweep. The fixes keep the **fast** kernels (register-tiled XgemmDirect, packed indirect Xgemm, vectorized softmax) — no naive bypasses.

### 1. XgemmDirect (direct path, M/N < 448) — invalid autotune VWN
The gfx1012 entry in `ClBlastXgemmDirectDatabase.Generated.cs` carried `VWN=4`. CLBlast requires `VWN` to divide `WGD/NDIMC` and `WGD/NDIMB`; with `WGD=32`, `NDIMBD=16` that quotient is `2`, so `VWN=4` made the float4 B-tile load read misaligned/out-of-bounds → wrong GEMM (64×64 ~3× off / NaN / 1e37). **Fix: `VWN 4 → 2`.**

### 2. Indirect packed-Xgemm (M/N ≥ 448) — missing A-operand transpose
The CLBlast `Xgemm` kernel reads its **A operand column-major** (`agm[idm + idk*ld]`) but its **B operand row-major** (`bgm[idk*ld + idn]`). The in-tree row-major-swap reinterpret trick (row-major X == column-major Xᵀ) is valid for the A slot (`agm = original B`, as-is) but **not** for the B slot: the kernel needs `bgm[k*ld + n] == A[n,k]`, i.e. the physical transpose **Aᵀ**. The code copied A *un-transposed* (and skipped the copy entirely when no padding was needed), so the kernel contracted the wrong axis → constant/garbage output for any non-symmetric input ≥ MinIndirectSize. (Symmetric inputs like all-ones masked it.) **Fix: always materialize the B operand as Aᵀ via `ClBlastTransposeMatrix`** (the surrounding code already computed the correct transpose dims).

Verified: a row-structured input (`A[i,k]=i, B=1`, true `C[i,j]=i·512`) returned the constant `Σ_m m = 130816` for *every* element before; exact after.

### 3. Softmax — float4 tail loop skipped the final element at non-aligned widths
The vectorized softmax covers `[0,(features/4)*4)`; the scalar tail seeded its start from each thread's *per-thread* leftover index (`f/4*4 + lid%localSize`) instead of the uniform `(features/4)*4 + lid`. For widths not divisible by 4 (e.g. 129) this evaluated past the end for every thread, so the final element was never visited in any phase — excluded from the max and sum and left as uninitialized garbage, and the row no longer summed to 1. **Fix: uniform tail start in all three phases.**

## Test suite (`GpuCpuCorrectnessTests`, 202 cases)
- `TensorMatMul`: 23 shapes — <128 (direct hot zone), tile boundaries around WGD=32, exact/off-by-one multiples, non-square, degenerate 1-dims, indirect path (M/N ≥ 448).
- `TensorMatMulTransposed`, `BatchMatMul` ([B,M,K]×[B,K,N]).
- Elementwise add/sub/mul/divide/divide-scalar, transpose.
- Reductions: ReduceSum/ReduceMax/ReduceMean (last axis).
- Softmax + 9 activations (ReLU, LeakyReLU, GELU, Sigmoid, Tanh, Swish, Mish, ELU, Softplus) + TensorExp.
- Sizes deliberately include non-multiples-of-4 (127, 129, 1023, 1025) to exercise every float4-vectorized kernel's scalar tail — the exact class of bug #3.

Each case returns early on CPU-only machines, so the suite stays portable. On gfx1012: **202/202 pass** (was 3 failing — 448/512 matmul and softmax 127×129 — before the fixes).

> Disclaimer: no performance/Pareto claims here — this PR is correctness-only. Verified on AMD gfx1012/RDNA1 (RX 5500 XT), .NET 10, Win x64.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/ClBlastXgemmDirectDatabase.Generated.cs
+++ b/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/ClBlastXgemmDirectDatabase.Generated.cs
@@ -56,12 +56,15 @@ internal static class ClBlastXgemmDirectDatabaseData
             }),
             // RX 5500 XT (gfx1012) - RDNA1 with 11 CUs, optimized for smaller workgroups
             // PadA=4, PadB=4 for 32-bank LDS conflict avoidance (indices 5,6)
-            // VWM=4, VWN=4 for wider vectorization (float4 instead of float2) - indices 7,8
+            // VWM=4 (valid: WGD/MDIMCD=WGD/MDIMAD=32/8=4, 4%4==0). VWN=2 (indices 7,8).
+            // FIX #364: VWN was 4 but is INVALID here — CLBlast requires VWN | (WGD/NDIMCD) and
+            // VWN | (WGD/NDIMBD); WGD/NDIMBD = 32/16 = 2, so VWN must divide 2 (i.e. <=2). VWN=4 made the
+            // float4 B-tile load read misaligned/OOB data -> numerically wrong GEMM (64x64 ~3x off / NaN).
             // KWID=2 (standard K-loop depth; higher KWID values cause register pressure on 11-CU RDNA1)
             new ClBlastArchitectureEntry("gfx1012:xnack-", new[]
             {
-                new ClBlastDeviceEntry("AMD Radeon RX 5500 XT", new short[] { 2, 8, 8, 16, 16, 4, 4, 4, 4, 32, 0, 0, 0, 0, 0, 0 }),
-                new ClBlastDeviceEntry("default", new short[] { 2, 8, 8, 16, 16, 4, 4, 4, 4, 32, 0, 0, 0, 0, 0, 0 }),
+                new ClBlastDeviceEntry("AMD Radeon RX 5500 XT", new short[] { 2, 8, 8, 16, 16, 4, 4, 4, 2, 32, 0, 0, 0, 0, 0, 0 }),
+                new ClBlastDeviceEntry("default", new short[] { 2, 8, 8, 16, 16, 4, 4, 4, 2, 32, 0, 0, 0, 0, 0, 0 }),
             }),
             new ClBlastArchitectureEntry("gfx1030", new[]
             {
--- a/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/Kernels/ActivationKernels.cs
+++ b/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/Kernels/ActivationKernels.cs
@@ -209,7 +209,7 @@ __kernel void softmax(
         float4 v = vload4(0, rowIn + f);
         threadMax = fmax(threadMax, fmax(fmax(v.x, v.y), fmax(v.z, v.w)));
     }
-    for (int i = f / 4 * 4 + lid % (localSize); i < features; i += localSize) {
+    for (int i = (features / 4) * 4 + lid; i < features; i += localSize) {
         if (i < features) threadMax = fmax(threadMax, rowIn[i]);
     }
 
@@ -232,7 +232,7 @@ __kernel void softmax(
         vstore4(e, 0, rowOut + f);
         threadSum += e.x + e.y + e.z + e.w;
     }
-    for (int i = f / 4 * 4 + lid % (localSize); i < features; i += localSize) {
+    for (int i = (features / 4) * 4 + lid; i < features; i += localSize) {
         if (i < features) { float e = fast_exp1(rowIn[i] - rowMax); rowOut[i] = e; threadSum += e; }
     }
 
@@ -252,7 +252,7 @@ __kernel void softmax(
     for (; f + 3 < features; f += localSize * 4) {
         vstore4(vload4(0, rowOut + f) * invSum, 0, rowOut + f);
     }
-    for (int i = f / 4 * 4 + lid % (localSize); i < features; i += localSize) {
+    for (int i = (features / 4) * 4 + lid; i < features; i += localSize) {
         if (i < features) rowOut[i] *= invSum;
     }
 }
--- a/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/Kernels/NeuralNetKernels.cs
+++ b/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/Kernels/NeuralNetKernels.cs
@@ -1333,7 +1333,7 @@ __kernel void gather_kernel(
 // scatter_add_kernel_deterministic below.
 __kernel void scatter_add_kernel(
     __global const float* source,
-    __global const float* indices,
+    __global const int* indices,
     __global float* destination,
     const int sourceSize,
     const int featureSize)
@@ -1341,7 +1341,7 @@ __kernel void scatter_add_kernel(
     const int idx = get_global_id(0);
     if (idx >= sourceSize) return;
 
-    int destIdx = (int)indices[idx];
+    int destIdx = indices[idx];
     for (int f = 0; f < featureSize; f++) {
         // Use atomic add for thread safety when multiple sources scatter to same destination
         atomic_add_float(&destination[destIdx * featureSize + f], source[idx * featureSize + f]);
@@ -1352,7 +1352,7 @@ __kernel void scatter_add_kernel(
 // One work-item per (dstRow, f) cell; scans source rows in fixed ascending order.
 __kernel void scatter_add_kernel_deterministic(
     __global const float* source,
-    __global const float* indices,
+    __global const int* indices,
     __global float* destination,
     const int sourceSize,
     const int destSize,
@@ -1364,7 +1364,7 @@ __kernel void scatter_add_kernel_deterministic(
 
     float sum = 0.0f;
     for (int i = 0; i < sourceSize; i++) {
-        if ((int)indices[i] == dstRow) {
+        if (indices[i] == dstRow) {
             sum += source[i * featureSize + f];
         }
     }
--- a/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/OpenClBackend.cs
+++ b/src/AiDotNet.Tensors/Engines/DirectGpu/OpenCL/OpenClBackend.cs
@@ -1752,16 +1752,25 @@ namespace AiDotNet.Tensors.Engines.DirectGpu.OpenCL
                         if (timingEnabled) { Synchronize(); packATime = sw!.ElapsedTicks; }
                     }
 
-                    // Pad "B" (originally A) if needed - NO TRANSPOSE, just copy with padding
-                    if (bNeedsPad)
+                    // Materialize "B" (originally A) as A^T — ALWAYS, even when no padding.
+                    //
+                    // The CLBlast Xgemm kernel's two operands use DIFFERENT load conventions:
+                    // it reads its A operand column-major (agm[idm + idk*ld]) but its B operand
+                    // row-major/"transposed" (bgm[idk*ld + idn]). The row-major-swap reinterpret
+                    // trick (row-major X == column-major X^T) therefore works for the A slot
+                    // (agm = original B, passed as-is) but NOT for the B slot: the kernel needs
+                    // bgm[k*ld + n] == A[n,k], i.e. the PHYSICAL transpose A^T (K×M, ld=nCeiled).
+                    // The previous code copied A un-transposed (and skipped the copy entirely when
+                    // no padding was needed), so the kernel contracted the wrong axis and produced
+                    // wrong results for every non-symmetric input >= MinIndirectSize (symmetric
+                    // inputs like all-ones masked the bug). Transposing A fixes it. (#364 follow-up)
                     {
                         if (timingEnabled) { sw!.Restart(); }
                         bTemp = AllocateBuffer((int)bSize);
                         if (timingEnabled) { allocTime += sw!.ElapsedTicks; sw.Restart(); }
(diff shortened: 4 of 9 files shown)


ooples:
## Update: reflection-driven auto-differential harness (covers *every* GPU kernel)

Added `GpuCpuAutoDifferentialTests` so accuracy coverage can never silently regress for any operation — the GPU analogue of the CPU-side invariant tests:

- **Auto-discovery**: reflects over every method `DirectGpuTensorEngine` overrides from its `CpuEngine` base (= every GPU kernel with a CPU reference) and runs each against CPU. No per-op maintenance — new kernels are picked up automatically.
- **Generic input generator** drives unary/binary/scalar ops plus `int`/`int[]` (axes, stride, kernel), `Tensor<int>`/`Tensor<bool>`/`Tensor<Complex>`, enums, nullables, tensor-arrays, and `out`-params. ~155 kernels get an automatic GPU-vs-CPU check (incl. LayerNorm/GroupNorm/attention, FFT-by-component, concat/stack, IoU).
- **Coverage guard** (`EveryGpuKernel_IsAutoTestedOrAllowlisted`) runs CPU-only (CI-portable, no GPU): every override must be auto-testable OR explicitly allowlisted. A new kernel with an undrivable signature **fails the guard** until a dedicated test is added.
- **`DedicatedlyCovered`**: ops needing distinct per-tensor shapes (conv3d/transpose/fold/unfold/roi/gridsample/backward) — flagged for dedicated tests; can't grow silently.
- **`DifferentialExempt`**: ops where random generic input violates the op contract (Erfinv domain, NLLLoss targets, box validity, FFT normalization convention, `[C]`-shaped norm params, 4-D maxpool), each with a reason.

**Harness already surfaced 4 suspected real discrepancies** (valid generic input, GPU≠CPU), tagged `FLAGGED` and explicitly *not* asserted to pass, pending dedicated investigation: `TensorL1Loss`, `TensorHuberLoss`, `TensorLogSumExp` (reduction at width 129 — same class as the softmax tail bug?), `TensorScatterAdd`.

**430/430** GPU correctness + auto-differential cases pass on gfx1012.

ooples:
## Final scope

**6 GPU kernel correctness bugs fixed on gfx1012/RDNA1** (all proven against the CpuEngine reference):

1. **XgemmDirect** — invalid autotune `VWN=4 → 2` (direct GEMM, M/N<448).
2. **Indirect packed-Xgemm** — B operand must be physically `Aᵀ`; was un-transposed → wrong contraction axis → constant/garbage for non-symmetric inputs ≥448.
3. **Softmax** — float4 tail skipped the final element at non-÷4 widths (e.g. 129).
4. **ScatterAdd** — kernel read `indices` as `float*` but receives an `int` buffer → all updates dumped into index 0.
5. **L1Loss / HuberLoss** — returned a per-row vector + read a deferred (un-materialized) buffer → zeros; now reduce to the scalar mean (CpuEngine contract).
6. **LogSumExp** — GPU path assumed a contiguous last axis → garbage on interior axes; now restricted to the last axis with CPU fallback.

Bugs 4–6 were surfaced by the reflection-driven harness itself.

**Two test suites (430/430 pass on gfx1012):**
- `GpuCpuCorrectnessTests` — hand-written shape sweeps (matmul tile boundaries, non-aligned softmax widths, etc.).
- `GpuCpuAutoDifferentialTests` — auto-discovers every GPU-kernel override and diffs vs CPU, with a coverage guard so no kernel can ship without an accuracy check.

### CI behavior (no GPU runner)
Designed to work on stock GitHub `ubuntu-latest`:
- The **coverage guard runs CPU-only** → executes in CI on every PR and fails the build if a new kernel is added that isn't auto-tested or explicitly allowlisted. This is the live regression gate.
- The **numeric GPU-vs-CPU comparisons no-op safely when no GPU is present** (early-return), so they don't fail CI; they run for real on GPU-equipped machines (verified here on gfx1012). A GPU CI job can be added later if/when a self-hosted runner is available.

`DifferentialExempt` documents the ops where generic random input is semantically invalid (Erfinv domain, FFT convention, `[C]`-shaped norm params, etc.) and the one precision exemption (LogSumExp multi-row composite, per-row verified exact). `DedicatedlyCovered` lists ops needing distinct per-tensor shapes (conv3d/roi/gridsample/backward) — flagged, can't grow silently.

PR #449, opened 2026-05-26
test(#443): force Parallel.For fan-out in ParallelForOrSerial dispatch test (unblock CI)

ooples:
## Unblocks CI

`ParallelForOrSerial_LargeWork_DeterministicFalse_FansOutToWorkers` (added by #443, merged 2026-05-26) has been flaking on shared CI. PR #445's run [26429102450](https://github.com/ooples/AiDotNet.Tensors/actions/runs/26429102450) failed with:

> totalWork (131072) >> grain size (32768) with DeterministicReductions=false should have dispatched to the worker pool, but all 4 chunks ran on caller thread 16.

## Root cause

.NET's raw `System.Threading.Tasks.Parallel.For` inlines very short iteration counts on the caller thread when the dispatch overhead would dwarf the per-iteration cost. With `numChunks = Environment.ProcessorCount = 4` and an empty body (the test body just probes thread IDs), TPL legitimately decides to run all four iters inline on the caller. That's a valid Parallel.For optimisation — but the test was written to verify *ParallelForOrSerial*'s **parallel branch** was taken (the dispatch decision), and Parallel.For's small-loop inlining makes the fan-out probe an unreliable proxy for that decision.

The sibling test `LightweightParallel_AtOrAboveGrainSize_FansOutToWorkers` passes reliably because `LightweightParallel` routes through `PersistentParallelExecutor`, which always fans out — not raw `Parallel.For`.

## Fix

- `ParallelForOrSerial`'s actual dispatch decision depends ONLY on `totalWork` vs grain size, `MaxDegreeOfParallelism`, and `DeterministicReductions` — **`numChunks` is independent**. Bump `numChunks` to `max(16, ProcessorCount * 4)` so TPL is forced to spawn tasks (small-loop inlining doesn't kick in at 16+ iters). The production code under test is **unchanged**.
- Lengthen the caller-thread wait from 200 ms to 1 s. The worker `Set()`s the event from the first off-thread iter so the wait is nearly instant on a warm pool; but a cold-pool first-ever-`Parallel.For` on a constrained CI runner can take >200 ms to spawn a worker.

## Verified

- Local run on 16-logical-CPU Windows: test passes in **14 ms**.
- Build clean (0 errors).

## Why a separate PR

PR #445 (licensing) and PR #446 (release pack fix) are blocked by this test failing on the merged CI. Per the project's workflow rule, don't bundle test infrastructure fixes into unrelated feature PRs. Landing this on `main` first lets #445 / #446 / #447 rebase clean.

### DIFF
--- a/tests/AiDotNet.Tensors.Tests/Helpers/ParallelForOrSerialDispatchTests.cs
+++ b/tests/AiDotNet.Tensors.Tests/Helpers/ParallelForOrSerialDispatchTests.cs
@@ -31,20 +31,34 @@ public class ParallelForOrSerialDispatchTests
             CpuParallelSettings.MaxDegreeOfParallelism = Environment.ProcessorCount;
 
             int callerThread = Thread.CurrentThread.ManagedThreadId;
-            int numChunks = Math.Max(2, Environment.ProcessorCount);
+            // The production dispatch decision depends ONLY on totalWork vs
+            // grain size, MaxDegreeOfParallelism, and DeterministicReductions —
+            // numChunks is independent. But .NET's raw Parallel.For inlines
+            // very short iteration counts on the caller thread when the
+            // dispatch overhead would dwarf the per-iter cost (CI run
+            // 26429102450 hit this with 4 iters × empty-body on a 4-vCPU
+            // ubuntu-latest box: all chunks ran on the caller). Use a large
+            // enough iteration count that TPL is forced to fan out, so we're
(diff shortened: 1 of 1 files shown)


ooples:
Superseded by #455 which bundles this fix together with two other CI-stabilization fixes (PrePackSpeedupTest + ConvTranspose2DL2PerfTest). Single merge unblocks the queue.

PR #454, opened 2026-05-26
perf(tensor): SIMD-vectorize BroadcastElementwise + pool result tensors

ooples:
## Summary

Profiling a model-family forward pass with dotnet-trace (ResNet50, float, CPU) found **~45% of CPU self-time in `Tensor<T>.BroadcastElementwise`** — the broadcast used by BatchNorm scale/shift, residual adds, and bias. Root cause:

1. a **per-element `Func<T,T,T>` delegate** (`op(aSpan[aIdx], bSpan[bIdx])`) the JIT cannot vectorize, and
2. **per-element multi-dim coordinate math** (a `%`/`/` rank-loop for *every* output element).

## Changes

**1. SIMD-vectorize `BroadcastElementwise`** (`LinearAlgebra/Tensor.cs`)
- Replace the per-element delegate with a `BroadcastOp` enum.
- **Coalesce trailing dims** into one contiguous inner block (turns `[N,C,H,W] op [1,C,1,1]` into one SIMD run per `(n,c)` instead of a per-element loop).
- Apply the inner block with `System.Numerics.Vector<T>` SIMD for `float`/`double` (net6+); net471 and other `T` use a tight scalar+enum loop. Both still avoid the per-element delegate.
- `_numOps` is read from the `TensorBase<T>` static singleton, not threaded as a parameter.
- `Tensor<T>` is not struct-constrained, so spans are reinterpreted via `Unsafe.As` + `MemoryMarshal.CreateSpan` (not `MemoryMarshal.Cast`).

**2. Pool broadcast result tensors** — `BroadcastAdd/Subtract/Multiply` now `Rent` the result (matching the same-shape fast paths) instead of `new Tensor<T>` on the GC heap. (General improvement for residual/non-channel broadcasts; see commit note for scope.)

## Verification

- **354 tests pass** (89 broadcast + 265 elementwise/gradient/BatchNorm/arithmetic), 0 failures. Compiles for both `net10.0` and `net471`.
- **Measured: ResNet50 float forward 1713 ms → 772 ms (2.2× faster)**; re-profile confirms `BroadcastElementwise` is gone from the hot path (now dominated by legitimate conv compute).

### DIFF
--- a/src/AiDotNet.Tensors/LinearAlgebra/Tensor.cs
+++ b/src/AiDotNet.Tensors/LinearAlgebra/Tensor.cs
@@ -3380,7 +3380,10 @@ public partial class Tensor<T> : TensorBase<T>, IEnumerable<T>
 
         // Get broadcast shape
         int[] broadcastShape = GetBroadcastShape(this._shape, other._shape);
-        var result = new Tensor<T>(broadcastShape);
+        // Pooled result — BroadcastElementwise writes every element (full
+        // coverage), so dirty rented memory is safe, and this keeps the
+        // broadcast result off the GC heap (matches the same-shape fast paths).
+        var result = TensorAllocator.Rent<T>(broadcastShape);
 
         // Span-based stride broadcast — replaces the previous
         // foreach (var index in result.GetIndices()) + indexer-per-element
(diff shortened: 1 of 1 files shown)


ooples:
@coderabbitai full review

ooples:
@coderabbitai full review

PR #451, opened 2026-05-26
test(#436): AIsEval MLP inference perf gate (verify MlpForward meets P1 target)

ooples:
Closes #456

## Summary

Part of #436 (AIsEval fair-comparison perf gaps). Adds the missing **AIsEval MLP inference perf gate** for the `MlpForward` fused primitive.

`MlpForward` shipped in #437 (with correctness tests) but was the **only** AIsEval primitive without a perf gate — CNN, LSTM, and MHA each got one in #437; MLP didn't. This closes that gap, mirroring the existing gates in `PerformanceRegressionTests`.

## What it measures — and the good news

At the AIsEval MLP shape — `Dense(784→512)→Dense(512→128)→Dense(128→10)` at `bs=128` — `MlpForward` measures **2.95 ms/iter** on net10.0 (Release) on the dev host:

| | ms/iter |
|---|---|
| PyTorch (baseline) | 1.91 |
| **MlpForward (fused, this primitive)** | **2.95** |
| Framework unfused per-layer `Predict` path (#436 P1) | 8.94 |

So **the P1 MLP gap is already closed at the primitive level** — 2.95 ms meets the issue's `<= 3 ms` acceptance (1.54× PyTorch). The reason the AIsEval rerun still showed 8.94 ms is that the framework's `DenseLayer.Forward` doesn't call `MlpForward` yet — that wiring is tracked framework-side in **ooples/AiDotNet#1447**.

## The gate

`MlpForward_Aiseval_FusedDispatch_BeatsUnfusedPath` budgets at **8 ms** — deliberately *below* the 8.94 ms unfused number so a regression back to the unfused/scalar dispatch path fails, with ~2.7× headroom over the measured 2.95 ms for CI noise. **Skip-by-default** to match the repo perf-gate convention (commit `978585c6`; run via `--filter PerformanceRegression`).

## Tests

- New gate compiles and is collected (skipped) on **net471 + net10.0**.
- The 6 existing `MlpForwardTests` correctness tests (float/double/chained-FusedLinear/GraphMode/bad-args) still pass on both TFMs.
- Manual run of the gate: **2.95 ms < 8 ms budget** ✅.

## Cross-references
- Umbrella: #436 (now updated with a cross-repo status section)
- Primitives (merged): #437
- Framework-side wiring of the primitives into NN layers + builder overhead: ooples/AiDotNet#1447

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #455, opened 2026-05-26
test: stabilize 3 flaky CI tests (PrePack, ConvT2DL2Perf, ParallelFor) — unblock 5 PRs

ooples:
## Unblocks 5 in-flight PRs

Three perf-gate/dispatch tests have been flaking on the `ubuntu-latest` 4-vCPU CI runner across PRs #446, #449 (originally a stand-alone fix for one of these), #450, #451, #452 (and likely future PRs). All three pass locally and on dev-host CI; all three fail on the small CI runner for runner-class-specific reasons that don't reflect any real regression. This PR consolidates all three fixes so a single merge unblocks the queue (was previously two separate PRs, #449 + #455, which got into a "neither can merge until the other lands" loop).

### 1. `PrePackSpeedupTest.PrePackedB_Reused_Across_Repeated_Gemms_Is_Faster`

**Failure**: gate ≥ 0.8× speedup, CI measures 0.25× (4× regression that doesn't reproduce locally).

**Root cause**: single-window wall-clock measurement of two GEMM loops; the ratio is noise-dominated whenever the GC fires (draining #441's GradientTape finalizer queue) or another xUnit thread spikes CPU.

**Fix**: drain pending finalizers + Gen 2 collect before EACH timing block; take min wall-clock across 3 repetitions per block; compute the speedup ratio from the two mins. Same hygiene pattern as #448. The 0.8× gate and `maxDelta < 1e-3` correctness assertion are unchanged.

### 2. `ConvTranspose2DL2PerfTest.L2Shape_FP64_BlasManagedGemm_BelowPerfTarget`

**Failure**: `Arch: AVX-512`, `Median: 207 ms`, `Gate: 5.0 ms (AVX-512)`.

**Root cause**: the 5 ms AVX-512 target was calibrated on a 32-core dev host. GitHub's `ubuntu-latest` runner pool now serves AVX-512-capable images, but the runners are still 4-vCPU and slow per-core — single-core SIMD throughput dominates the wall-clock at 200+ ms.

**Fix**: three-tier gate. AVX-512 + cores ≥ 8 → 5 ms (the post-K5 perf target); AVX-512 on smaller hosts or AVX2 → 1000 ms catastrophic-regression sentinel (sub-OpenBLAS/MKL 559 ms ceiling); scalar → no gate. Catches the same set of regressions on capable hosts; tolerates the per-core-bound CI runner.

### 3. `ParallelForOrSerialDispatchTests.ParallelForOrSerial_LargeWork_DeterministicFalse_FansOutToWorkers`

**Failure**: `totalWork (131072) >> grain size (32768) with DeterministicReductions=false should have dispatched to the worker pool, but all 4 chunks ran on caller thread`.

**Root cause**: `ParallelForOrSerial` uses raw `System.Threading.Tasks.Parallel.For`, which inlines very short iteration counts on the caller thread (4 iters × empty body on a 4-vCPU runner triggers the inlining heuristic). `ParallelForOrSerial`'s dispatch decision depends only on `totalWork`, `MaxDegreeOfParallelism`, and `DeterministicReductions` — `numChunks` is independent. The test's `numChunks = ProcessorCount = 4` was too small for TPL to fan out reliably.

**Fix**: bump `numChunks` to `max(16, ProcessorCount * 4)` so TPL is forced to spawn tasks (small-loop inlining doesn't kick in at 16+ iters); the production code under test is unchanged. Also lengthen the caller-thread worker-probe wait from 200 ms to 1 s to tolerate cold-pool spawn latency.

## Verified

All three tests pass locally; build clean.

## Supersedes

Closes #449 (the stand-alone Parallel.For fix). #455 already had the PrePack + ConvT2DL2Perf fixes; rolling #449's content in makes it one merge instead of two with a circular dependency.

### DIFF
--- a/tests/AiDotNet.Tensors.Tests/Engines/BlasManaged/ConvTranspose2DL2PerfTest.cs
+++ b/tests/AiDotNet.Tensors.Tests/Engines/BlasManaged/ConvTranspose2DL2PerfTest.cs
@@ -88,20 +88,23 @@ public class ConvTranspose2DL2PerfTest
         //   - Naive 7-nested loop:                100 ms
         // Spec target (peak): 0.44 ms (75 GFLOPS)
         //
-        // Gate selection — two-tier:
-        //   AVX-512 (dev/perf hosts):  5 ms (post-K5 perf target)
-        //   AVX2 (CI / older hosts):   1000 ms — catastrophic-regression
-        //     sentinel only. CI run 26304260634 measured 253 ms on the
-        //     ubuntu-latest 4-vCPU runner; the original 25 ms AVX2 gate was
-        //     calibrated on a 32-core dev host where the microkernel can
-        //     bind to enough physical cores. The 1000 ms ceiling still
-        //     catches a regression past the cblas_dgemm/MKL 559 ms ceiling
-        //     while accepting that the post-K5 perf target only materialises
-        //     on AVX-512 boxes (the spec target was always AVX-512-specific
-        //     per issue #358 — "Spec target: ≤1 ms on x64 AVX-512").
-        //   Scalar:                    no gate, just report.
+        // Gate selection — three-tier (#358 + ubuntu-latest-AVX-512 follow-up):
+        //   AVX-512 + ≥ 8 cores (dev/perf hosts):  5 ms (post-K5 perf target)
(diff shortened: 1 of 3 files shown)


ooples:
@coderabbitai full review

PR #461, opened 2026-05-26
perf(#407): measurement-based (Mc, Nc, Kc) block-size autotune

### LINKED ISSUE #407: perf: Sub-Q — measurement-based (Mc, Nc, Kc) autotune
**Parent:** #358 BlasManaged perf sprint

## Problem

`AutotuneDispatcher.FallbackToHeuristic` always returns `(Mc, Nc, Kc) = (min(64, m), min(64, n), min(64, k))`. Standard BLIS for AVX2 Zen-class: `(Mc=144, Nc=2496, Kc=384)`. Skylake-X: `(Mc=72, Nc=4080, Kc=192)`.

Small blocks are wrong by ~10× on memory-traffic terms:
- Pack-B writes Kc·Nc bytes per (jc, pc) block. With our (64, 64): 16 KB. With BLIS (256, 2496): 2.5 MB but each block is reused Nc times — far better amortization.
- Number of pack-B passes for an FFN_1024×3072×768: ours = (3072/64)·(768/64) = 576 packs. BLIS = 1·3 = 3 packs. **192× more pack work**.

The autotune cache exists (`BlasManagedAutotune.Store/TryLookup`) but the "measurement" is just heuristic-stored — never benchmarks candidates.

## Fix

Real measurement-based autotune:

1. On cache miss, generate a candidate set of `(Mc, Nc, Kc)` tuples — e.g., the cross product of `Mc ∈ {64, 128, 192}`, `Nc ∈ {256, 512, 1024, min(n, 2048)}`, `Kc ∈ {128, 256, 384, 512}` clipped to shape.
2. Benchmark each candidate via the existing `PrefersManagedCache.MeasureFp32/Fp64` probe pattern (already accurate to ~10%).
3. Store the winner with measured timing in `BlasManagedAutotune`.
4. Subsequent calls of the same shape return the cached winner.

Reuse `PrefersManagedCache.BypassAutotune` ThreadStatic to prevent recursion.

## Files

- `src/AiDotNet.Tensors/Engines/BlasManaged/Autotune/AutotuneDispatcher.cs` (real measurement path)
- `src/AiDotNet.Tensors/Engines/BlasManaged/Autotune/BlasManagedAutotune.cs` (cache schema already supports it per H1 commit)
- New `src/AiDotNet.Tensors/Engines/BlasManaged/Autotune/BlockSizeSweep.cs` (candidate generation + measurement driver)

## Acceptance

- First call on a new shape: 5-15ms of measurement overhead (cached forever after).
- Bench catalog re-snapshot after warmup: median loss ratio drops from 17× → ≤8×.
- Cache persists across processes (already does via Sub-F4).
- Add a `AIDOTNET_AUTOTUNE_OFFLINE=1` env to pre-populate the cache via the bench harness so first-call overhead doesn't surprise production users.

## Risk

- Measurement noise can pick a sub-optimal winner. Mitigate via multiple-run-median and a 10% tolerance band (prefer cache hits within 10% of new measurement).
- Cache invalidation: hardware fingerprint covers most cases; on CPU upgrade the cache is auto-discarded (Sub-F4 already does this).

ooples:
> ⚠️ **Correction (later commit on this branch).** The S.2 conclusion below — "FP32 kernel already matches a register-only FMA loop / residual gap is a managed-runtime ceiling, not kernel structure" — was **WRONG**, due to a calibration bug. The peak loop used **12 accumulator chains, which the RyuJIT spills** (it enregisters ~8 YMM before spilling), so it under-measured the true managed FMA ceiling ~2×.
>
> Corrected (8-chain calibration + new `FmaCeilingProbe`): the kernels run at **~42–44% of the managed FMA ceiling (~90 GFLOPS FP32 / ~44 FP64)**, and **OpenBLAS hits ~62 GFLOPS FP64 on this same chip** — so there is **~2.4× kernel headroom and the ≤1.5×-vs-OpenBLAS goal is a real, legitimate target, not a wall.** The bottleneck is load/broadcast latency (source-level unroll/reorder doesn't move it — the JIT reschedules), so closing it is JIT-disasm-driven microkernel work (the open **S.3** task). The text below is kept for history.

---

Closes #407

## Summary

Implements Sub-Q (#407): replace the heuristic-only block-size choice with a real **measurement-based** `(Mc, Nc, Kc)` autotune. The dispatcher previously only ever stored the heuristic decision on a cache miss — the `MeasurementEnabled` flag existed but was **wired to nothing**. This adds the candidate-sweep measurement path it was a placeholder for.

## How it works

- **`BlockSizeSweep.Measure<T>`** generates a candidate set — the cross product of the BLIS-style ranges from the issue (`Mc∈{64,128,192}`, `Nc∈{256,512,1024,min(n,2048)}`, `Kc∈{128,256,384,512}`), each clamped + aligned to the shape and deduped (so moderate shapes collapse to a handful) — times the **real managed `Gemm`** on each (one warmup + median of 3 runs), and returns the fastest. Always includes the heuristic default `(128,512,256)` as a safety candidate, so the measured winner is never worse than the heuristic.
- **Candidate injection + recursion guard:** a per-thread `AutotuneDispatcher.BlockOverride` carries the `(Mc,Nc,Kc)` under test. A nested `Gemm → Decide` during a sweep returns the override instead of starting another sweep — this is the role #407 earmarked for `BypassAutotune`; a dedicated override is cleaner because it also carries the blocking being probed.
- **`Decide` wiring:** on cache miss, measure-and-cache the winner when enabled **and** the shape is big enough that a block choice beyond the heuristic clamp actually exists (`ShouldMeasure`: `m·n·k ≥ 1M` and a dim exceeds the clamp); otherwise fall back to the heuristic exactly as before. **Off by default** (`AIDOTNET_BLAS_AUTOTUNE_MEASURE=1`), so default behavior is byte-for-byte unchanged.
- **Offline pre-population:** `BlockSizeSweep.PrepopulateCommonShapes()` + `AIDOTNET_AUTOTUNE_OFFLINE` — a warmup entry that measures-and-caches a curated set of common transformer FFN/QKV/MLP shapes (forced measurement on the warmup thread) so production never pays first-call overhead. No auto-trigger at assembly load (that would slow CI/startup); consumers call it during warmup.
- Winners persist across processes via the **existing on-disk autotune cache** (Sub-F4) — no new persistence code.

## Acceptance mapping

- ✅ First call on a new shape measures candidates and caches the winner (cached forever after).
- ✅ Cache persists across processes (existing disk cache; hardware-fingerprinted, auto-discarded on CPU change).
- ✅ `AIDOTNET_AUTOTUNE_OFFLINE` pre-population path.
- ✅ **Median loss ratio vs native — measured and confirmed.** Ran the loss-ratio bench (`BlockSizeSweepLossRatioBench`, FP64) on `x64-amd-avx2-cpu32`:

| Shape | native | heuristic | autotune | lossH | lossA |
|---|---|---|---|---|---|
| FFN_128x768x768 | 1.30ms | 37.5ms | 4.66ms | **29.0×** | **3.6×** |
| FFN_up_512x2048x512 | 3.79ms | 23.5ms | 12.4ms | 6.2× | 3.3× |
| FFN_down_512x512x2048 | 6.58ms | 22.8ms | 11.7ms | 3.5× | 1.8× |
| BERT_FFN_1024x3072x768 | 13.3ms | 58.6ms | 29.2ms | 4.4× | 2.2× |
| GPT2_FFN_1024x3072x768 | 17.5ms | 55.3ms | 29.5ms | 3.2× | 1.7× |
| ViT_QKV_768x2304x768 | 7.81ms | 39.6ms | 21.5ms | 5.1× | 2.8× |
| Large_1024sq | 9.60ms | 25.5ms | 12.7ms | 2.7× | 1.3× |

  **Median loss ratio: heuristic 4.4× → autotune 2.2×** — the autotune halves the gap to native; every shape improves (FFN_128x768x768 goes 29×→3.6×, ~8× faster), comfortably under the ≤8× target. The measured winners consistently prefer `Mc=64` with much larger `Nc` (2048) / `Kc` (384–512) than the heuristic `(128,512,256)` — the defaults were leaving real perf on the table. (Bench gated behind `AIDOTNET_RUN_407_BENCH=1` + `[Category=Benchmark]`, excluded from normal CI.)

## Risk / safety

- Default path unchanged (measurement is opt-in). The sweep only runs the **managed** kernel with different tilings — blocking changes loop traffic, never the math.
- Measurement failures are caught and fall through to the heuristic, so a probe error can never break a GEMM.

## Tests

13 new tests (net471 + net10.0):
- arbitrary injected blockings (tiny / over-shape / unaligned) compute a correct GEMM vs a scalar reference,
- `Measure` returns a valid clamped + aligned winner and leaks no thread override,
- the measured winner itself computes a correct GEMM,
- the `Decide` override path and the forced-measure **end-to-end** path,
- double support, and `ClampBlocking` edge cases.

All **380 existing BlasManaged tests still pass**; library builds clean on both TFMs.


ooples:
@coderabbitai full review

Issue #465, opened 2026-05-26
feat(simd): W8A8 int8×int8 GEMM — true int8 compute (activation quant + VPDPBUSD/AVX2 emulation)


ooples:
## Motivation

The current weight-only INT8 path (`SgemmWithInt8RowScaledCachedB`, #427) keeps **fp32 activations** and dequantizes the int8 weights to fp32 before an fp32 macro-kernel. Benchmarking (AiDotNet #1417) showed this is **strictly fp32-cost + overhead** — it cannot beat fp32 in a compute-bound regime, and the INT8/FP32 wall-clock ratio actually *worsens* with shape (16×64 ≈ 21×, 128×512 ≈ 57× on an AVX2 host). Weight-only INT8's only speed win is **memory bandwidth** (4× smaller weight loads) in memory-bound (m≈1 decode) regimes.

A real *compute* speedup needs **int8 weights × int8 activations** accumulated in **int32** (VPDPBUSD / VPMADDUBSW), which has genuinely higher throughput than fp32 FMA. That requires quantizing activations too (W8A8), which this issue tracks.

## Scope (pure Tensors)

Encapsulate the whole thing behind one self-contained entry point so the consumer (AiDotNet's `Int8WeightOnlyMatMul`) is a trivial call swap:

```
SgemmA8W8RowScaledCachedB(
    ReadOnlySpan<float> a,        // fp32 activations [m, k]
    sbyte[] bInt8,                // pre-quantized weights [n, k]
    ReadOnlySpan<float> rowScales,// per-row weight scales, length n
    Span<float> c, int m, int k, int n)
```
The kernel quantizes `a` to int8 internally (dynamic, per-row symmetric), runs the int8×int8 GEMM, and dequantizes the int32 accumulator with `actScale[r] · weightRowScale[c]`.

## Design

- **Activation quant (new, internal):** dynamic per-call, **per-row (per-token) symmetric** int8 + the unsigned-shift trick (`VPMADDUBSW` wants `uint8 × sbyte`). Per-row, not per-tensor, to preserve accuracy across tokens.
- **Kernel:**
  - **VNNI path** — `VPDPBUSD` (`Avx512Vnni.DotInt8(byte,sbyte)` already has the signature; gate on AVX-512-VNNI / AVX-VNNI).
[...]

Issue #401, opened 2026-05-19
Add SimdGemm.SgemmWithInt8RowScaledCachedB — per-row-scaled INT8 weight-only GEMM


ooples:
## Goal

Add a public `SimdGemm` entry point that accepts **pre-quantized INT8 weights with per-row scales** and runs the same cache-blocked tiled SGEMM pipeline as `SgemmWithInt8CachedB`. The existing `SgemmWithInt8CachedB` requires FP32-input weights (it computes a per-tensor symmetric scale internally), which is incompatible with the per-row quantized weights AiDotNet's `Int8InferenceModel` produces via `Int8WeightOnlyQuantization.QuantizePerRow`.

This is the Tensors-side prerequisite for closing the remaining wall-clock gap in [ooples/AiDotNet#1349](https://github.com/ooples/AiDotNet/issues/1349).

## Background

AiDotNet's INT8 inference path (`QuantizedDenseLayer`, `QuantizedAttentionLayer`) currently routes through `AiDotNet.Inference.Quantization.Int8WeightOnlyMatMul.MultiplyAddBias`. That helper does:

1. Per-row AVX2 dequant (`Int8Quantizer.DequantizeInt8ToFloat32`) into an ArrayPool-rented FP32 scratch buffer.
2. Standard FP32 `SimdGemm.Sgemm` over the scratch.
3. SIMD bias-add scatter.

On a 16×64×64 transformer canary this lands at ~15× the FP32 baseline. The matmul body is fast — the gap is per-call overhead (per-row dequant, ArrayPool rent/return, scatter loop). On BERT-class shapes the kernel body dominates and the ratio is much closer to 1×, but the small-shape ratio is what consumers configure their fixtures around.

`SgemmWithInt8CachedB` already exists and does the right thing in principle: pre-pack INT8 B into the tile layout `SgemmTiledParallel2D` expects, cache by reference (`ConditionalWeakTable`), dequant per-tile inside the macro-kernel into an L1-resident float scratch. It just isn't reachable from a caller whose weights are already INT8 with per-row scales.

## Proposed API

```csharp
namespace AiDotNet.Tensors.Engines.Simd;

internal static partial class SimdGemm
{
    /// <summary>
    /// Compute C = A · dequant(B_int8, rowScales) using a cached pre-packed
    /// form of B that's built once per (sbyte[] reference, k, n) tuple and
    /// reused across calls. Per-row scales are folded into the macro-kernel's
[...]

PR #466, opened 2026-05-26
perf(#401): fused int8×fp32 micro-kernel — widen weights in-register, fold row-scale into store

ooples:
## Summary

`SgemmWithInt8RowScaledCachedB` (#427) previously **dequantized the cached int8 weights into a full FP32 buffer on every call**, then ran the FP32 macro-kernel — i.e. the same per-call work as the old dequant+Sgemm path. So tiny shapes saw no win and even regressed slightly from the cache-lookup + extra `ArrayPool.Rent` overhead.

This PR adds **fused int8×fp32 micro-kernels** that keep the weights packed int8 all the way into the kernel:

- Each K-row's 16 int8 weights are widened to fp32 **in-register** (`VPMOVSXBD` + `CVTDQ2PS`) during the FMA loop — no separate dequant pass, no FP32 B buffer, no per-call dequant `ArrayPool` rent.
- The per-column row-scale is constant across the K loop, so it's **folded into the store via one FMA per output row** (`C += accum·scale`) instead of a multiply per K step.
- Full `MicroKernel6x16Int8RowScaledFused` + masked-edge variant; the sequential **and** parallel drivers both route through the new `MacroKernelInt8RowScaledFused`. Removed the now-dead packed dequantizer.

This realizes the "INT8 all the way through the kernel" intent: `C[r,c] += scale[c] · Σ_k A[r,k]·B_int8[c,k]`.

## Verification

- **20/20 int8 GEMM correctness tests pass** (full + masked-edge tiles, SNR envelope), both `net10.0` and `net471`.
- Measured downstream (AiDotNet #1417 canary, 16×64 transformer): INT8/FP32 wall-clock ratio **~18× → ~13-15×** consistently — moves the canary from flaky-at-the-20×-ceiling to passing with margin. Benefits the memory-bound / m≈1 decode case.

## Scope note

This is the **weight-only (W8A16)** path: activations stay fp32, so the kernel does fp32 FMA + in-register int8 widening. By design it **cannot beat fp32 in a compute-bound regime** — its win is eliminating the dequant round-trip and the 4× weight-DRAM saving in memory-bound regimes. The true *compute* speedup (int8×int8 → int32) is tracked separately as W8A8 in #465.

## Follow-up

- Single-step K loop (dropped the 4-way unroll for the first cut) — a possible micro-optimization.
- Downstream consumer (AiDotNet #1417) bumps to a published version containing this.


ooples:
@coderabbitai full review

PR #472, opened 2026-05-27
feat(#467): Float32Precision opt-in for FusedAttention<double> + fix float SDPA nested-parallelism starvation

### LINKED ISSUE #467: perf: FlashAttentionLayer&lt;double&gt; dominates SD UNet forward at seqLen=4096 — 70% of ConsistencyModel.Predict time
## Summary

Deep sub-phase profiling of `ConsistencyModel<double>.Predict` at the canonical [1, 4, 64, 64] FP64 latent shape (the failing `ScaledInput_ShouldChangeOutput` test in [AiDotNet #1305](https://github.com/ooples/AiDotNet/issues/1305) / [#1394](https://github.com/ooples/AiDotNet/issues/1394) cluster #6) shows **70.3% of the 14.95 s/step UNet forward is in `FlashAttentionLayer<double>`**. Tensors-side fix needed to close the 120 s timeout budget; AiDotNet-side wrapper code is already < 1% of total time.

Companion to [#413](https://github.com/ooples/AiDotNet.Tensors/issues/413) (SD UNet "no single cliff" residual) — this issue narrows the residual to a specific layer/kernel.

## Workload

ConsistencyModel<double>(seed: 42) — SD-scale UNet:
- baseChannels: 320, channelMultipliers: [1, 2, 4, 4], 2 ResBlocks/level, numHeads: 8 (assumed default)
- attentionResolutions: [4, 2, 1] — `Array.IndexOf` indexing puts self-attention at encoder/decoder levels 1 and 2
- Input shape: [1, 4, 64, 64] FP64 (the diffusion test fixture)
- `DefaultInferenceSteps = 2` (paper-canonical per Song et al. 2023)
- Test calls Predict 2× → 4 UNet forwards per test

## Per-section breakdown (one warm UNet forward, `tools/ConsistencyModelPerfDiag`)

Captured via the new opt-in `UNetNoisePredictor<T>.ForwardProfilingSink` hook landing in AiDotNet PR #1456. Single-process, Release, net10.0, 16-core host:

| Sub-phase | Time | % of UNet step |
|---|---:|---:|
| **encoder attn × 8** | **4763 ms** | **34.4%** |
| **decoder attn × 8** | **4482 ms** | **32.4%** |
| decoder resblock × 16 | 1635 ms | 11.8% |
| decoder upsample × 6 | 1408 ms | 10.2% |
| encoder resblock × 16 | 799 ms | 5.8% |
| middle attn × 2 | 488 ms | 3.5% |
| encoder downsample × 6 | 135 ms | 1.0% |
| decoder concat × 16 | 78 ms | 0.6% |
| middle resblock × 4 | 38 ms | 0.3% |
| input_conv × 2 | 19 ms | 0.1% |
| output_conv × 2 | 2 ms | 0.0% |
| **Total UNet step** | **13.85 s** | **100%** |

Combined attention (encoder + decoder + middle) = **9733 ms = 70.3%**.

## Slowest individual attention calls

| Section | Time | Notes |
|---|---:|---|
| enc[3].attn | 1845 ms | Level 1, channels=640, spatial=64, **seqLen=4096** |
| dec[7].attn | 1626 ms | Level 1 mirror, same shape |
| dec[6].attn | 1340 ms | Level 1 mirror, same shape |
| enc[4].attn | 1246 ms | Level 1 (2nd ResBlock), same shape |
| enc[6].attn | 916 ms  | Level 2, channels=1280, spatial=32, **seqLen=1024** |
| dec[4].attn | 780 ms  | Level 2 mirror |
| enc[7].attn | 757 ms  | Level 2 (2nd ResBlock) |
| dec[3].attn | 736 ms  | Level 2 mirror |
| mid[0].attn | 488 ms  | Level 3 bottleneck, channels=1280, spatial=8, seqLen=64 |

**Pattern**: highest-resolution self-attention dominates. seqLen=4096 × headDim=80 (assuming numHeads=8) is the per-call hot spot.

## What's already engaged

`AiDotNet.Diffusion.Attention.DiffusionAttention<T>` routes to `FlashAttentionLayer<T>` (not standard MultiHeadAttention) for these sequence lengths — `flashAttentionThreshold = spatialSize² / 16` puts everything above 4 sequences through Flash. So Flash IS the path being measured here.

`FlashAttentionConfig`:
- BlockSizeQ = 64, BlockSizeKV = 64
- RecomputeInBackward = true
- ReturnAttentionWeights = false
- **Precision = FlashAttentionPrecision.Float32** ← already mixed-precision intent, model is FP64

So the model declares `T = double` but FlashAttention is supposed to be running in FP32 internally. Worth confirming this is engaging (FP32 path) on the FP64 model surface and whether the precision conversion has overhead.

## Reproducer (AiDotNet side)

After PR #1456 lands:

```bash
cd <AiDotNet repo>
dotnet build src/AiDotNet.csproj -c Release -f net10.0
dotnet build tools/ConsistencyModelPerfDiag/ConsistencyModelPerfDiag.csproj -c Release
dotnet run --project tools/ConsistencyModelPerfDiag/ConsistencyModelPerfDiag.csproj -c Release --no-build
```

Output includes the per-section breakdown above. Output also written to `perf-1305-consistencymodel.log`.

## What's likely contributing (Tensors-side)

1. **FP64 SIMD lane count**: AVX2 = 4 lanes for double vs 8 for float. If `FlashAttentionPrecision.Float32` isn't fully engaging on `T = double` surface — or only the inner matmul converts and outer ops stay FP64 — we pay full FP64 cost.

2. **Q · K^T scratch at seqLen=4096**: even with block-streaming, the per-block FP64 working set is 64 × 64 × headDim × 8 bytes. headDim=80, blockQ=blockKV=64 → 3.3 MB per active tile, well into L2 churn.

3. **Per-call workspace allocation**: each FlashAttention call allocates Q/K/V projection outputs + intermediate score buffers. 8 attention calls per encoder × 4 forwards per test = 32 allocations. Pooled per-shape scratch (similar to the Conv2D im2col reuse discussed in [#415](https://github.com/ooples/AiDotNet.Tensors/issues/415) Phase C) would help.

4. **Backward-vs-inference dispatch**: `RecomputeInBackward = true` is correct for memory but in inference mode (NoGradScope active during Predict), the layer should be able to skip recompute-bookkeeping. Worth verifying it does.

## Expected scope

Phase A — **Precision audit**: confirm `FlashAttentionPrecision.Float32` engages end-to-end for `T = double` callers. If only the inner matmul downcasts and the surrounding softmax / aggregation stays FP64, fix the FP32 boundary to cover the full kernel. Estimated win: 1.5-2× on FP64 surface based on AVX2 lane count.

Phase B — **Inference fast path**: when called under `NoGradScope`, skip backward-recompute bookkeeping. Smaller win but compounds with A.

Phase C — **Per-shape scratch reuse**: thread-local Q/K/V projection workspace keyed by (seqLen, headDim, batch), similar to Phase C of #415. Cuts per-call alloc cost.

Phase D — **Block size tuning at seqLen=4096**: BlockSizeQ=BlockSizeKV=64 is GPU-optimal but CPU L2 sweet spot may differ. Sweep {32, 64, 128} on FP64 to find CPU optimum.

## Affected callers (AiDotNet)

Any LatentDiffusionModelBase subclass with self-attention in its noise predictor at seqLen ≥ 256 on `T = double`. From the diffusion test corpus that maps to the 75+ test classes with `InputShape >= [1, 4, 64, 64]`:

- `ConsistencyModelTests.ScaledInput_ShouldChangeOutput` (currently visible victim, [AiDotNet #1305](https://github.com/ooples/AiDotNet/issues/1305))
- All other model-family tests in the cluster #6 long-training timeout group (AiDotNet #1314)
- Most likely the same fix helps SD3/Flux/HiDream/HunyuanDiT/AuraFlow tests that share the SD-UNet attention pattern

## NOT this issue

- Cross-attention performance — `_contextDim > 0` is false for ConsistencyModel (no text conditioner), so DiffusionCrossAttention isn't in the measured hot path. Cross-attention would add cost on text-conditioned models but is a separate measurement.
- The 17.6% Conv2D / GroupNorm cost in ResBlocks — already tracked in [#415](https://github.com/ooples/AiDotNet.Tensors/issues/415).
- The 10.2% Deconv2D in Upsample — sibling to #415.

## Cross-references

- AiDotNet #1305 — original test failure
- AiDotNet #1394 — perf umbrella for these tests
- AiDotNet #1456 — PR adding the sub-phase profiling hook + perf-diag tool
- Tensors #413 — SD UNet residual perf (this issue narrows that to attention specifically)
- Tensors #415 — FP64 Conv2DBackward (sibling for the conv slice)
- Tensors #411 — closed: FlashAttention<double> seq=1024 hang (so the kernel works at these shapes — it's now a perf-tier issue, not a correctness one)

🤖 Filed with [Claude Code](https://claude.com/claude-code)

ooples:
## Summary

Closes #467 (FlashAttentionLayer<double> dominating SD-UNet forward), and fixes a latent thread-pool starvation bug uncovered while implementing it.

### `feat(#467)` — `Float32Precision` opt-in
SD-UNet self-attention at seqLen ≥ 1024 dominates `ConsistencyModel.Predict` on the FP64 surface. New `FlashAttentionConfig.Float32Precision` lets a `double` caller run the whole attention path in `float` (8 AVX2 lanes vs 4 for `double`, via the SimdGemm-backed SDPA) and upcasts the result. It re-enters `FusedAttention<float>.Forward`, so rank-3 promotion, causal masking, additive bias and attention weights all behave identically — just faster and lossy (FP32 accumulation). No-op for `T == float` (the branch is gated to `double`), so the recursive call can't re-enter.

### `fix` — nested-parallelism ThreadPool starvation in float SDPA
While validating the FP32 path, the float SDPA was found to **hang for minutes** at `[1,8,1024,64]` (stack-trace confirmed). Root cause: `ScaledDotProductAttentionFloat` runs `Parallel.For` over batch·heads and, inside each worker, called a **multi-threaded** GEMM (native OpenBLAS `cblas_sgemm` / managed machine kernel). The outer workers occupied every ThreadPool thread and then blocked on the inner GEMM's tasks — the pool only recovers via slow thread injection. The `double` path was immune because it uses a single-threaded inner GEMM.

Fix:
- `CpuParallelSettings` gains a thread-local *in-parallel-region* guard. `ParallelForOrSerial` / `ParallelForChunks` mark their workers in-region and nested parallel loops collapse to serial (the no-nested-parallelism default of OpenMP / PyTorch). New `ParallelForRegion` helper carries the same behavior to the raw `Parallel.For` sites in `StreamingStrategy` and `PackAOnlyStrategy`.
- `ScaledDotProductAttentionFloat` consults `IsInParallelRegion`: a parallel head-worker forces the single-threaded AVX2 `SgemmSequential` path (mirroring the double path); a serial outer loop keeps the fast multi-threaded GEMM for the lone head.

## Measured speedup (FP32-internal vs FP64)

Best-of-6 runs on a 32-core box. Per-shape minimum of each path (a concurrent
workload on the box only ever inflates timings, so the minimum best approximates
quiet-machine performance):

| shape [B,H,S,D] | best FP64 | best FP32-internal | speedup |
|---|---|---|---|
| [1,8,1024,64] (SD-UNet) | 109.9 ms | 24.7 ms | **4.45×** |
| [1,8,2048,64] (long ctx) | 281.3 ms | 97.6 ms | **2.88×** |
| [4,16,256,72] (DiT-XL) | 24.0 ms | 17.4 ms | **1.38×** |

Above the issue's 1.5–2× estimate for the large shapes (the float path gets both
8-lane SIMD and the better-tuned `SgemmDirect` kernel vs the double `MultiplyBlocked`);
the small batched shape gains less, as the smaller per-head GEMMs amortize the
f64↔f32 conversions less. The FP64 baseline was the noisier of the two under load
(its multi-threaded `MultiplyBlocked` is more load-sensitive); per-run ratios ranged
3.2–7.9× at [1,8,1024,64], 2.3–3.3× at [1,8,2048,64], 1.3–1.7× at [4,16,256,72].
Reproduce with the env-gated benchmark: `AIDOTNET_RUN_JIT_PERF=1 dotnet test --filter Float32Precision_IsFasterThanFp64_AtUNetShape`.

## Tests
- `FusedAttentionFp32PrecisionTests` — FP32 within 2e-3 of the FP64 reference across shapes / causal / bias / rank-3 / non-block-aligned lengths; proves the FP32 path genuinely engages (diff > 0); still returns valid attention weights; bit-identical for `float` T.
- `NestedParallelismTests` — region-guard semantics, nested loops serializing, and float SDPA at `[1,8,1024,64]` / `[2,8,512,64]` / `[1,16,256,72]` completing without starvation and matching the double reference.
- No regressions: 56/56 across `ParallelForOrSerial`/`GrainSize` dispatch, MachineKernel + SimdGemm GEMM correctness, and the attention suites. Builds on net471 and net10.0.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/FusedAttention.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/FusedAttention.cs
@@ -80,6 +80,23 @@ public sealed class FlashAttentionConfig
     /// </summary>
     public bool UseFlashAttention2 { get; set; }
 
+    /// <summary>
+    /// Mixed-precision opt-in for <c>T == double</c> callers: run the attention
+    /// kernel internally in <c>float</c> (FlashAttention-2), then upcast the result
+    /// back to <c>double</c>. AVX2 packs 8 floats vs 4 doubles per register, so the
+    /// FP32-internal kernel is ~2× the throughput of the full FP64 path, and it
+    /// inherits FlashAttention-2's O(seqLen) memory — the win that dominates SD-UNet
+    /// self-attention at seqLen≥1024 on the FP64 model surface (#467 Phase A).
+    /// <para>
+    /// This is LOSSY (FP32 accumulation, ~1e-3 relative vs the FP64 reference), so
+    /// it is opt-in and intended for inference where FP32 accuracy suffices. No-op
+    /// when <c>T != double</c> (float already runs the float kernel; Half is
+    /// unaffected). Like the FlashAttention-2 path it cannot return attention
+    /// weights (the block-tiled kernel never materialises the score matrix).
+    /// </para>
+    /// </summary>
+    public bool Float32Precision { get; set; }
+
     /// <summary>Default configuration — every field at its spec default.</summary>
     public static FlashAttentionConfig Default => new();
 }
@@ -152,6 +169,25 @@ public static class FusedAttention<T>
         if (value is null) throw new ArgumentNullException(nameof(value));
         config ??= new FlashAttentionConfig();
         engine ??= new CpuEngine();
+
+        // #467 Phase A — mixed-precision opt-in for double callers: re-run the WHOLE
+        // attention path in float (8 AVX2 lanes vs 4 for double, via the float
+        // SimdGemm-backed SDPA / FlashAttention-2 / bias path), then upcast the
+        // result. Reusing the float Forward means rank-3 promotion, causal masking,
+        // additive bias and attention-weights all behave identically — just faster
+        // and lossy (FP32 accumulation). The flag is a no-op for T==float (this
+        // branch is gated to double), so the recursive call can't re-enter.
+        if (config.Float32Precision && typeof(T) == typeof(double))
+        {
+            var qf = DoubleToFloat((Tensor<double>)(object)query);
+            var kf = DoubleToFloat((Tensor<double>)(object)key);
+            var vf = DoubleToFloat((Tensor<double>)(object)value);
+            var biasF = attentionBias is null ? null : DoubleToFloat((Tensor<double>)(object)attentionBias);
+            var (fOut, fWeights) = FusedAttention<float>.Forward(qf, kf, vf, config, biasF, engine);
+            var outD = (Tensor<T>)(object)FloatToDouble(fOut);
+            var weightsD = fWeights is null ? null : (Tensor<T>)(object)FloatToDouble(fWeights);
+            return (outD, weightsD);
+        }
 
         // Rank normalisation — 3D [B, S, D] → 4D [B, 1, S, D]. Dispatcher
         // below always sees 4D so shape-validation lives in one place.
@@ -227,6 +263,7 @@ public static class FusedAttention<T>
                     nameof(config));
             return (outT, null);
         }
+
 
         // Causal no-bias → synthesize the offset-aware -inf upper-triangle
         // bias and route through the bias path. Honours queryOffset so the
@@ -330,6 +367,28 @@ public static class FusedAttention<T>
             return engine.Reshape(t, new[] { t._shape[0], t._shape[2], t._shape[3] });
         }
         return t;
+    }
+
+    // #467 Phase A: contiguous element-wise narrowing/widening for the
+    // Float32Precision mixed-precision path. Plain loops — the cost is O(elements)
+    // and dwarfed by the GEMM-heavy attention kernel; the win is running that kernel
+    // in float (8 AVX2 lanes) instead of double (4).
+    private static Tensor<float> DoubleToFloat(Tensor<double> t)
+    {
+        var src = t.AsSpan();
+        var result = Tensor<float>.CreateZeros((int[])t._shape.Clone());
+        var dst = result.AsWritableSpan();
+        for (int i = 0; i < src.Length; i++) dst[i] = (float)src[i];
+        return result;
+    }
+
+    private static Tensor<double> FloatToDouble(Tensor<float> t)
+    {
+        var src = t.AsSpan();
+        var result = Tensor<double>.CreateZeros((int[])t._shape.Clone());
+        var dst = result.AsWritableSpan();
+        for (int i = 0; i < src.Length; i++) dst[i] = src[i];
+        return result;
     }
 
     private static Tensor<T> TransposeLastTwo(IEngine engine, Tensor<T> t)
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/Strategies/PackAOnlyStrategy.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/Strategies/PackAOnlyStrategy.cs
@@ -110,7 +110,7 @@ internal static class PackAOnlyStrategy
                 int mLocal = m, kLocal = k, mcLocal = mc, kcLocal = kc, mrLocal = mr, nrLocal = nr;
                 bool taLocal = transA;
 
-                Parallel.For(0, procsClamped, p =>
+                CpuParallelSettings.ParallelForRegion(procsClamped, p =>
                 {
                     int tileStart = (int)(((long)p * nTilesTotal) / procsClamped);
                     int tileEnd = (int)(((long)(p + 1) * nTilesTotal) / procsClamped);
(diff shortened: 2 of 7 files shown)


ooples:
@coderabbitai full review

Issue #475, opened 2026-05-27
perf(#436): route small inference GEMMs to fastest kernel (MLP OpenBLAS-vs-MKL floor)


ooples:
Follow-up from #436 / PR #474.

The fresh head-to-head (PR #474) showed `MlpForward` at ~1.37ms median / 1.75ms p95 after the native-BLAS thread cap, vs torch ~0.63ms median. The **zero-alloc raw 3-GEMM floor** (`--ab-aiseval-floor`) is ~1.2ms at optimal threads — already at/above torch's *entire* MLP median. So the residual is **small-GEMM kernel quality**: `FusedLinear` routes to native **OpenBLAS** (`SgemmRaw`), torch uses **MKL**. Our `IsMklVerified` path also measured ~1.4ms (not torch-grade MKL).

**Goal:** route small-batch inference GEMMs to the fastest available kernel so MlpForward reaches p95 < torch median (~0.63ms) on the AIsEval shape (Dense 784→512→128→10, bs=128).

Directions: investigate why our MKL-verified path is slower than torch's bundled MKL; consider a dedicated small-GEMM microkernel (cf. #368/#375 BlasManaged Sub-S machine-code path) wired into `FusedLinear`; INT8.

Reproduce: `dotnet run -c Release -- --ab-aiseval-floor` and `--ab-aiseval-h2h`.

ooples:
### A/B investigation (disciplined, scientific) — root cause is parallel-scaling overhead, not microkernel

Measured on a 16-core box (net10.0), isolated single-GEMM A/B + scaling sweep. **The reported "managed ~37× slower" was a red herring** — it was the `Sgemm(a,b,c,m,k,n)` 6-arg overload (11 GF), which `FusedLinear` does **not** use. The real path (`SgemmWithCachedB`) is healthy:

| MLP layer GEMM | managed `SgemmWithCachedB` | native (MKL here) |
|---|---|---|
| `[128×784]@[784×512]` (layer 0, ~85% FLOPs) | **216–243 GF** | ~309–393 GF |
| `[128×512]@[512×128]` (layer 1) | **236–246 GF** | 92–183 GF (**managed wins ~1.5–2×**) |
| `[128×128]@[128×10]` (layer 2) | ~31 GF | ~35 GF |

So the only gap is layer-0, and decomposing it:

- **Single-thread microkernel = 107–116 GF — fine.** Not the problem.
[...]

PR #479, opened 2026-05-27
perf(#476): zero-alloc ScaledDotProductAttention in MultiHeadAttentionForward

### LINKED ISSUE #476: perf(#436): zero-alloc ScaledDotProductAttention + faster managed GEMM (MHA)
Follow-up from #436 / PR #474.

`MultiHeadAttentionForward` measured ~5.31ms median / 7.25ms p95 vs torch ~2.74ms (PR #474 head-to-head). Two root causes:
1. **7.3 MB allocation per call** (`--ab-aiseval-diag`), dominated by the internal `ScaledDotProductAttention` scores/softmax tensors (the projection scratch already uses ArrayPool).
2. Managed `SimdGemm` projection/score GEMM quality. The MaxDOP sweep (`--ab-aiseval-dopsweep`) shows MHA scales monotonically to 16 cores (no thread-cap lever).

**Goal:** MHA p95 < torch median (~2.74ms) at [128,32,64] h=4.

Directions: pooled/scratch SDPA (eliminate the per-call scores+softmax allocation), faster managed projection GEMM or route to native, fused softmax.

ooples:
## Summary

Addresses #476 root cause **1 — the 7.3 MB allocation per `MultiHeadAttentionForward` call** measured by `--ab-aiseval-diag`. The MHA projection scratch was already pooled, but the inner `ScaledDotProductAttention<float>` still allocated, *every call*:

- the `[B,H,Sq,Sk]` **attention-weights** tensor (≈2 MB) — discarded by MHA (`out _`), and
- a separate **SDPA output** tensor (≈1 MB) — copied into a pooled buffer then dropped.

## Fix

New private `ScaledDotProductAttentionFloatInto`:
- rents the `[B,H,Sq,Sk]` scores from `ArrayPool` (as before),
- runs the numerically-stable softmax **in-place** in that scratch — no separate weights tensor,
- writes `P·V` **straight into a caller-provided buffer** — no SDPA output tensor, no copy.

`MultiHeadAttentionForwardFloat` calls it with the already-pooled `attnHeadBuf` as the destination, so the inverse-transpose reads it directly.

## Result

| | per-call allocation |
|---|---|
| before | ~7.3 MB |
| after | **~1.04 MB** (the inherent returned `[B,seq,dModel]` result tensor) |

**86% reduction** — eliminates the ~3 MB SDPA scratch (plus the rest of the old profile), cutting GC pressure in the inference hot loop and the Transformer RSS footprint (#436 measured 3.5 GB).

## Tests

- Correctness unchanged — the existing `MultiHeadAttentionForward_MatchesDecomposedChain` suite (which exercises the new path) passes: **30/30 net471 + 37/37 net10.0** runnable MHA tests.
- New `MultiHeadAttentionForward_ZeroAllocSdpa_AllocationBelowThreshold` guard: asserts < 2 MB/call on net10.0 (measured 1.04 MB); exercises the path on net471 (`GC.GetAllocatedBytesForCurrentThread` is net5+).

## Scope / follow-up

This closes #476 root cause 1 (allocation/GC/memory). Root cause 2 — managed projection/score **GEMM quality vs native** — is the remaining piece and converges with #475 (small-GEMM kernel quality); tracked as follow-up.


ooples:
### Follow-up: root-cause-2 (projection GEMM quality) — fused QKV projection

Pushed `0ec309fa`. The earlier commit closed root-cause-1 (per-call SDPA allocation); this addresses root-cause-2 (managed projection-GEMM quality).

**Change:** the Q/K/V projections were three separate `[B*seq, dModel] @ [dModel, dModel]` (N=64) GEMMs reading `input` 3×. Fused into **one** `[B*seq, dModel] @ [dModel, 3*dModel]` GEMM (concatenated `[qW|kW|vW]`, built once/call), so: input read once, one dispatch, and a 3× wider N=192 the SIMD microkernel utilizes far better. `TransposeQkv` now reads q/k/v from the interleaved fused buffer via `(srcRowStride, srcColOffset)` — no extra split.

**Result** (best-of-15, `[128,32,64]` h=4):

| | before | + zero-alloc SDPA | + fused QKV | torch |
|---|---|---|---|---|
| **net10.0** | ~5.31 ms | 4.95 ms | **3.06 ms** | ~2.74 ms |

→ **1.6× faster; most of the gap to torch closed.** Correctness unchanged (full MHA suite 31/32 — the 1 skip is the env-gated perf test — on net10.0 + net471, incl. causal mask + non-pow2 seq). net471 ~21.8 ms (no AVX2/FMA intrinsics; no regression).

**Honest standing:** best-of-15 is 3.06 ms vs torch 2.74 ms, so MHA isn't yet *under* torch at p95. The projection GEMMs (the literal root-cause-2) are no longer the bottleneck — the residual is the **SDPA per-head transposes + tiny score/context GEMMs** (512 `[32×16]·[16×32]` slices). Closing that last ~0.3 ms would mean folding the QKV head-transpose into the SDPA read (strided per-slice gather) so the 3 transpose passes + buffers disappear — a larger change to the shared SDPA path; happy to do it as a next step if we want MHA strictly under torch.

ooples:
### Follow-up: transpose-fused SDPA — MHA now UNDER torch ✅

Pushed `3c0fc898`. Folded the per-head Q/K/V transpose into the SDPA loop: each `(b,h)` worker gathers its Q/K/V slice directly from the fused QKV buffer (K gathered transposed), runs `Q·Kᵀ → softmax → P·V`, and scatters straight into the concat layout — removing **3 transpose passes + the inverse transpose + 4 pooled buffers** (and the now-superseded `ScaledDotProductAttentionFloatInto`).

| | zero-alloc SDPA | + fused QKV | **+ transpose-fused SDPA** | torch |
|---|---|---|---|---|
| **net10.0** | 4.95 ms | 3.06 ms | **2.26 ms** | ~2.74 ms |

→ **MHA is now below torch's ~2.74 ms.** #476 goal met (both root causes: allocation + projection/score GEMM quality).

Correctness: `MultiHeadAttentionForward` 31/32 + the 130-test Attention/SDPA set green on **net10.0 + net471** (MatchesDecomposedChain, causal mask, non-pow2 seq, zero-alloc threshold). net471 ~22.5 ms (no AVX2/FMA intrinsics; no regression). Still zero-alloc-per-call (per-slice scratch is ArrayPool-rented).

Full progression on this PR: ~5.31 → 4.95 → 3.06 → **2.26 ms**. Closes #476.

ooples:
Closing — superseded by PR #485 and would regress merged work.

This PR adds exactly the same lean output-only SDPA-in-MHA method (`ScaledDotProductAttentionFloatInto` + MHA wiring) that I shipped today in PR #485 against current main. They're functionally equivalent.

It would also undo merged work: this branch was opened before #472 merged, so its diff *deletes* the `Tests/Helpers/NestedParallelismTests.cs`, `Tests/.../FusedAttentionFp32PrecisionTests.cs`, and 116 lines from `Helpers/CpuParallelSettings.cs` that #472 added. Rebasing would conflict massively and lose the FP32-precision opt-in / nested-parallelism guard.

The CI failure here (`TapePerformanceTests.FullForwardBackward_MLP_Performance` — 136 ms vs 100 ms budget) is a CI wall-clock flake on an unrelated test; no need to investigate since the PR is being closed.

Use PR #485 instead.

Issue #477, opened 2026-05-27
perf(#436): fused sequence recurrent kernel (LSTM, 32 serial small GEMMs)


ooples:
Follow-up from #436 / PR #474.

`LstmSequenceForward` measured ~5.76ms median / 6.63ms p95 vs torch ~2.78ms (PR #474). The MaxDOP sweep (`--ab-aiseval-dopsweep`) shows LSTM is **thread-insensitive** (flat ~5.8ms across 1..16 cores) — the bottleneck is the **unfused per-step recurrent kernel**: 32 serial small `h_prev @ wHh^T` GEMMs ([128,64]×[64,256]), each ~2M FMAs, dominated by per-GEMM dispatch over the sequence. torch routes `nn.LSTM` to a fused sequence kernel (OneDNN/MKL).

**Goal:** LSTM p95 < torch median (~2.78ms) at [128,32,32]→64.

Directions: fuse the per-step gate GEMM across the batch with a persistent packed wHh; vectorize the recurrent step; reduce per-step overhead. Float fast path is in `CpuEngine.LstmSequence.cs` (`LstmSequenceForwardFloat`).

PR #480, opened 2026-05-27
perf(#477): cut LSTM recurrent-step overhead (cached-B wHh + de-interleaved gates)

ooples:
## Summary

Reduces per-step overhead in `LstmSequenceForwardFloat` at the AIsEval shape `[128,32,32→64]` (32-step sequence). Two structural changes, both profiled:

1. **Recurrent GEMM** — pre-transpose `wHh [gateRows,hidden] → wHhT [hidden,gateRows]` **once** and drive the per-step `h_prev·wHhT` through `SgemmWithCachedB` (pre-pack cached on array identity → packed once, reused for all 32 steps), instead of a `transB=true` `Sgemm` that re-transposed/re-packed the constant `wHh` every step.
2. **Gate activations** — combine into a **de-interleaved `[gateType][batch][hidden]`** layout so each activation is **one** vectorized `SimdKernels` call per step (4/step) instead of 4-per-batch (**512/step** at B=128), and `tanh(c)` becomes a single batched call. Cell/hidden update is now a flat contiguous pass.

## Result

| | latency |
|---|---|
| before | ~5.94 ms |
| after | **~5.35 ms** (~10%, stable across runs) |

Correctness unchanged — `LstmSequenceForward_MatchesDecomposedReference` suite: **6/6 on net10.0 + net471**. Adds an env-gated latency bench.

## Honest scope (does NOT reach PyTorch's ~2.78 ms)

Profiling the 5.9 ms breaks down as: serial per-step recurrent GEMM **~2.2 ms** (already ~58 GFLOP/s single-thread) + input GEMM + activations **~1.3 ms** + combine/copy.

- **Parallelizing the per-step recurrent GEMM across the batch was measured ~4× SLOWER** (20 ms) — 32 parallel-for dispatches over tiny ~2M-FMA chunks dominate. This **confirms the issue's "thread-insensitive" finding**, so the recurrent GEMM is kept serial.
- Closing the remaining gap to torch requires a **custom fused recurrent kernel** (what `nn.LSTM` routes to via OneDNN/MKL) — a substantial effort, tracked as the remaining #477 work.

This PR is the clean, low-risk incremental win + the profiling/analysis that scopes the rest.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/CpuEngine.LstmSequence.cs
+++ b/src/AiDotNet.Tensors/Engines/CpuEngine.LstmSequence.cs
@@ -214,6 +214,18 @@ public partial class CpuEngine
         out Tensor<float> finalHidden,
         out Tensor<float> finalCell)
     {
+        // Empty-batch / empty-sequence early return — matches the generic LSTM path's
+        // tolerance of degenerate inputs and shortcuts the per-step GEMM/activation work
+        // (avoids any zero-size buffer surprises and pool churn on a no-op call).
+        if (batch == 0 || seqLen == 0)
+        {
+            finalHidden = AutoTensorCache.RentOrAllocate<float>(new[] { batch, hidden });
+            finalCell = AutoTensorCache.RentOrAllocate<float>(new[] { batch, hidden });
+            return returnSequences
+                ? AutoTensorCache.RentOrAllocate<float>(new[] { batch, seqLen, hidden })
+                : AutoTensorCache.RentOrAllocate<float>(new[] { batch, hidden });
+        }
+
         // Pre-compute Wx = input @ wIh^T as one big GEMM via SimdGemm.
         // input is [B, seq, in] — treat as [B*seq, in] row-major.
         // wIh is [4H, in] — we want input @ wIh^T which is [B*seq, 4H].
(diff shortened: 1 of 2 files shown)


ooples:
### Follow-up: fused batch-parallel recurrent kernel added (closes the gap to torch)

Pushed `acd53a96` — the remaining #477 work this PR had scoped out (the custom fused recurrent kernel). Rather than a hand-written microkernel, the win comes from exploiting that **LSTM batch samples are independent**, so the whole 32-step recurrence parallelizes across the batch with a **single** `Parallel.For` dispatch (each worker runs the full fused recurrence — recurrent GEMM + gate activations + cell update — for its batch slice). The earlier "thread-insensitive" / "batch-parallel 4× slower" findings were artifacts of a per-**step** parallel-for (32 dispatches over tiny chunks); one outer dispatch breaks that wall.

**Results** (best-of-15, `[128,32,32]→64`):

| | before this PR | + incremental (6d111eda) | + fused kernel (acd53a96) | torch |
|---|---|---|---|---|
| **net10.0** | ~5.94 ms | ~5.57 ms | **1.95 ms** | ~2.78 ms |

→ **1.4× faster than torch on net10.0** — the issue's `< 2.78 ms` goal is met. (net471: 56.98 → 46.76 ms; no AVX2/FMA intrinsics so inherently slow, but a win, no regression.)

Correctness unchanged: full `LstmSequenceForward` suite **7/7 on net10.0 + net471** (covers returnSequences, final-state/chunked inference, initial h0/c0, and the small-batch path where a worker handles a single sample). `SgemmWithCachedB` is thread-safe across the parallel workers (`[ThreadStatic]` A-pack + per-thread B-slot caches, lock-guarded shared B cache; small per-slice m keeps each GEMM serial → no nested parallelism).

This supersedes the "Honest scope (does NOT reach PyTorch)" section in the PR description above. Closes #477.

PR #484, opened 2026-05-28
fix: serialize nested BlasManaged GEMMs to stop per-head attention oversubscription

ooples:
## Problem

`ScaledDotProductAttention<float>` at moderate-to-large sequence lengths (e.g. `[1,8,1024,64]`) **thrashed for minutes** — thousands of CPU-seconds without completing — due to ThreadPool oversubscription.

The float SDPA path parallelizes over `B·H` heads via `CpuParallelSettings.ParallelForOrSerial`, which marks each worker thread as inside a parallel region. Inside each head it runs the QK^T and P·V GEMMs through BlasManaged. But the BlasManaged strategies resolved their worker count like this:

```csharp
int procs = options.NumThreads > 0 ? options.NumThreads : Environment.ProcessorCount;
if (options.NumThreads < 0) procs = 1;
```

With `NumThreads` unset (the default), that is `ProcessorCount` **regardless of the parallel region**. So when a per-head GEMM is large enough to multithread, you get `heads × ProcessorCount` workers fighting over `ProcessorCount` cores. The tiny inference shape `[128,4,32,16]` escaped only because its per-head GEMM was too small to trigger multithreading.

`CpuParallelSettings` already *documented* that BlasManaged's parallel sites should consult `IsInParallelRegion` and serialize when nested ("Raw Parallel.For sites (e.g. the BlasManaged GEMM strategies) consult this to fall back to a serial loop when nested, preventing ThreadPool starvation") — it was simply never wired in.

## Fix

Add `CpuParallelSettings.ResolveWorkerThreads(numThreads)`:

| `NumThreads` | resolves to |
|---|---|
| `> 0` | that explicit count |
| `-1` | `1` (deterministic single-thread) |
| `0` (unset), top level | `Environment.ProcessorCount` |
| `0` (unset), **inside a parallel region** | `1` |

Every BlasManaged thread-count decision now routes through it: `BlasManaged.Gemm` (`splitProcs`, `procs`), `PackBothStrategy` (×2), `PackAOnlyStrategy`, `StreamingStrategy`.

**Top-level GEMM behavior is unchanged** (`ProcessorCount` when not nested). Only nested per-head GEMMs collapse to single-thread — which is exactly what the SDPA fallback already did by hand-calling `SgemmSequential`. Numerics are unchanged (thread count only).

## Result

`SDPA float [1,8,1024,64]`: **non-terminating thrash → 20.7 ms** (best-of-3).

## Tests

New `SdpaNestedParallelismTests`:
- `ResolveWorkerThreads` unit cases, including the regression: unset `NumThreads` inside `EnterParallelRegion()` → `1`.
- SDPA float output matches a naive double-precision reference across `[2,4,48,32]` and `[1,8,64,64]`.
- Gated repro `Sdpa_Float_LargeHeads_DoesNotOversubscribe` (`AIDOTNET_RUN_JIT_PERF=1`).

Existing suites green on net10.0 + net471: **532** BlasManaged/strategy, **71** attention (FusedAttention/MHA/SDPA), **427** Gemm/MatMul.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/BlasManaged.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/BlasManaged.cs
@@ -250,15 +250,10 @@ public static class BlasManaged
         // get an inner options with Epilogue stripped to avoid double-application.
         int splitMAligned = m - (m % mr);
         int splitNAligned = n - (n % nr);
-        // Honor NumThreads = -1 as explicit single-thread (deterministic mode).
-        // Pre-fix: -1 fell into the "> 0 is false" branch and used all processors,
-        // breaking determinism/perf expectations. PR #402 CodeRabbit fix.
-        int splitProcs = options.NumThreads switch
-        {
-            -1 => 1,
-            > 0 => options.NumThreads,
-            _ => System.Environment.ProcessorCount,
-        };
+        // Honor NumThreads = -1 as explicit single-thread (deterministic mode); PR #402.
+        // Also collapses to 1 when nested inside a parallel region so the split gate
+        // matches the strategies' single-threaded execution (no oversubscription).
+        int splitProcs = CpuParallelSettings.ResolveWorkerThreads(options.NumThreads);
         if (m >= mr && n >= nr && (m % mr != 0 || n % nr != 0)
             && splitMAligned >= splitProcs * mr)
         {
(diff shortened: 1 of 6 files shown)


ooples:
Closing — this was based on a misdiagnosis. The SDPA float [1,8,1024,64] does not actually hang on current main: it completes in ~27.7 ms because the per-head GEMM is already forced single-threaded by `useSequentialGemm = CpuParallelSettings.IsInParallelRegion` (merged in #472, the current HEAD). The 'thrash' I observed earlier was several leaked zombie testhost processes contending for CPU, not a per-call SDPA hang.

The BlasManaged change here (routing thread-count resolution through `CpuParallelSettings.ResolveWorkerThreads` so the strategies honor `IsInParallelRegion`) is correct and matches the documented contract, but it has no demonstrated beneficiary: the only nested-GEMM caller (the float SDPA) self-protects with its own `SgemmSequential` workaround, which bypasses BlasManaged entirely. So this is defensive hardening, not a bug fix — closing rather than merge a PR whose stated motivation doesn't hold. Can revisit as an honest hardening PR (or to let the SDPA drop its manual workaround) if desired.

PR #490, opened 2026-05-28
DIAGNOSTIC (do not merge): capture #486 pre-pack drift contamination

ooples:
**Temporary diagnostic PR — do NOT merge. Safe to close once we capture a drift.**

The build workflow only runs on PRs to main, so this draft PR exists solely to run CI and capture the intermittent `PrePackedB_Output_BitMatches_LivePack` drift with the new failure instrumentation.

On drift, the failure message now reports:
- `baseVsNaive` / `prepackVsNaive` — which GEMM path diverges from a naive ground-truth reference
- `base2VsNaive` — whether a re-run of the live-pack baseline is clean (transient cross-test contamination vs a deterministic consume-path bug)
- suspect global state at the failure instant: arena active, determinism flag, MKL-verified, JIT kernel count, thread id

Assertion threshold is unchanged (1e-3). Once the smoking gun is captured, this branch's instrumentation (or the resulting fix) lands properly and this draft is closed.


ooples:
Superseded by #492. This draft captured the drift on CI, ruled out coverage instrumentation as the trigger (it reproduced without coverage), and helped pin the contamination to StreamingWorkerPool's shared-slot race. The fix + regression test now live in #492; the diagnostic instrumentation here is discarded.

PR #514, opened 2026-05-30
feat(#477): fused differentiable RWKV-7 sequence kernel (forward + BPTT backward)

ooples:
Adds `CpuEngine.Rwkv7SequenceForward<T>`: the whole RWKV-7 time-mixing WKV recurrence (diagonal decay + rank-1 injection + gated readout, with per-position r/k/v/a/b sigmoids applied internally) over a `[batch, seqLen, modelDim]` sequence as **one** engine op.

Unlike the inference-only `LstmSequenceForward`/`MultiHeadAttentionForward` fused ops, this one is **differentiable**: it records a single tape node whose custom backward runs the BPTT adjoint of the recurrence (gradients w.r.t. r/k/v/a/b), so it is safe under an active `GradientTape`. This eliminates the ~10 per-timestep tape micro-ops the decomposed `RWKV7Block` loop records — the dominant training cost on long sequences (consumed by ooples/AiDotNet#1464).

Double fast path + generic-T fallback for both forward and backward. Exposed on `IEngine`; `DirectGpuTensorEngine` inherits the `CpuEngine` virtual.

**Tests** (net10.0 + net471): forward matches an independent reference recurrence to 1e-10; the custom backward matches central finite differences of `sum(output)` to 1e-5.

Consumed by AiDotNet's #1464 branch (pins `AiDotNet.Tensors 0.86.7`).

### DIFF
--- a/src/AiDotNet.Tensors/Engines/ActivationHandler.cs
+++ b/src/AiDotNet.Tensors/Engines/ActivationHandler.cs
@@ -32,6 +32,28 @@ internal abstract class ActivationHandler
     /// from the pre-activation input, or override to accept it via other means.
     /// </summary>
     public abstract Tensor<T> ApplyBackward<T>(CpuEngine engine, Tensor<T> gradOutput, Tensor<T> input);
+
+    // #506 review: parametric-activation overloads. Most activations are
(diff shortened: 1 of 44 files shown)


ooples:
@coderabbitai full review

Issue #403, opened 2026-05-19
perf: DCGAN training perf plan — autograd alloc, Conv2D small-N, training pre-pack


ooples:
**Related:** ooples/AiDotNet#1390 (DCGAN training 5-10× slower than PyTorch CPU; the `DCGANTests.MoreData_ShouldNotDegrade` test times out at 250 iters)

## Why this is a Tensors issue

`AiDotNet.GenerativeAdversarialNetwork.Train()` orchestrates the training loop but ~95% of wall time per step lives in this repo:

- `Conv2DWithIm2ColDouble` (`Engines/CpuEngine.cs:8793`) — Im2Col + GEMM per batch slice
- `Im2ColHelper.Im2Col`
- `BlasProvider.TryGemm` → OpenBLAS or BlasManaged
- `DifferentiableOps.RecordBinary` + `GradNode` allocation (`Engines/Autodiff/`)
- All backward functions (`Engines/Autodiff/BackwardFunctions.cs`)
- `CompiledBackwardGraph` / `CompiledBackwardWalk` / `CompiledDelegateChain`
- Per-op tape recording closure capture

## Goal

≥3× speedup on the DCGAN<double> training step (519 ms → ≤170 ms) so `DCGANTests.MoreData_ShouldNotDegrade` clears 120 s budget at 250 iters.

## Substep profile from AiDotNet#1390

| Substep | Time | % |
|---|---:|---:|
| Generator.Predict (no-grad fwd) | 19.3 ms | 3.7% |
| Discriminator.Train (fwd + bwd + optim, batch=2) | 186.5 ms | 35.9% |
| Generator adversarial (tape gen+disc fwd + bwd + optim) | 313.3 ms | 60.4% |
| **Total** | **519 ms / step** | |

### Phase A — Investigation (1-2 days, prereq for B-E)

Without this no one knows where the time really goes. Three measurements:

1. **Shape-catalog dump.** Reuse `ShapeInstrumenter` + `BlasProvider.ShapeLogHook` (exists from Sub-A) to record every (M, N, K) GEMM driven by a 1-step DCGAN run. Output a catalog of unique shapes + call counts.
2. **Per-component microbench.** Stopwatch around: Im2Col, GEMM-forward, GEMM-backward-gradInput, GEMM-backward-gradWeight, `RecordBinary`, optimizer `Step()`, BatchNorm forward, BatchNorm backward.
[...]

PR #516, opened 2026-05-30
perf(#403): 3.5× DCGAN<double> conv step — profile-driven GEMM/conv fixes

ooples:
@-


ooples:
@coderabbitai full review

PR #524, opened 2026-06-01
perf: MaxPool2DInto small-batch serial gate (skip parallel-dispatch overhead at bs1)

ooples:
@
## Summary

Profiling the parity CNN inference stem (the non-conv path) surfaced an anomaly: at **bs1, `pool1` MaxPool2DInto = ~44 µs — *more than* conv1 (~37 µs)** despite doing ~100× less work. The cost was **constant across machine load** — the signature of a fixed per-call dispatch overhead, not compute.

### Root cause
The three float `MaxPool2DInto` kernels (2×2 / 3×3 NoPad + Generic) gated the parallel path on **channel count alone** (`bc >= MaxDegreeOfParallelism`). At small batch the pooled tensor is below the parallel grain (bs1: 16ch × 14² = 12 544 < 32 768 `DefaultSerialGrainSize`), so it routed through `ParallelForOrSerial`, ran **serial anyway**, yet still paid the closure allocation + per-channel delegate dispatch — ~25 µs of pure overhead on top of a ~19 µs kernel.

### Fix
Require `bc*hw >= DefaultSerialGrainSize` before taking the parallel branch. Below the grain the existing **direct-serial inline** path (no closure, fully JIT-optimized inner loop) handles it. Output is **bit-identical** — only serial-vs-parallel routing changes. The above-grain (bs128, memory-bound) path is unchanged.

## Measured (bs1, best-of-400)
| pool | before | after | Δ |
|---|---|---|---|
| pool1 16ch×28²→14² | 44.1 µs | 18.6 µs | −25 µs |
| pool2 32ch×14²→7² | 18.7 µs | 11.7 µs | −7 µs |

≈ **32 µs off the parity CNN bs1 Predict** (~272 µs → ~240 µs), closing the CNN-vs-`torch.compile` gap on the non-conv path. (im2col+SimdGemm conv was already shown to be at the managed-CPU ceiling — direct conv loses 2–7×; the remaining headroom was here.)

## Tests
`MaxPool2DIntoSerialGateTests` — 9 bit-exact-vs-allocating-`MaxPool2D` cases across the gate boundary (bs1/bs4/bs128, 2×2/3×3/padded/generic) on **net10.0 + net471** (10/10 pass both), plus an env-gated latency probe.

@

### DIFF
--- a/src/AiDotNet.Tensors/Engines/CpuEngine.cs
+++ b/src/AiDotNet.Tensors/Engines/CpuEngine.cs
@@ -7660,7 +7660,15 @@ public partial class CpuEngine : ITensorLevelEngine
             int hw = h * w;
             int ohow = oH * oW;
 
-            if (bc >= CpuParallelSettings.MaxDegreeOfParallelism)
+            // Only take the parallel (delegate + closure) path when there is enough
+            // total work to amortize its fixed dispatch cost. At small batch the pooled
+            // tensor (e.g. bs1: 16ch × 14² = 12 544 < grain) would otherwise route through
+            // ParallelForOrSerial, run *serial anyway* (work < grain), yet still pay the
+            // closure alloc + per-channel delegate dispatch — measured ~25 µs of pure
+            // overhead on top of a ~19 µs kernel. Below the grain the direct serial inline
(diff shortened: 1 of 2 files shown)


ooples:
@coderabbitai full review

PR #530, opened 2026-06-02
perf(gemm): pack-fresh tiled kernel for small/medium-M float matmul (2–5.8×, clears 75 GFLOP/s)

ooples:
## Problem

Small-M (batch-sized) float GEMMs run at ~12–20 GFLOP/s, far below the ~75–140 GFLOP/s the kernel reaches at large M. Three hot float-GEMM paths route small-M matmuls through native `TryGemm`, whose small-M dispatch leaves cores idle / pays worker-pool wakeup overhead. **`dotnet-trace` of an M=64 GEMM loop: >95% of samples in `StreamingWorkerPool` spin/wakeup, <1% in the AVX kernel.** The packed tiled managed kernel does the same shape 2–5.8× faster.

## Fix

Route small/medium-M (`M ≤ 1024`) contiguous float GEMMs through the packed managed kernel **before** falling to native `TryGemm`, in the three paths that currently miss it:

1. **`CpuEngine.TensorMatMul2D`** (eager matmul / `MatrixMultiply`).
2. **`CompiledTrainingPlan`** matmul delegate.
3. **`CpuFusedOperations.FusedGemmBiasActivationUnchecked`** (the `FusedLinear`/`DenseLayer` forward GEMM — it tried `TryGemm` first and only used the packed kernel when BLAS was *unavailable*).

Above `M = 1024`, native BLAS's raw throughput wins, so `TryGemm` is kept.

### Correctness: pack-fresh, not the identity cache
`SgemmWithCachedB` caches packed B by **array identity**, not contents. In training the optimizer mutates weights **in place** between steps, so the cached path would serve a **stale pre-pack → wrong gradients** (verified directly). These paths use the **pack-fresh** `SimdGemm.Sgemm` when weights can mutate (re-packs every call — correct under in-place updates), and the identity-cached kernel only where B is frozen (inference).

## Benchmark methodology — noisy shared rig

Calibration/measurement was done on a **shared (noisy) machine**, so all GFLOP/s figures are **min-of-N** (best of 12–15 timed windows) — the standard way to read perf on a contended host: the fastest window has the least interference and best estimates the true kernel cost. Mean/single-run numbers are discarded.

## Before/after (min-of-12, `[M×784]×[784×512]`, fresh B every call = training-representative)

| M | before (native) | after (pack-fresh) | speedup |
|---|---|---|---|
| 64 | 14.3 | **82.9** | 5.8× |
| 128 | 27.7 | **88.0** | 3.2× |
| 256 | 33.3 | **105.4** | 3.2× |
| 512 | 50.2 | **119.6** | 2.4× |
| 1024 | 72.5 | **93.0** | 1.3× |
| 2048 | 99.5 | native (> cutover) | — |

Every M from 64–1024 clears **75 GFLOP/s** (M=64 hits 83), and is correct for in-place-mutated weights.

## Correctness verification
- Numerically equivalent to native (float reduction-order only): checksums match to ~7 sig figs; M=512 bit-identical.
- In-place weight updates + FusedLinear: `CompiledTrainingPlanRebindingTests` + FusedLinear/FusedGemm suites — **77 passed / 0 failed** (net10.0); MatMul/Gemm suite **193 / 0**; rebinding **4/4 on net10.0 + net471**.

## Scope / honest note
This routes the three **forward / eager** float-GEMM paths. The FusedLinear **backward** GEMMs (`dW`, `dX`) use transposed `TryGemmEx` and are **not** changed here — a follow-up could give them the same small-M treatment. End-to-end gains therefore vary by workload: GEMM-bound ops (inference, larger models) benefit most; small-MLP *training* steady-state was already near PyTorch because of a large non-GEMM floor (tape replay + optimizer). Cutover (`AIDOTNET_CACHEDB_MAXM`, default 1024) calibrated on `[M×784]×[784×512]`; an `InferenceGemmStrategyAbTests` sweep is recommended to fine-tune per-shape, but min-of-N on the target shapes shows a clean, monotone win 64→1024 with native preserved above.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -340,6 +340,12 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
     // them from a test/diag harness via the static accessors below to break
     // down where Step's wall time is going (forward / grad-zero / backward /
     // optimizer). Zero overhead in hot path when disabled — single env-var
+    // Small/medium-M cutover for routing compiled-plan GEMM delegates through the
+    // pack-fresh tiled managed kernel instead of native TryGemm (see the matmul
+    // delegate factory). At/below this M the managed kernel wins; above it native
+    // BLAS's raw throughput wins. Mirrors CpuEngine's TensorMatMul2D cutover.
+    private const int TrainingSmallMCutover = 1024;
+
     // read at static init, then a single bool check per Step.
     private static readonly bool _profileStepEnabled =
         System.Environment.GetEnvironmentVariable("AIDOTNET_PROFILE_STEP") == "1";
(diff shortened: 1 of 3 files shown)


ooples:
@coderabbitai full review

PR #531, opened 2026-06-02
perf(#531): route ParallelForOrSerial onto the low-latency cooperative pool (default ON)

ooples:
## What this is (was a prototype; now the real fix — and ON by default)

The draft started as an env-tunable keep-alive spin on `StreamingWorkerPool`. Measurement disproved that approach and revealed the actual gap, which this PR fixes: **the low-latency cooperative pool was never wired into the general parallel-op path.**

## The investigation (why the spin knob was a no-op)

Added dispatch/park counters (`--ab-aiseval-poolstats`). Across MLP/Transformer/LSTM, in both native and managed-forced configs:

```
StreamingPool: parallel=0  parks=0      |  Parallel.For dispatches: MLP=1/call, others=0
```

`StreamingWorkerPool` is reachable only through `StreamingStrategy`, which these workloads never select — so tuning its spin window did nothing. The general op path (`CpuParallelSettings.ParallelForOrSerial`) uses **raw `Parallel.For`** (the .NET ThreadPool — the `LowLevelLifoSemaphore.WaitForSignal` that dominated the original trace). The low-latency pool simply wasn't on that path.

## The fix

Route `ParallelForOrSerial`'s parallel path through `CooperativeGemmScheduler` (concurrency-safe via caller-participation), behind a **dedicated switch `CpuParallelSettings.UseCooperativePool`, default ON**. This is decoupled from `CooperativeGemmScheduler.Enabled` (the GEMM-strategy master switch), which stays **off** pending its own concurrency benchmark.

The first cut fanned every dispatch out to `maxDegree` chunks, which oversubscribed the box (cores-1 workers + participating caller + runtime threads) and produced a ~6× p95 tail. A worker-count sweep confirmed pure oversubscription (64K-relu, 32-thread box: 31 active → p95 224µs; 8 → 29.6µs). Fix: **scale the chunk count with the work** (~one chunk per 8K work units) so small ops wake few workers while a large compute-bound op still fans out to `maxDegree`.

## Measured (`--ab-pool-wiring`, 64K-relu, 32-thread box)

| path | median | p95 | alloc/dispatch | run-to-run |
|---|---|---|---|---|
| `Parallel.For` | 28.9µs | 39µs | 2721 B | 28.9 → 52.8 (noisy) |
| **cooperative** | **21.2µs** | **33µs** | **176 B** | 21.3 → 21.2 (stable) |

Wins median, p95, and allocation (**15×** less — no per-call task/range state), and far steadier. The allocation win is structural (every `Parallel.For` call builds task/range state; the cooperative path enqueues to a pooled queue), so it generalizes across shapes; for large compute-bound work both converge.

## Why default ON, and the validation behind it

A proven win shouldn't sit behind a flag nobody flips. With `UseCooperativePool` default-on:
- **Full net10 test suite: 6511 passed.** The only failures were 3–4 *non-deterministic* perf-budget / GPU tests (`MachineCode...BreaksRyuJitCeiling` GFLOPS ceiling, `ConvTranspose2D...BeatsOpenBlasBudget` ms-budget, `MaxPool2DBackward_Gpu`, `NonArena...buffer-reuse`) — they vary run-to-run and **pass in isolation**, i.e. load flakes under full-suite concurrency, not this change.
- It's the **Action overload only** — disjoint-write iterations, so chunking is **bit-identical** to `Parallel.For` (the reduction-style `TLocal` overload is untouched). `ParallelForCooperativeRoutingTests` pins bit-identity, concurrent correctness (8×100), and exception rethrow.
- Escape hatch: `CpuParallelSettings.UseCooperativePool = false` falls back to `Parallel.For` (e.g. for thread-count-sensitive hosts — the cooperative pool keeps `cores-1` dedicated parked workers).

## Scope / safety

- The GEMM-strategy cooperative routing stays **off** (its `Enabled` switch is untouched) until its concurrency benchmark lands — this PR only enables the independently-validated general-op routing.
- A time-based keep-alive (the prototype's idea, on the right pool) was tried and reverted — the tail was oversubscription, not park/wakeup, so it had no effect.
- Diagnostics retained: `--ab-aiseval-poolstats` (where parallelism actually goes) and `--ab-pool-wiring` (the A/B above); `AIDOTNET_COOP_WORKERS` knob for the sweep.
- Both TFMs (net10.0 + net471) build.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/CooperativeGemmScheduler.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/CooperativeGemmScheduler.cs
@@ -95,7 +95,14 @@ internal static class CooperativeGemmScheduler
         lock (_initLock)
         {
             if (_initialized == 1) return;
+            // Worker count. Default cores-1 (the caller participates as the cores-th).
+            // PR #531 tail study: at full subscription the workers + participating caller +
+            // .NET runtime threads exceed the logical cores, so the OS preempts active
+            // workers and a stolen chunk stalls the caller's join — the p95 tail. Tunable
+            // to leave runtime headroom (AIDOTNET_COOP_WORKERS).
             _numWorkers = Math.Max(1, Environment.ProcessorCount - 1);
+            if (int.TryParse(Environment.GetEnvironmentVariable("AIDOTNET_COOP_WORKERS"), out var wOverride) && wOverride >= 1)
+                _numWorkers = Math.Min(wOverride, Math.Max(1, Environment.ProcessorCount - 1));
             for (int i = 0; i < _numWorkers; i++)
             {
                 var t = new Thread(WorkerLoop)
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/StreamingWorkerPool.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/StreamingWorkerPool.cs
@@ -66,21 +66,13 @@ internal static class StreamingWorkerPool
         }
     }
 
-    // PROTOTYPE (option 2 — low-latency hot pool): keep-alive spin window before a
-    // worker parks. Small ops dispatched back-to-back (e.g. a transformer's many
-    // per-layer attention/FFN/layernorm ops, or a training step's 7 GEMMs) otherwise
-    // park between each dispatch and pay the full ManualResetEventSlim wakeup latency —
-    // profiled as the dominant cost (>70% of samples in semaphore/Monitor wait, <1%
-    // in the AVX kernel) for small-op inference. Holding workers hot across the burst
-    // collapses that wakeup latency. Trade-off: more CPU burned while genuinely idle,
-    // so it's tunable and defaults to the original window (1000 spin / 5000 yield).
-    // Multiplier via AIDOTNET_POOL_SPIN (e.g. 20 keeps workers hot ~20x longer); 0/1 =
-    // original behaviour. Validate on a quiet rig with the AB suite before changing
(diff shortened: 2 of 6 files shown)


ooples:
## Evaluation result (local-package full stack, min-of-25, noisy rig)

Tested via local packages (AiDotNet source w/ fused MHA `0.207.14-localfix` + Tensors `0.91.4-localfix` carrying #530 + this) so no release was needed. Swept `AIDOTNET_POOL_SPIN` ∈ {1, 20, 100}, min-of-25 per shape:

| shape | spin=1 | spin=20 | spin=100 |
|---|---|---|---|
| cnn bs1 | 0.31 | 0.56 | 0.13 |
| cnn bs8 | 0.74 | 1.26 | 2.70 |
| cnn bs32 | 1.23 | 2.21 | 6.17 |
| cnn bs128 | 5.18 | 7.40 | 11.2 |
| transformer bs8 | 1.20 | 1.13 | 1.18 |
| transformer bs32 | 3.90 | 3.70 | 3.81 |

**Net-negative as a global knob.** Hot-spinning workers give only marginal transformer gains (~3–5%) but **regress CNN badly** (bs128 5.18→7.40→11.2) — the spinning workers oversubscribe the cores conv's heavy parallelism needs. `spin=1` (default/off) is best overall.

**Conclusion:** the dispatch-latency idea is right, but it can't be a *global* spin — it needs to be **per-op-type / contention-aware** (don't keep workers hot when a parallel-heavy op like conv is also running). Keeping this DRAFT as a documented negative result + starting point; not for merge as-is. The real shipped wins are #530 (GEMM) + the fused-MHA path (already in AiDotNet source).

Issue #560, opened 2026-06-06
FP16 MatrixMultiply<Half> ~1000x slower than FP32: not routed to CudaBackend.Hgemm


ooples:
On RTX 3080, a 2048x2048 FP16 GEMM via IEngine.MatrixMultiply<Half> takes ~56,000 ms vs ~28 ms for FP32 - a ~1000x regression. Root cause: a working tensor-core Hgemm exists (src/.../Engines/DirectGpu/CUDA/CudaBackend.HalfPrecision.cs, SupportsHgemm for CC>=5, calls cublasHgemm) but the public MatrixMultiply<T> GPU dispatch only handles float/double; Half falls through to a scalar path. Fix: route typeof(T)==Half to CudaBackend.Hgemm. Expected: FP16 becomes the FASTEST path (tensor cores), unlocking large speedups for FP16/mixed-precision training. Repro: Ooples.Gpu.Tests BenchmarkTests.Resident_chain_and_fp16.

ooples:
Re-assessment after finding PR #557 (feat/fp16-activation-storage): #557 lands the intended FP16 compute path - AutocastScope(Float16) routes matmul to FP16-in/FP32-accumulate tensor cores (GemmFp16In32fOut / cublasGemmEx), guarded by Fp16MatMulCorrectnessTests. My repro used raw MatrixMultiply<Half>, which is NOT the supported entry point and still falls back to CpuEngine scalar MultiplyBlocked (the ~56s). So the FP16 GPU win is delivered by #557 via AutocastScope; residual is a minor hardening gap - raw MatrixMultiply<Half> could route to Hgemm or fail fast instead of a silent 1000x scalar path. Track under #557.

PR #563, opened 2026-06-06
feat(fp16): route raw Half MatMul to FP16 Tensor-Core GEMM (stacked on #557)

ooples:
Completes a gap that ties into downstream GPU training work: a raw `MatrixMultiply<Half>` on the GPU took the FP32-downcast SGEMM path instead of the FP16 Tensor Cores. This routes Half through #557's `ConvertToFp16` + `GemmFp16In32fOut` (cublasGemmEx, FP16-in/FP32-accumulate) when the backend `SupportsHgemm`, falling through to FP32 otherwise.

**Stacked on #557** (feat/fp16-activation-storage) - depends on its FP16 primitives; retarget to main after #557 merges.

Verification: local full build is blocked by a pre-existing `build/LicenseValidator` netstandard/Cryptography issue in my environment (unrelated). Please let CI validate; correctness is covered by the Fp16MatMulCorrectnessTests pattern (FP16 within tolerance of FP32). Context: issue #560.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/DirectGpu/DirectGpuEngine.cs
+++ b/src/AiDotNet.Tensors/Engines/DirectGpu/DirectGpuEngine.cs
@@ -546,6 +546,32 @@ public sealed class DirectGpuEngine : IDisposable
         // than silently downcasting through the FP32 GPU boundary.
         if (ShouldFallbackForPrecision<T>()) return null;
 
+        // FP16 Tensor-Core fast path (composes with the #557 mixed-precision infra):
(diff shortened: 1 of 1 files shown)


ooples:
Update: builds locally now. The earlier local-build blocker was a transient fresh-worktree bootstrap-restore artifact in build/LicenseValidator (validator NuGet deps not yet restored on a brand-new tree); after the validator's packages restore, a normal `dotnet build AiDotNet.Tensors.csproj -c Release` (no enterprise disable-flags) succeeds and **this FP16 change compiles cleanly with 0 errors**. No license-gate or netstandard code change was needed.

Issue #558, opened 2026-06-06
FP16 activation storage: fused-plan heterogeneous rewrite (perf follow-on to #557)


ooples:
Tracking the multi-week follow-on to #557: give the fused `CompiledTrainingPlan` mixed-dtype (FP16 activation storage) support for compile-once-replay PERFORMANCE. The mechanism + the memory win (measured 0.500x on GPU) already landed via the lazy-graph path in #557; this is the optimization to run the win on the fused/CUDA-graph path.

Single-type today: `CompileTraining` (~line 2005) only emits `CompiledStep<float>` and skips `CrossTypeLazyNode`/`LazyNode<Half>`; fused groups / IL backward walker / CUDA-graph capture all assume one element type.

Phases (each behind `AIDOTNET_FP16_ACTIVATIONS`, default FP32 fused path byte-identical):
- **A** heterogeneous compiled forward â€” `MixedPrecisionCompiledPlan`, dtype-tagged replayable step list + stable FP32/FP16 buffers. Gate: forward == eager realize.
- **B** compiled mixed-dtype backward â€” capture `MixedPrecisionGraphBackward` topo + per-node backward, stable dual grad buffers. Gate: == non-compiled backward.
- **C** optimizer + loss scaling â€” FP32 master weights + `GradScaler` in `Step()`. Gate: compiled training descends loss, 0 NaN.
- **D** GPU + perf â€” `Hgemm` realize, stable GPU buffers, CUDA-graph capture. Gate: d512/L6/B256 resident ~1/2 vs FP32 baseline (~5105 MiB @ B64), loss-parity, 0 NaN.
- **E** AiDotNet routing â€” `NeuralNetworkBase.TrainWithTape` / `CompiledTapeTrainingStep` route to the mixed-dtype plan under the flag.

Full design: docs/fp16-activation-storage-design.md. A-C CPU-testable; D-E GPU-bound. Est ~2-4 weeks.

Generated with Claude Code.

ooples:
Progress: Phases A-D landed on feat/fp16-activation-storage (CPU + GPU verified), all behind AIDOTNET_FP16_ACTIVATIONS, default FP32 fused path byte-identical.

- A heterogeneous compile-once/replay forward (MixedPrecisionCompiledPlan)
- B compiled mixed-dtype backward over the captured order (shared BackwardOverOrder)
- C compiled training Step (SGD on FP32 masters + GradScaler, skip-on-overflow)
[...]

PR #572, opened 2026-06-08
ci(release): don't let the oversized CUDA package block the managed NuGet publish

ooples:
## Summary

Fixes the **release blocker** that has left every post-FP16 Tensors version unpublished — which in turn blocks AiDotNet PRs (e.g. #1513) that consume the new mixed-precision API.

## Root cause

The `Automated Release Pipeline` → `Push to NuGet` step loops over `out/*.nupkg` calling `dotnet nuget push`. NuGet.org rejects packages over **250 MB** with **HTTP 413**, and the native `AiDotNet.Native.CUDA` package (cudart + cuBLAS + the nvRTC redist added in #559) now exceeds that limit.

GitHub Actions runs `bash` with `set -e`, so the failing `dotnet nuget push` on the oversized CUDA package **aborted the entire step** before the existing `if [ $? -eq 0 ]` warning could run. Result: the managed **`AiDotNet.Tensors`** package — the one downstream consumers actually pin — never got pushed.

Concretely, the release for `821316e7` ("FP16 path … [#558]") failed exactly this way:
```
Pushing AiDotNet.Native.CUDA.0.92.0.nupkg ...
error: Response status code does not indicate success: 413 (The package file exceeds the size limit...).
##[error]Process completed with exit code 1.
```
So `AiDotNet.Tensors 0.92.0` (which exports `MixedPrecisionCompiledPlan`) is **absent from NuGet** despite building and testing green — and AiDotNet #1513 fails to compile with `CS0246: MixedPrecisionCompiledPlan could not be found` against the latest published 0.91.12 (verified: the type is not in 0.91.12's DLL).

## Fix

In the push loop:
- **Skip packages over 250 MB** with a `::warning::` — they can't go to nuget.org and already ship as GitHub Release assets (the `github-release` job attaches `out/*.nupkg`).
- **Make per-package pushes non-fatal** (`set +e` around the loop, aggregate failures), so one bad package can't strand the rest. The job still fails if an *under-limit* package fails to push.

The oversized native CUDA package can no longer block the managed package.

## After merge

This push to `main` triggers the release pipeline, which will publish the managed `AiDotNet.Tensors` package (skipping the oversized CUDA package). Once that version is on NuGet, AiDotNet #1513's pin can be bumped to it and it will compile.

## Verification

- `bash -n` clean on the rewritten step; workflow YAML parses.

### DIFF
--- a/.github/workflows/automated-release.yml
+++ b/.github/workflows/automated-release.yml
@@ -498,7 +498,27 @@ jobs:
             exit 1
           fi
 
+          # NuGet.org rejects packages larger than 250 MB with HTTP 413. The native
+          # CUDA package (cudart + cuBLAS + nvRTC redist, ~hundreds of MB) exceeds
+          # that and CANNOT be pushed to nuget.org — it ships as a GitHub Release
+          # asset instead (the github-release job attaches out/*.nupkg). Previously
+          # its 413 aborted the whole step (GitHub Actions runs bash with `set -e`,
+          # so the failing `dotnet nuget push` killed the loop before the status
+          # check), which blocked the managed AiDotNet.Tensors package — the one
+          # downstream consumers actually pin — from ever publishing. Skip
+          # over-limit packages with a warning and make per-package pushes
+          # genuinely non-fatal so one bad package can't strand the rest.
+          NUGET_MAX_BYTES=$((250 * 1000 * 1000))
+
+          set +e
+          failed=()
           for pkg in "${pkgs[@]}"; do
+            size=$(stat -c%s "$pkg")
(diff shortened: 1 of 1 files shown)


ooples:
Superseded by #570 (now merged into main, commit fd791d87). #570 fixes the release blocker at the root: the CUDA runtime now ships to nuget.org as TorchSharp-style <250 MB `AiDotNet.Native.CUDA.Fragments.*` packages that reassemble byte-for-byte at the consumer's build, so every package is under the 250 MB limit and the managed `AiDotNet.Tensors` package publishes normally — no package is skipped. This PR's approach (skip the oversized CUDA package / ship it only as a GitHub Release asset) was the band-aid that prompted building #570 instead. Closing as redundant.

PR #581, opened 2026-06-10
WIP: GPU-resident compiled-plan execution for CUDA-graph training capture

ooples:
## Summary (DRAFT — work in progress)

Makes the compiled training-plan forward pass **fully GPU-resident** so the whole training step is **CUDA-graph-capturable** (one `cuGraphLaunch` per step) instead of falling back to the launch-bound eager path (~7–14% util). On the token-LM cortex this takes the capture from **aborting at the first op → running through ~95% of the forward**, with GPU util climbing **7% → 29–41% peak**. **Loss is verified correct (9.62, no drift) at every step.**

## Root cause

Every compiled-plan op did a host round-trip — **download its result**, **upload a non-resident input** (`GetDataArray` forces DtoH; an un-cached param HtoD's), or **fresh `AllocateOutputBuffer`** (`cuMemAlloc`). All three are `CUDA_ERROR_STREAM_CAPTURE_UNSUPPORTED` (900) inside a stream capture — and one 900 silently invalidates the whole graph (every op after gets a 901 "victim" error). The fix routes each op's GraphMode `Execute` through a **resident `*Into` path**: resident inputs → `backend.<op>` into a **stable cached output buffer** → mark resident (`BindResidentBuffer`), gated on the compiled step (`EvictionSuspended`).

## What's done (committed, loss-parity verified)

- Central `CopyResultInto` residency alias (183 host-copy sites) + `BindResidentBuffer` / `GetOrCreateResidentBuffer` / `GetResidentOrPersistentInputBuffer` helpers
- Resident ops: **embedding-from-float-indices** (the chain source), FusedLinear weights to persistent cache, permute/transpose, binary/scalar elementwise, **LayerNorm** (plus a real correctness fix: `normSize` must be the last dim, not `dims[0]`, for rank-3 `[B,S,D]`), BroadcastAdd (positional), TensorMatMul, SliceAxis
- 194 `GetOrAllocateBuffer(X.GetDataArray())` to `(X)` input-resolution sites (resident-first, no forced download)
- `_gpuBuffer`-clearing paths gated by `!EvictionSuspended` (don't orphan alias chains; fixes a #226 "buffer released" race)
- Permanent diagnostics: `AIDOTNET_GRAPH_CAPTURE_DEBUG=1` capture-fail reason+stack, `[CAPTURE-INVALIDATED-BY] forwardAction#N op=X`, producer-tracking

## Remaining (why this is a draft)

- One untagged/host `[B,D]` producer near the LM head still cascades a host result through a reshape
- The **backward half** (`_backwardActions`) is untouched — same treatment needed
- Graph capture stays **opt-in** (`AIDOTNET_CUDA_GRAPH_STEP=1`); default path unchanged

Repro: `d256/L2/12K` cortex with `HE_RUN_HEAVY=1 AIDOTNET_CUDA_GRAPH_STEP=1 AIDOTNET_GRAPH_CAPTURE_DEBUG=1`.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/CooperativeGemmScheduler.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/CooperativeGemmScheduler.cs
@@ -100,7 +100,7 @@ internal static class CooperativeGemmScheduler
             // .NET runtime threads exceed the logical cores, so the OS preempts active
             // workers and a stolen chunk stalls the caller's join — the p95 tail. Tunable
             // to leave runtime headroom (AIDOTNET_COOP_WORKERS).
-            _numWorkers = Math.Max(1, Environment.ProcessorCount - 1);
+            _numWorkers = AiDotNet.Tensors.Helpers.CpuParallelSettings.WorkerPoolThreads;   // capped (was ProcessorCount-1)
             if (int.TryParse(Environment.GetEnvironmentVariable("AIDOTNET_COOP_WORKERS"), out var wOverride) && wOverride >= 1)
                 _numWorkers = Math.Min(wOverride, Math.Max(1, Environment.ProcessorCount - 1));
             for (int i = 0; i < _numWorkers; i++)
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/StreamingWorkerPool.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/Pool/StreamingWorkerPool.cs
@@ -20,7 +20,7 @@ internal static class StreamingWorkerPool
 {
     [ThreadStatic] private static bool _isExecuting;
 
-    private static readonly int _numWorkers = Math.Max(1, Environment.ProcessorCount - 1);
+    private static readonly int _numWorkers = AiDotNet.Tensors.Helpers.CpuParallelSettings.WorkerPoolThreads;   // capped (was ProcessorCount-1)
     private static readonly WorkerSlot[] _slots = new WorkerSlot[_numWorkers];
     private static readonly Thread[] _workers = new Thread[_numWorkers];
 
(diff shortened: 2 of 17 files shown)


ooples:
## ✅ UPDATE: the CUDA-graph whole-step capture now ENGAGES (core goal met), correctness verified

`[GRAPH-CAPTURE] launch() THREW` count is now **0** — the **entire training step (forward + backward) captures as a single `cuGraphLaunch`**. Verified on the d256/L2 token-LM cortex.

**Correctness preserved (loss parity vs eager):** graph-ON `9.6325 / 9.5625 / 9.4191` ≈ graph-OFF `9.6341 / 9.5657 / 9.4159` over 3 epochs (within FP-reduction nondeterminism; the model learns, `pL1=127`).

### The keystone fix
The cortex's embedding is `TensorEmbeddingLookupFromFloatIndices`, which (a) was not mapped to `OpType.Embedding` so the in-graph **index-refresh machinery never engaged**, and (b) ran a pure host gather. Fix: map it → `OpType.Embedding` + make the float-indices resident path mirror `TryEmbeddingGatherGpu` (indices via the stable `_cachedEmbIndexBuffer` re-uploaded **outside** capture before each replay; table via the persistent weight cache; output left GPU-resident). Final blocker after that: the positional BroadcastAdd resolving its bias via `GetDataArray()` (fired a deferred materializer → DtoH inside capture) → resolve via `GetResidentOrPersistentInputBuffer` instead.

### Honest util note
On the tiny test config util peaks ~26% / ~147 W. The capture **engaging** is the fix here; high utilization is an orthogonal lever — the RTX 3080 is power-capped (needs `nvidia-smi -pl 350`) and a launch/data-light test config doesn't amortize the graph's launch-overhead saving. Worth confirming at a realistic config.

### Remaining before "ready"
- Confirm capture + speedup at a realistic (compute-heavy) config
- Consolidate with the pool-cap branch; squash the WIP history

ooples:
## ✅✅ 100% GPU utilization confirmed — goal achieved end-to-end

With the capture engaged + the full residency refactor + a compute-real config (d512/dff2048/L4), the cortex trains at **100% GPU utilization, sustained (22/22 samples at 100%)**, ~165–175 W.

The earlier ~26% was purely the tiny d256/L2 test config (launch/data-light). The power cap turned out to already be at 350 W (max) — power was never the live limiter; the limiters were (a) the capture aborting [now fixed] and (b) config compute-density. Once a realistic config runs with the whole step captured as one `cuGraphLaunch`, the GPU saturates.

Loss stays correct/learning. This closes the core objective: **the CUDA-graph whole-step capture engages and the cortex trains at 100% GPU util.**

ooples:
@coderabbitai resolve — the rethrow is now a StepEager() fallback (state rolled back, graph disabled).

Issue #26, opened 2026-03-04
perf: comprehensive GPU performance audit - beat cuBLAS/PyTorch across all operations


ooples:
# GPU Performance Comprehensive Audit Report

## Executive Summary

A deep analysis of all GPU tensor operations reveals **systemic performance bottlenecks** across the entire DirectGpu engine stack. The codebase has impressive breadth -- custom CUDA, OpenCL, and HIP kernels covering GEMM, convolution, attention, normalization, pooling, optimizers, and more -- but nearly every kernel has critical performance issues that leave 4x-16x performance on the table compared to cuBLAS, PyTorch, and vendor-optimized libraries.

**No GPU benchmarks have ever been run and recorded.** All GFLOPS numbers in the code are aspirational targets or hardcoded competitor estimates. The 25,000 GFLOPS target for AMD RX 6800+ has never been validated.

### Root causes fall into 6 categories:

1. **Tiny tile sizes** -- CUDA/HIP fused GEMM uses 16x16 tiles (should be 128x128), wasting 98% of shared memory capacity
2. **No tensor cores** -- Zero WMMA/mma.sync usage in any CUDA kernel, leaving 4-8x throughput on the table for Volta+ GPUs
3. **CPU-GPU transfer epidemic** -- Every operation uploads, computes, downloads; no GPU residency between operations
4. **Fake async transfers** -- `UploadBufferAsync` immediately calls `cuStreamSynchronize`, eliminating all overlap
5. **Serial reductions** -- Normalization/attention kernels use 1 thread per channel/batch with serial loops, giving ~3% GPU occupancy
6. **No shared memory** -- Convolution, attention, and normalization kernels read everything from global memory repeatedly

---

### 1. CUDA Fused GEMM: 16x16 Tiles Are 1990s-Era Design

**Files:** `src/.../DirectGpu/CUDA/Kernels/CudaFusedKernels.cs`, `src/.../DirectGpu/HIP/Kernels/HipFusedKernels.cs`
**Impact:** ~8-16x throughput loss vs optimal tile sizes

```c
#define TILE_SIZE 16
__shared__ float As[TILE_SIZE][TILE_SIZE];  // 16x16 = 1 KB per tile
__shared__ float Bs[TILE_SIZE][TILE_SIZE];  // Total: 2 KB used
```

[...]

PR #584, opened 2026-06-11
Scope resident-buffer behavior to the compiled capture path (fixes graph-mode per-step VRAM leak)

ooples:
## Summary
Fixes the per-step VRAM leak that OOM'd long training epochs under CUDA-graph capture (follow-up to #581).

**Defect:** the GPU-resident op branches gated on `EvictionSuspended`, which is true for the whole graph
**lifetime** — so the non-captured per-step work (the eager optimizer update, warmup steps, shape-mismatch
eager fallbacks) also routed into resident allocations. Every step leaked pinned VRAM with nothing evictable.
Measured: a 76-step canary trained clean; a 488-step epoch OOM'd mid-epoch-1 at two model sizes (ff2048 and
ff1024); the v4 forensics showed an 11.9/12.0 GB plateau with the pool slowly exhausted ("drained 10" -> "drained 0").

**Fix:** a thread-static `CompiledCapturePathScope` set by `CompiledTrainingPlan` around only the
pre-residency pass + the `cuStreamBeginCapture` body. Resident branches now gate on
`ResidentStepActive` (= in-scope AND eviction suspended); all off-path work takes the normal transient path.
(Self-review follow-up: the capture-path flag is INSTANCE-scoped, not `[ThreadStatic]`, because op execution
fans out to the BLAS pool threads.)

**Diff hygiene:** `DirectGpuTensorEngine.cs` had been re-saved with LF line endings while main's blob is CRLF,
which had inflated this PR's diff to ~38k lines of pure EOL noise (and would have conflicted with every other
branch touching the file). The file's CRLF endings are restored; the real change is **46 insertions / 11
deletions** across the two files.

**Verification:** the `CompiledCapturePathScope` / `ResidentStepActive` gating is the only behavioral change and
builds clean on net10.0 + net471. The full GPU long-run confirmation (d256/L2 graph smoke for loss parity +
THREW=0, and a 488-step-epoch run with the graph re-enabled) must be run on the CUDA runner before merge — it
cannot be exercised on a CPU-only host.


ooples:
## Progress update — 6 real defects fixed in this series; capture re-enablement on 0.92.x has one route left

Landed (each verified by the d256/L2 smoke advancing past its predecessor):
1. **Capture-path scoping** (the headline #26 fix): resident-buffer behavior gated to the pre-residency/capture body via an instance-scoped depth (NOT `[ThreadStatic]` — op execution fans out to BLAS pool threads where a thread-local is invisible; caught by smoke).
2. **Rollback materializer flush**: both capture-failure rollbacks now `FlushAllAnyThread()` before resuming eviction (pool-thread registrations survived the this-thread-only drain → "released before materialization" killed the eager fallback in 12s).
3. **General-permute meta-buffer cache**: per-call `AllocateIntBuffer` (cuMemAlloc+HtoD) for strides/permutation aborted any capture containing a rank-4 permute (the attention head-split). Now content-keyed, warmed outside capture.
4. **Batched resident matmul**: rank≥3 × rank≥3 via `BatchedGemm` (merged main routes the attention matmuls through `TensorMatMul`).
5. **Dispatch-class fix**: GraphMode recordings calling `cpuEng.XInto(...)` directly bypassed the GPU engine's explicit-interface resident paths (`DirectGpuTensorEngine` IS a `CpuEngine`, so `eng is CpuEngine` matches it). Routed via `IEngine` for the ops the interface declares.
6. **Allocating `TensorMatMul` capture branch**: main's BLAS routing records `eng => eng.TensorMatMul(...)` actions; the host result + download aborted the capture.

**Remaining**: one matmul route still reaches `TensorMatMulFloatInto` during capture; the gate-state diagnostic is dedupe-blinded (identical strings from warmup suppress the in-capture occurrence). Next: per-phase diag counters, then the long-run (488-step epoch) verification that gates this PR's ready state.

**Workaround in production meanwhile**: the cortex A(N) sweep runs graph-OFF (eager-fused, correct loss) — training is not blocked.

ooples:
## Forensics addendum: the eager-path OOM is a per-step transient RETENTION leak (not cache hoarding)

Five independent eliminations on the d512/L4/ff1024/V15001 cortex (0.91.108, graph OFF, box otherwise idle):

1. Batch size B=64→B=32: **identical OOM** at the identical ~21.5-min mark (v7: 1289s, v15: 1292s walltime)
2. `actCacheSize` 131072→16384 (verified applied in-log): **identical OOM**, faster (~4 min into training)
3. The failing request is always **33554432 bytes ≈ a d512×V15001 fp32 buffer** (the output-projection gradient/transposed-weight size class)
4. `drained 11 pooled buffers; device free=0` — the pool can only reclaim 11 buffers; a recurring identical-size transient should recycle through the pool indefinitely, so the prior instances **must still be referenced** (tape/plan history retaining per-step grads?)
5. Desktop headroom and farm contention eliminated separately (box drained, 8.4 GB free at start)

This is the **eager-path face of #26**: the graph path leaked via eviction-suspension; the eager path leaks via per-step retention of the [d,V] transient. Currently testing `HE_GPUSCOPE=1` (deferred-download scope) as a lifecycle workaround; if it fails, the leak hunt needs an allocation-site dump (tag pool allocations by size-class + stack once, count live instances at the 10-min mark).

PR #586, opened 2026-06-11
0.91 backport: producer-map retention leak + activation-cache VRAM cap override (gcdump-proven)

ooples:
## What this PR actually contains

This branch was originally cut from `b11e4dc` (the verified 0.91.108 tip) to ship two targeted host-memory hotfixes to the 0.91 training line for the HRE A(N) sweep. Subsequent work on the GPU-residency / CUDA-graph-capture path landed on top of the same branch (rather than being moved to #584), so this PR now ships **both** layers together. Listing them honestly so reviewers know what they're looking at:

### Layer A — 0.91 host-heap hotfixes (gcdump-proven, the original 0.91.110 / 0.91.111 scope)

1. **Producer-map retention leak** — `s_producerOf` (a capture-time diagnostic that maps a compiled-plan output's backing array → producing op name) was written on every compiled-forward output with strong array keys, never evicted. gcdump at minute 10 of a sweep: **7,207 × 4 MB `Single[]` = 28.8 GB live heap**. GC pressure starved the finalizer-based device reclaim and `cuMemAlloc` started OOMing regardless of batch size, cache cap, GPUSCOPE, or farm state. Now env-gated behind `AIDOTNET_GRAPH_CAPTURE_DEBUG=1` and hard-capped at 4096 entries.
2. **Activation-cache VRAM cap over-commit** — `_maxActivationCacheBytes = GlobalMemoryBytes * 3/4` sizes off **total** device memory and is blind to other processes (the Windows desktop holds ~4 GB on a typical workstation). The cache saturated toward 9 GB of an effectively-8.4 GB card and per-step transient activations OOM'd at a fixed step horizon. New env override `AIDOTNET_ACT_CACHE_VRAM_MB` lets a sweep pin the cap to whatever leaves headroom for transients. The proper root-cause fix (size off `cuMemGetInfo` free at init) is follow-up on the main line.

### Layer B — GPU residency & in-graph capture (the bulk of the diff, came in after the hotfix base)

This is the work that turns the eager-prefix GPU forward into a fully captured CUDA graph. Was not in the original PR description because the branch was carrying it as a deployment vehicle while the main-line work in #584 / #581 settled. Now that the main-line resident-buffer scoping has merged (and the eager-fallback VRAM-leak fix in `a86ede3f` and the GPU-missing-kernel audit in `70e761f4` are present), this layer is built on top of that and the conflicts with main are resolved (commit `42baf0e3`).

Material changes in this layer:
- **CUDA-graph capture FULLY ENGAGES end-to-end** — BroadcastAdd bias now resolves resident (no `GetDataArray` round-trip inside capture), GPU-resident MatMul / SliceAxis / LayerNorm / permute / transpose, persistent param/constant binary inputs, FusedLinear with persistently cached weights+bias.
- **On-device cortex embedding gather** — embedding lookup runs INSIDE the captured graph against a stable index buffer refreshed each step via `RefreshGraphEmbeddingIndicesNow`, removing the per-step `cuMemAlloc` that was the util killer (7% → 90%+).
- **GPU-resident compiled-plan execution** — 194 input-resolution sites in the compiled-plan ops now stay resident through the step; binary/scalar ops, permute/transpose, LayerNorm, BroadcastAdd, embedding-from-float-indices, slice, MatMul are all on the resident path.
- **Dataflow fusion + analytic specs disabled on the GPU engines** — they were doing CPU work; the residency-aware path supersedes them.

### Layer C — Byte-cap eviction race fix (commit `e0a97f26`, the "known remaining" item now closed)

The PR previously documented a `CUDA error 700` race in `EvictOldestActivationsUnsafe` when the byte cap was pinned below the working set (~4 GB on an 8 GB card). Root cause: the compiled-plan `StepEager` already brackets its forward+backward with `SuspendActivationEviction` so the byte-cap eviction path can't run mid-step and free buffers a still-queued kernel will reference; **eager training under a tape didn't have the same atomic bracket**. The outermost tape on a GPU engine matches the "one step" granularity (every activation it caches is consumed by its own backward), so freeing them mid-step is exactly the race that surfaces.

Fix: `GradientTape`'s ctor for the outermost tape on a GPU engine now calls `SuspendActivationEviction`; `Dispose` pairs it with `ResumeActivationEviction` **before** the deterministic-per-step `EvictActivationsCreatedAfter` (otherwise eviction would still be suspended when the deterministic eviction runs and the step's activations would linger). The deterministic per-step eviction takes over the memory-bounding role the mid-step eviction was trying to play. Net effect: byte-cap overflow becomes a clean CUDA OOM (reduce batch / enable gradient checkpointing / use mixed precision) instead of a 700 race.

**Test.** `BytecapEvictionRaceTests` is a CUDA-required stress harness (`[SkippableFact]` + `Skip.IfNot(CudaNativeBindings.IsAvailable, ...)` so CI without CUDA skips cleanly). Pins the byte cap below the per-step working set via reflection on `_maxActivationCacheBytes` (8 MB and 4 MB cases), runs 200/100 cap-overflowing matmul steps, and verifies (a) no CUDA error emerges on `Synchronize()`, (b) `_currentActivationCacheBytes` stays non-negative (the double-decrement signal that previously surfaced when the byte-cap eviction path raced its own bookkeeping).

### Merge status

- Was conflicting with main against the resident-buffer scoping that #581 review introduced. Conflicts resolved in `42baf0e3` by taking main's safer versions (the `EnterCompiledCapturePath` scoped block, the `ResidentStepActive` predicate that requires the capture-path depth, the CUDA buffer-handle safety guards in `UploadBufferInPlace`/`UploadIntBufferInPlace`) while keeping HEAD's `s_producerOf` env-gated diagnostic hotfix.
- 0 build errors on both `net10.0` and `net471` after merge.
- The CUDA-700 race that was previously documented as a workaround-only known limitation is now fixed (`e0a97f26`) with a CUDA-required repro test.

### Why one PR and not two

Splitting the branch now would re-test both layers independently, doubling the burn while the 0.91 line is mid-sweep. The two hotfix commits (`eaf5905d`, `39bc32d7`) are also already in `fix-graph-vram-creep` (#584) for the main line per the original commit messages, so the 0.91 deployment can ship from this branch and #584 carries the same fixes forward for 0.92.


ooples:
## v20 verification run (6GB cap): 2 hours of training — the deepest run yet — then a NEW failure class

The two memory fixes held (no OOM; GPU never saturated). At ~2h, **AggregateException: deferred GPU-to-CPU materialization callbacks failed** — a free site that bypasses the materialize-then-free contract (the cache evictor itself has the #226 fix; this is a different site, likely the scope-end bulk drain racing a pool DrainAll or the deterministic per-step release). Next session: enumerate every `buf.Dispose()/cuMemFree` site and audit each against `DeferredArrayMaterializer.IsPending`. The ready-gate for this PR remains a clean epoch-1 completion at N=250K.

PR #609, opened 2026-06-15
Default the stream-ordered CUDA allocator ON — fixes a sticky CUDA-700 use-after-free under VRAM-pressure buffer churn

ooples:
## Problem

The custom GPU buffer pool frees device memory with **synchronous `cuMemFree`** — on pool-bucket overflow (`GpuBufferPool.Return`) AND via the `CudaGpuBuffer` finalizer — the instant the host requests it. But a kernel launched earlier may not have **executed** yet (work is queued async on `_stream`). That kernel then reads/writes the freed memory: a **use-after-free** that faults the context with a sticky **CUDA 700 (illegal address)**.

It reproduces only on large models under VRAM-pressure buffer churn (a d768/L6 transformer; eviction + transient-buffer churn fills the pool buckets so the overflow `cuMemFree` fires while kernels are in flight) and **vanishes entirely under `CUDA_LAUNCH_BLOCKING=1`** — the signature of an async free-before-execute race. Making the lease/pool stream FLAGS blocking does NOT fix it (the free is host-side, not stream-ordered); only stream-ordered freeing does.

Diagnosis path: a non-blocking `cuMemGetInfo` ring-buffer probe localized the fault to a deterministic, late, non-GEMM buffer-management window; `CU_STREAM_NON_BLOCKING→DEFAULT` (blocking streams) did NOT fix it while `CUDA_LAUNCH_BLOCKING=1` and `AIDOTNET_CUDA_ASYNC_ALLOC=1` both did — isolating it to host-side synchronous freeing.

## Fix

Default the existing `AIDOTNET_CUDA_ASYNC_ALLOC` path **ON** (opt out with `=0`). It allocates/frees via **`cuMemAllocAsync` / `cuMemFreeAsync` on `_stream`**, so every free is enqueued **after** the in-flight kernels on that stream — the race is eliminated by construction. The existing driver-support probe (`cuDeviceGetDefaultMemPool`) falls back to the legacy sync path on older drivers / no pool support.

Plus: trim the stream-ordered pool (`cuMemPoolTrimTo`, newly bound) on a **sync-path `cuMemAlloc` OOM** in `AllocDeviceMemoryWithRetry`. The pool's caching retention (release threshold maxed for reuse) otherwise reports device `free=0` to the legacy allocator, so a mixed async+sync allocation would spuriously OOM.

## Verification

- A d768/L6 token-LM (GPT-2-small scale) that previously fell back to the eager path with a sticky 700 part-way through training now **trains fused across all epochs** — `eagerFallback=False`, no 700, no crash.
- Full Tensors test suite: **7053 passed**. The 18 failures are pre-existing and reproduce identically with the allocator forced off (`AIDOTNET_CUDA_ASYNC_ALLOC=0`): perf-budget flakiness (`*_Delivers_Speedup`, `*_BeatsOpenBlasBudget`), OpenCL tests (this change is CUDA-only), and a 4th-decimal float-tolerance assertion (`BatchedGemmFanout` vs strided accumulation order).

## Release Notes

* **New Features**
  * Async GPU memory allocation enabled by default for improved performance

* **Bug Fixes**
  * Enhanced GPU memory recovery when out-of-memory scenarios occur

### DIFF
--- a/docs/fp16-activation-storage-design.md
+++ b/docs/fp16-activation-storage-design.md
@@ -113,7 +113,7 @@ Root cause (bottom of the stack): compression allocates the FP16 buffer but free
 
 ## ⚠️ SEVENTH / CONCLUSIVE FINDING — convert-based compression can't reduce peak at any scale; the win needs FP16-NATIVE kernels
 
-Layer 6 (stream-ordered allocator, `cuMemAllocAsync`/`cuMemFreeAsync`, race-safe + pool-reuse, pool release-threshold maxed) IMPLEMENTED + committed (`AIDOTNET_CUDA_ASYNC_ALLOC`, default off). Measured the full stack (ACTIVATIONS+PAGING+GPU_CACHE+ASYNC_ALLOC) at BOTH scales on the cortex:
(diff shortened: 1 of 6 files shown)


ooples:
Addressed in `aa0b43c`:

1. **(critical) Silent fallback to the unsafe sync allocator** — fixed. The env var is now tri-stated: `=0` is a deliberate, silent sync opt-out; **explicitly set but no stream-ordered mem-pool support → throws `NotSupportedException`** (we don't substitute the use-after-free path behind a caller who asked for the guarantee); **default (unset) + no pool support → falls back to sync but emits a `Trace.TraceWarning` naming the residual CUDA-700 risk** (so older drivers still run, but it's no longer a silent "looks safe"). 
2. **(critical, 320) same fallback block** — same fix.
3. **(minor) stale default-off docs** — updated the `_asyncAlloc` field comment and `docs/fp16-activation-storage-design.md` to the default-on contract.

Build green (Release net10.0).

PR #632, opened 2026-06-17
fix(compile): fix BlasBatch hoist vs MemoryPlanning aliasing in attention inference (remove #1624 gate)

ooples:
## Summary

Makes compiled attention/diffusion **inference** correct *and* fast by fixing the real root cause and removing the temporary `#1624` correctness gate.

The draft originally **gated** the step-mutating passes (OperatorReordering + MemoryPlanning) off for any plan containing `Softmax`/`LogSoftmax`/`ScaledDotProductAttention`. That masked the miscompile but cost every attention/diffusion inference plan its peak-memory reduction, and it blamed the wrong passes.

## Root cause

`MemoryPlanningPass` runs **pre-specialization** and computes buffer lifetimes from the **original** step order. It legally donates a buffer `D` (dead at its original last-use) to a later matmul `M`'s output via `RebindStorageFrom` (which AddRefs `D`'s `TensorStorage` → refcount 2).

`BlasBatchPass` runs **later** and **hoists** all same-shape independent matmuls to the group's first position, running them back-to-back there. If `M` is hoisted ahead of `D`'s last consumer, `M`'s write lands in `D`'s storage **before** that consumer reads it → silent corruption.

Concrete repro (`Parity_SharedInput_QKV_Reshape_Permute_SDPA`): one shared input → Q/K/V projection matmuls → reshape → permute → SDPA. The V-projection buffer dies at its reshape, MemoryPlanning donates it to the K-projection matmul, and BlasBatch hoists that K matmul ahead of the V reshape — clobbering V.

## Fixes

1. **`BlasBatchPass` consecutive-run guard** — a batch group must be a CONTIGUOUS run of matmuls (`continue`→`break` in the candidate scan). Since the batched step runs every member at the first member's position and nothing non-member sits between them, no matmul is ever hoisted across an intervening step's buffer lifetime — a structural fix that holds regardless of how MemoryPlanning aliases buffers. (This consolidates the approach from #637, which is cleaner than this PR's earlier refcount-based `HasAliasedOutput` guard; verified sufficient against the full replay-aliasing suite.)
2. **Remove the `#1624` Softmax/SDPA gate** in `RunStepMutatingPasses` — memplan + reorder now run for all plans, restoring peak-memory reduction for attention/diffusion inference.
3. **Fix `AttentionFusionPass` (structural K recovery) and re-enable it** — the pass fed Kᵀ to FlashAttention as "K" and produced silently wrong results whenever `seqK == headDim` (a latent bug, since it ran via `RunCpuOptimizationPasses` even under the old gate). `TryRecoverNaturalKey` now traces the `TensorTranspose` that produced the QK matmul's second operand back to its input to feed the natural key, and **refuses to fuse** when the operand's layout cannot be proven (no producing transpose / not a clean last-two-dim swap) — so the silent-corruption case can no longer occur. The pass is forward-only (inference) so it needs no backward kernel. `EnableAttentionFusion` is on.

## Validation

- `MultiInputReplayAliasingTests` **16/16**
- `Engines.Compilation` **538/538** (net10.0) + **528/528** (net471), 6 pre-existing skips
- Optimization / BlasBatch / MemoryPlan / Reorder **150/150**
- Full project net10.0: 6181 passed; the 8 failures are pre-existing/environmental (GPU-kernel tests needing a real GPU, distributed TCP, GPU-pinned, allocation-churn) and all pass in isolation — none touch the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented matmul batching from hoisting operations when outputs share backing storage, reducing the risk of incorrect results.
  * Hardened attention fusion so FlashAttention is only applied after reliably recovering the expected (non-transposed) key tensor layout; fusion is refused when the layout can’t be proven.
* **Tests**
  * Added a comprehensive replay/determinism and compiled-vs-eager parity test suite covering many graph/aliasing and attention/fusion scenarios.
* **Documentation**
  * Updated the `EnableAttentionFusion` option notes to clarify its behavior and layout constraints.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledInferencePlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledInferencePlan.cs
@@ -1302,30 +1302,19 @@ internal sealed class CompiledInferencePlan<T> : ICompiledPlan<T>
     {
         if (steps.Count < 4) return steps;
 
-        // CORRECTNESS GATE (#1624): the step-mutating passes (OperatorReorderingPass + MemoryPlanningPass)
-        // produce a DETERMINISTIC-but-WRONG result for the attention fan-out topology — one shared input
-        // feeding q/k/v through matmul->reshape->permute, joined at scaled-dot-product attention (which
-        // GraphMode decomposes to permute/matmul/Softmax/matmul). The compiled output silently diverges
-        // from eager (masked for years because attention compile tests check shape/finiteness/determinism,
-        // never eager parity — see MultiInputReplayAliasingTests). Bisected: each pass corrupts it
-        // independently; bypassing BOTH makes the whole compile suite pass and matches eager. The root
-        // cause is subtle (a topologically-valid reorder still corrupts, so it is not a broken dependency)
-        // and the proper fix is deeper; until then, skip these passes whenever the plan contains a Softmax
-        // (the reliable signature of an attention block) so transformer/diffusion inference is CORRECT.
-        // Non-attention plans (CNN/MLP) keep the passes unchanged. Trade-off: attention inference plans lose
-        // peak-memory reduction — acceptable vs. silently-wrong attention, and revisited when the passes are
-        // made fan-out-correct. Also gate on a FUSED ScaledDotProductAttention step: a plan that keeps SDPA
-        // as one op (rather than the GraphMode-decomposed permute/matmul/Softmax/matmul) has no Softmax step,
-        // so the Softmax check alone would let the same fan-out topology slip through and miscompile.
-        for (int s = 0; s < steps.Count; s++)
-        {
-            var op = steps[s].OpType;
-            if (op == OpType.Softmax || op == OpType.LogSoftmax || op == OpType.ScaledDotProductAttention)
-                return steps;
-        }
-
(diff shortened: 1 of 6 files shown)


ooples:
Follow-up note — culprit localized further: bisecting `RunCpuOptimizationPasses` shows the corrupting pass is specifically **`AttentionFusionPass`**, not `OperatorReorderingPass`/`MemoryPlanningPass` (those are innocent). With only `AttentionFusionPass` enabled + the step-mutating passes on, the parity repro fails; with the step-mutating passes on but `RunCpuOptimizationPasses` off, it passes.

Mechanism: `AttentionFusionPass` runs in the post-specialization stage and reads Q/K/V buffers that `MemoryPlanningPass` already donated (based on the lifetimes of the pre-fusion attention matmuls, which the fusion then removes), so the fused FlashAttention step reads clobbered data.

I attempted the obvious proper fix — running `AttentionFusionPass` *first*, before the step-mutating passes — but it made things worse (broke `Parity_ScaledDotProductAttention_4D` and an inference-plan serialization test): the fused step has a hard post-specialization ordering constraint. So the real fix is in the `AttentionFusionPass` ↔ `MemoryPlanningPass` interaction (e.g. don't donate buffers a later fusion will read, or defensively copy on fuse), which is left as the tracked follow-up. The gate here is the validated interim.

Issue #478, opened 2026-05-27
Paper-scale training step allocates ~1 GB/step (GC churn): transient backward buffers bypass arena & ArrayPool>2^20


ooples:
## Summary

A paper-scale (~BERT-Large) training step allocates **~1 GB of transient GC garbage per step** and triggers a gen2 collection **every step**, even for a batch=1 / seq=4 input where the actual compute is trivial. The cost is **allocation + GC, not math** — the custom SIMD never gets to be the bottleneck. This makes paper-scale model training pathologically slow (~2.5 s/step here) and is the root cause of downstream AiDotNet model-family training-invariant tests timing out (especially on net471, which GCs slower).

This is a general engine perf bug affecting **all** paper-scale training, on **both** TFMs — not specific to one model.

## Reproduction / measurement

Probed with a single `MatryoshkaEmbedding<float>` (TransformerEmbeddingNetwork: EmbeddingLayer[30522×768] + 12× TransformerEncoderLayer, hidden 1536 / FFN 3072), batch=1, seq=4, repeated `Train(input, target)` inside an outer `TensorArena`:

```
iter 1: 2148 ms, alloc=985.8 MB, gc=(3/2/1)   <- gen2 GC every step
iter 2: 2671 ms, alloc=985.8 MB, gc=(6/6/1)
...stable ~985.8 MB/step, loss decreasing correctly (training is correct)
```

~985 MB/step ≈ one full copy of the parameter/gradient set, re-allocated and thrown away every step.

## Allocation breakdown (instrumented `new T[]` sites + GC counters)

- **~90 MB/step**: `CpuEngine.TensorEmbeddingLookupBackward` allocates the embedding gradient via `new Tensor<T>(...)` (stack: `BackwardFunctions.TensorEmbeddingLookupBackward <- CompiledDelegateChain.Execute <- GradientTape.ComputeGradientsViaGraphCore`).
- **~895 MB/step**: bypasses **both** `TensorBase(int[])` and `VectorBase`'s array constructor. The only remaining large-array sources are:
  1. `TensorArena.TryRentTensor` / `TryAllocateCore` allocating `new T[]` directly (lines that `new T[totalSize]` without going through `Vector<T>`'s ctor), i.e. the arena allocating **fresh every step instead of reusing** across steps; and/or
[...]

PR #634, opened 2026-06-17
fix(arena): byte-budget the cross-arena persistent pool to stop training-scale OOM (#1624)

ooples:
## Problem

The cross-arena persistent buffer pool (#478) retained at most **4 buffers per `(type, size)`** (`MaxPersistPerSize = 4`), on the assumption — stated in its own comment — that "a single-threaded training step rents each distinct size O(1) times."

A deep model trained with a **persistent gradient tape** (`GradientTapeOptions.Persistent = true`, the default) violates that assumption: the tape keeps **every** forward intermediate live for the whole backward pass, so the arena cannot reuse buffers *within* a step, and a single step rents the **same large size thousands of times**. With only 4 retained per size, the next step re-allocates thousands of multi-hundred-KB buffers from scratch — **every step**. Under a constrained heap (`DOTNET_GCHeapHardLimit`) that per-step churn outpaces gen2 collection and OOMs **even though the live set is flat** — the AiDotNet **#1624** training-scale failure.

### Dump evidence
A 14-layer ~110M-param model (SimCSE, tiny `[1,768]` input) grew the heap from ~1.7 GB to the 16 GB limit over a handful of steps. A captured dump showed **one live arena** holding **6,928 backing arrays (~6.46 GB) across only 8 distinct sizes** (e.g. `196608` = a `[256,768]` activation, repeated thousands of times).

## Fix

Bound the persistent pool by **total bytes** instead of a per-`(type,size)` count. A high-fan-out size now pools fully, so its buffers simply cycle between "in the pool" (between steps) and "rented by the current step" — steady-state footprint becomes the **single-step working set**, not working-set × churn.

- Byte ceiling defaults to **half of `GC.GetGCMemoryInfo().TotalAvailableMemoryBytes`** (reflects the heap hard limit when one is set, else physical RAM), overridable via `AIDOTNET_ARENA_PERSIST_BYTES` (`0` disables the pool).
- `net471` (no `GetGCMemoryInfo`) falls back to an 8 GiB default via `#if NET5_0_OR_GREATER`.
- New `_persistentBytes` thread-static counter kept in lock-step with the pool (inc on push, dec on pop).
- Recycled buffers are still zeroed — the zero-init contract is preserved.
- Budget is **per-thread** (the pool is `[ThreadStatic]`); documented on `MaxPersistTotalBytes`.

## Tests

- `HighFanOutLargeSize_PoolsEveryBufferAcrossLifetimes`: rents one large size **16 times per lifetime** and asserts all 16 are reused on the next lifetime. **Fails under the old cap-of-4, passes now.**
- `ByteBudget_BoundsPool_DropsSurplusBuffers`: sets the budget to exactly 4 large buffers, returns 16, and asserts the next lifetime reuses **exactly 4** — proving the budget actually *bounds* the pool (the surplus 12 are dropped for GC). Exercised through a `[ThreadStatic]` test seam (`MaxPersistTotalBytesOverrideForTests`) since the production budget is gigabytes.

Arena persistent-pool tests: **6/6** (net10.0), **5/5** (net471); CI green.

## Note

The end-to-end SimCSE 16 GB heap-bound repro is a consumer-side (AiDotNet) validation. The Tensors-side mechanism is now unit-proven in both directions: it pools the full high-fan-out working set **and** is bounded by the byte budget.

### DIFF
--- a/src/AiDotNet.Tensors/Helpers/TensorArena.cs
+++ b/src/AiDotNet.Tensors/Helpers/TensorArena.cs
@@ -90,11 +90,18 @@ public sealed class TensorArena : IDisposable
     // Thread-static: each thread keeps its own pool (no locks, matches the
     // [ThreadStatic] _current contract). Bounded: only arrays at/above
     // PersistThresholdElems are retained (small buffers GC cheaply and would
-    // bloat the dictionary), and at most MaxPersistPerSize per (type,size) so
-    // a model with many distinct shapes can't grow the pool without bound.
+    // bloat the dictionary), and the TOTAL retained bytes are capped by
+    // MaxPersistTotalBytes so the pool can't grow without bound.
     // ---------------------------------------------------------------------
     [ThreadStatic]
     private static Dictionary<(Type, int), Stack<Array>>? _persistent;
+
+    // Running total of bytes currently retained in this thread's persistent
+    // pool. Kept in lock-step with _persistent (incremented on push, decremented
+    // on pop) so ReturnPersistent can enforce the byte budget in O(1) without
+    // walking the dictionary.
+    [ThreadStatic]
+    private static long _persistentBytes;
 
     // Test/diagnostic: counts how many times a large buffer was served from the
     // cross-arena persistent pool (a reuse hit) rather than freshly allocated.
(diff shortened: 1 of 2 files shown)


ooples:
Closing: this targeted the WRONG part of TensorArena. The #1624 leak is the per-arena tensor ring (_ringBackingArrays) accumulating across steps under one un-Reset outer arena, NOT the cross-arena _persistent pool. The real fix is a per-step arena scope in the streaming training path (consumer fix/1624-streaming-per-step-arena), validated to fix the OOM on SimCSE @ Tensors 0.101.0.

Worse, this PR's byte budget ('half of GC available memory') reads PHYSICAL RAM under a DOTNET_GCHeapHardLimit rather than the heap limit, so the cap is effectively unbounded — tested with the per-step arena fix it OOM'd, climbing to 45 GB. Not merging.

ooples:
Closing: this byte-budget on the arena `_persistent` pool reads physical RAM via GC.GetGCMemoryInfo under a heap hard limit, so it re-introduces OOM, and it targets the wrong part of the arena. The real #1624 root cause was the **fused-optimizer training path opening no per-step arena** (cross-step accumulation in the caller's outer arena), fixed consumer-side in ooples/AiDotNet#1641 with a full A/B proof. No Tensors change is needed.

ooples:
Re-closing — after investigating, @ooples's diagnosis is correct and the byte-budget approach should not be revived. Independent verification:

1. **`TensorArena` is dormant in production.** There are zero `TensorArena.Create()` calls in Tensors `src/` or in the AiDotNet consumer (either the streaming or the model-bug branches). The cross-arena `_persistent` pool this PR byte-budgets is never on the #1624 path.
2. **The only arena users are Tensors' own perf tests/benchmarks, and they `Reset()` every iteration** (DeepProfileTests, GradientTapeBenchmarks) — correct usage, so the `_ringBackingArrays` growth the leak would require never happens there either.
3. **The real #1624 OOM cause is cumulative retention** (InferenceWeightCache pinning disposed weights + `GC.Collect` not compacting the LOH), already addressed in AiDotNet via `InferenceWeightCache.InvalidateAll()` + `GCLargeObjectHeapCompactionMode.CompactOnce` in the model-test teardown bases — not anything in the arena pool.
4. The default budget resolves at static type-init via `GC.GetGCMemoryInfo().TotalAvailableMemoryBytes`, which before a GC under the hard limit can report physical RAM — so "half of that" can exceed a configured `DOTNET_GCHeapHardLimit` and re-OOM, exactly as noted on close.

(The two exploratory commits pushed here while investigating are moot and left on the closed branch.) Keeping closed.

ooples:
Correction to my previous comment — I overstated point (1). After checking more branches: `NeuralNetworkBase` on `feature/agentic-phase2-multi-agent` DOES use a per-step `using var arena = TensorArena.Create();` (in both `TrainWithTape` and the grad-accumulation chunk loop). So the arena is NOT dormant on every consumer branch, and under that per-step create+dispose pattern the cross-arena `_persistent` pool IS on the path — this PR's *diagnosis* (cap-of-4 → per-step re-alloc churn for a high-fan-out single-step working set) is plausible there. It's the streaming and nn-oom branches that don't use the arena.

What remains solid and branch-independent is the **second** close objection: the default budget resolves at static type-init via `GC.GetGCMemoryInfo().TotalAvailableMemoryBytes`, which before a GC under a configured `DOTNET_GCHeapHardLimit` can report physical RAM — so "half of that" can exceed the heap limit and re-OOM. That flaw justifies closing this PR *as written* regardless of the diagnosis dispute.

If the churn is worth fixing for the per-step-arena branches, the correct shape would be a **lazily-resolved, heap-hard-limit-aware** budget (not static-init physical RAM), aligned with @ooples on the intended arena usage first. Leaving closed pending that.

PR #635, opened 2026-06-17
perf(compile): fusion-compatible liveness pooling for backward gradient buffers (standalone compiled-path optimization)

ooples:
## Context

Standalone memory optimization for the COMPILED training path (`CompiledTrainingPlan<T>`'s per-traced-tensor gradient buffers, ~417 MB on SimCSE's compiled path). Per @ooples this is **not** the #1624 fix (that's the streaming-path arena scope, AiDotNet#1638) — it's judged on its own merits.

`CompiledTrainingPlan` pre-allocates ONE gradient buffer per traced tensor and holds the entire backward intermediate-gradient working set resident at once. In the reverse walk, an intermediate gradient buffer is live only from the first backward step that writes it to the step that reads it; same-shape gradients with **disjoint live intervals** share one physical buffer (linear-scan register allocation), shrinking the resident set from "every traced tensor" to the **peak live frontier**.

## What changed since the first revision

The original revision was a prototype: it forced a 1:1 backward (disabling dataflow fusion + analytic-loss backward) so its re-zero schedule, indexed by backward *position*, aligned with action indices. That made it a memory-vs-speed trade that could never be a default.

**This revision makes pooling fusion-compatible** — it keeps the default-on optimizations:

- **Action-space liveness.** `BackwardGradientBufferPlanner.PlanOverActions` computes liveness over the FINAL backward **action** stream, treating each (possibly fused) action atomically. A buffer is therefore only shared across an action *boundary*, so the re-zero always lands where `Step()` can inject a clear — correct even when one fused action runs several forward steps' backward internally.
- **Two-detection wiring.** `BuildPooledGradMap` runs fusion + analytic *detection* against a throwaway shape-keyed gradMap to learn the decisions (graph-based ⇒ identical to the real pass), derives the action coverage via a shared enumerator that mirrors the real backward construction, plans buffers, and installs the aliased gradMap **before** the real fusion pass builds delegates against it.
- **Dataflow fusion + analytic-loss backward re-enabled under pooling.** Slice-prefix / dW-batching (env-opt-in) and backward pruning stay off (niche / index-shifting). A **drift guard** asserts the real decision sets + action count match the plan before applying the schedule — fail loud, never silently misalign.
- **First-class option.** Promoted the env-var gate to `TensorCodecOptions.EnableBackwardGradientPooling` (env still overrides). Opt-in (a memory/throughput trade-off, and it fails loud on a backward pattern the planner hasn't validated, so it stays opt-in rather than surprising every model), CPU-only (the GPU-graph backward is a separate execution model), default off.

## Tests

- `CompiledGradPoolParityTests`: pooled vs unpooled parameter gradients are bit-identical **with fusion enabled on both paths**, and the pooled physical-buffer count is strictly smaller — across the 4-layer base case, deeper/rectangular/very-deep shapes (`[Theory]`), and the config-option path.
- `BackwardGradientBufferPlannerTests`: the liveness invariants (no overlapping-lifetime tensors share a buffer; params dedicated; deep-chain reduction).
- **8/8** grad-pool tests + the full `Engines.Compilation` suite green on net10.0 (529) and net471 (519). Non-pooled path byte-for-byte unchanged.

## Scope (by design, not deferred)

This is the **CPU** compiled-path buffer pool, and it is **opt-in** — both are deliberate design decisions, not punted work:

- **CPU-only:** the GPU path executes the backward as a captured CUDA graph over device-resident buffers, a fundamentally different memory model than this host-array liveness pool — host buffer aliasing has no meaning there, so the option is ignored on a GPU engine rather than pretending to apply.
- **Opt-in (default off):** pooling is a memory-for-throughput trade (it adds a small per-step re-zero), and it fails loud on a backward pattern the planner hasn't validated. Both make per-workload opt-in the correct permanent design — same as the other `TensorCodecOptions` trades (mixed precision, spectral decomposition). It is complete and parity-validated as an opt-in option; there is no half-finished default-flip pending.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/BackwardGradientBufferPlanner.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/BackwardGradientBufferPlanner.cs
@@ -84,6 +84,90 @@ internal static class BackwardGradientBufferPlanner
     }
 
     /// <summary>
+    /// One backward ACTION's gradient data-dependency, expressed over tensor ids,
+    /// in the order the actions actually execute. Unlike <see cref="GradStep"/>
+    /// (one entry per forward step, reverse-walked positionally), this models the
+    /// FINAL backward action stream after fusion/analytic/skip transforms — where a
+    /// single fused action covers a contiguous range of forward steps and runs as
+    /// one atomic unit. Liveness is computed in ACTION-index units so a buffer is
+    /// only shared across an action BOUNDARY (never within a fused action, whose
+    /// internal grad hand-offs cannot be re-zeroed) and the re-zero schedule maps
+    /// directly onto backward action indices.
+    /// </summary>
+    internal readonly struct GradAction
+    {
+        /// <summary>Tensor ids whose gradient this action READS (the output grads of the steps it covers).</summary>
+        public readonly int[] ReadIds;
+        /// <summary>Tensor ids whose gradient this action WRITES/accumulates (the input grads of the steps it covers).</summary>
+        public readonly int[] WriteIds;
+
+        public GradAction(int[] readIds, int[] writeIds)
+        {
+            ReadIds = readIds;
+            WriteIds = writeIds;
+        }
+    }
+
+    /// <summary>
+    /// Liveness-based physical-buffer assignment computed over the FINAL backward
+    /// ACTION stream (fusion-compatible). Each action is treated atomically: a
+    /// tensor's gradient buffer is live from the first action that writes it to the
+    /// last action that touches (reads or writes) it, in action-index units. Two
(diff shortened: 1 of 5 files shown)


ooples:
Reframing: this is a standalone optimization for the COMPILED training path (CompiledTrainingPlan's per-traced-tensor gradient buffers, ~417 MB on SimCSE). It is NOT the #1624 fix — root-causing on Tensors 0.101.0 showed the OOM is the TensorArena ring accumulating across steps in the STREAMING path (SimCSE uses streaming, not the compiled path). Real fix: ooples/AiDotNet#1638 (per-step arena scope). Keeping #635 open only as an independent compiled-path memory optimization (parity-tested), to be judged on its own merits.

PR #636, opened 2026-06-17
perf(autodiff): release activations during streaming backward to bound peak (#1624)

ooples:
## Context

Component #2 of the #1624 activation-memory fix. The float-path heap dump showed the OOM is the deep model's **activation set** held through the backward; the failure throws **during** backward (`ComputeGradientsStreaming → FusedLinearBackwardCore → AutoTensorCache.RentOrAllocate`) — the backward's own buffer allocations pile on top of the resident forward activations and exceed the heap.

## The retention

`ComputeGradientsStreaming` already releases each node's **gradient** as the reverse walk consumes it, but kept every node's **activations** (`Output` / `SavedState` / backward closure) alive for the whole backward via the live node graph.

## Change

After each step's backward runs, null that node's `Output` / `SavedState` / `Backward` (and the `BackwardStep` copies). In reverse topological order a node's output is only consumed by steps that **already ran**, so this never drops a tensor a later step needs; node outputs are never sources (emitted + released separately), so no gradient the optimizer needs is lost. Combined with the consumer-side layer-cache skip (AiDotNet #1637), the live activation set is bounded to the **backward frontier** instead of the whole forward — freeing room for the backward's allocations.

On by default; `AIDOTNET_STREAMING_RELEASE_ACTIVATIONS=0` (or the internal `ReleaseStreamingActivations` flag) disables it.

## Correctness

- New `StreamingActivationRelease_OnVsOff_BitIdentical`: A/B the flag on a deeper chain, gradients **bit-identical** on vs off.
- Existing `Streaming_MatchesComputeGradients_BitIdentical` (streaming vs non-streaming reference) still passes with release on.
- 5 streaming tests green; net10.0 + net471 build clean. (2 pre-existing `ConvGroupNormGradCheck` flakes are in the **non-streaming** `ComputeGradients` path and fail with the release off too — unrelated.)

## Together with #1

#1 (layers stop pinning activations) + #2 (tape releases them mid-backward) is the pair that actually drops the peak; either alone leaves the activations pinned by the other. End-to-end SimCSE validation needs both landed + a Tensors release the consumer can consume.

## Summary by CodeRabbit

* **New Features**
  * Added optional streaming gradient activation release to reduce peak memory usage.

* **Improvements**
  * Improved reuse handling for persistent gradient tapes after streaming backward.
  * Strengthened gradient computation validation with earlier, clearer failures when required node data is missing.
  * Ensures streamed gradients remain bit-identical when the activation release option is toggled.

* **Tests**
  * Added serial streaming tests to prevent race conditions when toggling the activation release setting.
  * Added coverage verifying bit-identical results and that intermediate activations become collectible when release is enabled.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/GradNode.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/GradNode.cs
@@ -25,8 +25,10 @@ namespace AiDotNet.Tensors.Engines.Autodiff;
 /// </remarks>
 internal sealed class GradNode<T>
 {
-    /// <summary>The backward function that computes input gradients.</summary>
-    public BackwardFunction<T> Backward = null!;
+    /// <summary>The backward function that computes input gradients. Null after
+    /// the node has been released (pool return, or streaming-backward activation
+    /// release once its backward has run).</summary>
+    public BackwardFunction<T>? Backward;
 
     /// <summary>Input tensors to the operation (1-3 inline, overflow for 4+).</summary>
     public Tensor<T> Input0 = null!;
@@ -35,8 +37,10 @@ internal sealed class GradNode<T>
     public Tensor<T>[]? InputsOverflow;
     public byte InputCount;
 
-    /// <summary>The output tensor (for gradient seeding during backward).</summary>
-    public Tensor<T> Output = null!;
+    /// <summary>The output tensor (for gradient seeding during backward). Null
+    /// after the node has been released (pool return, or streaming-backward
+    /// activation release once its backward has run).</summary>
+    public Tensor<T>? Output;
 
     /// <summary>Optional saved state for backward (dropout mask, max indices, etc.).</summary>
     public object[]? SavedState;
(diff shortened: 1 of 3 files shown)


ooples:
Closing: superseded and made unnecessary by the real #1624 fix (per-step arena scope in TrainWithTapeStreaming — consumer fix/1624-streaming-per-step-arena), and this approach is unsafe.

Root cause was actually the TensorArena ring accumulating across steps (one outer arena, no Reset between steps, streaming path opened no per-step arena), not the tape holding activations. The arena owns the backing arrays, so releasing the tape's node graph doesn't free them.

This PR's node-graph nulling (node.Output/SavedState/Backward during ComputeGradientsStreaming) also CRASHES the real model with ArgumentNullException because the streaming backward runs under TryTrainWithFusedOptimizer where the persistent tape is being recorded by AutoTrainingCompiler — nulling the live graph corrupts it. It only passed toy-graph parity tests. Not reviving without a faithful persistent-tape/multi-consumer repro and a non-destructive design.

PR #637, opened 2026-06-18
fix(compile): BlasBatchPass must not reorder matmuls across buffer lifetimes

ooples:
## Problem

`BlasBatchPass` groups independent matmuls with matching K,N and executes the **whole group at the first member's position** (no-op'ing the originals). It runs **after** `MemoryPlanningPass` has aliased step buffers based on the *original* execution order, so moving a later matmul's write earlier can clobber a buffer the planner reused for a tensor still live across the skipped region — silently corrupting the result.

This was **latent while only single-input compiled plans existed**: their secondary leaves are baked constants, so `matmul(const, const)` is constant-folded away before this pass and never batched. A **multi-input** plan (the `GetOrCompileInference(Tensor<T>[] inputs, …)` path from #616 — e.g. a diffusion denoiser reading the noisy sample **and** a per-step timestep embedding, both mutable) keeps those matmuls dynamic, so they reach this pass and get reordered.

Surfaced by ooples/AiDotNet#1620: the DiT/UNet per-step denoising forward, routed through the multi-input compiled path, diverged from eager by ~2.2 (a 12-block graph reading the secondary input in every block diverged by ~6e5). The bug only triggers at 20+ steps, where the CPU optimization passes run — which is why the existing shallow #616 coverage missed it.

## Fix

Only batch a **consecutive run** of matmuls, so no member is ever moved across an intervening step's buffer lifetime. The genuine win (adjacent independent matmuls — attention q/k/v projections) is preserved; scattered matmuls stay in place.

## Tests

`MultiInputSecondaryConsumerReproTests` (new): secondary-through-matmul, multiple-consumers, broadcast-modulation, a **deep 12-block** graph (non-consecutive secondary matmuls — the failing repro), and a **consecutive-secondary** block graph (attention/adaLN pattern — proves consecutive batching itself is correct). Full `Engines.Compilation` suite: **527 passed, 0 failed, 6 skipped**.

End-to-end: with this fix the ooples/AiDotNet DiT multi-input compiled replay matches eager (verified against a local package build), so the verify-then-trust gate adopts the compiled plan instead of falling back to eager.

Pairs with the consumer wiring in ooples/AiDotNet (#1620 follow-up).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batching optimization to group only consecutive, compatible matmul steps and stop at the first non-joinable candidate, ensuring inputs are available at the batched execution point and preventing incorrect batching when dependencies would be violated.

* **Tests**
  * Added a new repro test suite covering multi-input compilation and replay across multiple scenarios (single/multiple consumers, broadcast modulation, deep graphs, and consecutive patterns), verifying results match an eager reference within `1e-3f`.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Optimization/BlasBatchPass.cs
+++ b/src/AiDotNet.Tensors/Engines/Optimization/BlasBatchPass.cs
@@ -56,17 +56,30 @@ internal sealed class BlasBatchPass : ICpuOptimizationPass
 
                 for (int j = i + 1; j < steps.Length; j++)
                 {
-                    if (consumed.Contains(j)) continue;
-                    if (steps[j].OpType != OpType.TensorMatMul || steps[j].Inputs.Length != 2) continue;
-                    if (steps[j].Inputs[0].Rank != 2 || steps[j].Inputs[1].Rank != 2) continue;
+                    // CONSECUTIVE-RUN ONLY: a batched group must be a CONTIGUOUS run of
+                    // matmuls. The batched step executes every member at the FIRST member's
+                    // position (and no-ops the originals). If a NON-group step sat between two
+                    // members, moving the later matmul before it would write that matmul's
+                    // output buffer early — and MemoryPlanningPass may have aliased that buffer
+                    // onto a tensor still live across the skipped region (buffers are reused by
+                    // original-order lifetime). Writing early then clobbers a live value →
+                    // silently wrong replay. This was latent while only single-input plans
+                    // existed: their secondary leaves are baked constants, so matmul(const,
+                    // const) is constant-folded away before this pass and never batched. A
+                    // multi-input plan (diffusion denoiser: noisy sample + per-step timestep
+                    // embedding) keeps those matmuls dynamic, so they reach this pass and get
+                    // reordered (#1620). Stop the run at the first step that can't join, so no
+                    // member is ever moved across an intervening step's buffer lifetime.
+                    if (consumed.Contains(j)) break;
+                    if (steps[j].OpType != OpType.TensorMatMul || steps[j].Inputs.Length != 2) break;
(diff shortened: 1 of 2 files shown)


ooples:
Consolidated into #632. After comparing the two BlasBatch fixes, this PR's **consecutive-run** approach (a batch group must be a contiguous matmul run, so no matmul is hoisted across an intervening step's buffer lifetime) is the cleaner, more robust fix — it holds regardless of how MemoryPlanning aliases buffers, unlike #632's earlier refcount-based `HasAliasedOutput` guard. So #632 now **adopts this consecutive-run approach** (replacing `HasAliasedOutput`) and **folds in this PR's `MultiInputSecondaryConsumerReproTests`**.

While consolidating I found those two **deep-graph repro tests flake on net471**: a 10–12 block unnormalized matmul chain compounds output magnitude to ~1e3–1e4 with random weights, so the compiled path's legitimate ~1-ULP *relative* GEMM rounding (net471 sums in a different SIMD order than eager; net10 is bit-exact) becomes a large *absolute* diff that trips the absolute `1e-3` replay tolerance. It's test non-determinism, not a code bug — deterministic relative error is 3e-5, identical across every optimization-pass config. #632 fixes it by seeding the deep tests with small deterministic weights (magnitude stays O(1)); tolerance unchanged. Both suites are now 24/24 on net10.0 **and** net471.

Recommend closing this PR in favor of #632 (which also carries the #1624 gate removal + the structural AttentionFusion fix). Flagging rather than closing your PR unilaterally.

ooples:
Superseded by #632, which merged (2026-06-18). #632 adopted this PR's consecutive-run BlasBatch approach verbatim (the superior, structural fix) and folded in this PR's `MultiInputSecondaryConsumerReproTests` — with the net471 deep-graph flakiness fixed by seeding the deep tests to bounded magnitude (tolerance unchanged). Closing as merged-via-#632.

PR #655, opened 2026-06-20
perf(conv): route high-channel convs to im2col+BLAS; fix batch=1 backward-kernel transB

ooples:
## Summary

Two CPU-engine conv routing fixes that make **foundation-scale diffusion models viable on CPU**. Surfaced while diagnosing the foundation-scale Diffusion ModelFamily CI shard timeouts (Kandinsky and friends time out at the 120 s `[Fact(Timeout)]` budget).

A per-block profile of one Kandinsky decoder (1.17 B params) forward showed **99% of the time was ResBlock convs** running at **0.2–1.2 GFLOP/s** — they were being mis-routed to the on-the-fly fused kernel / SIMD-direct 3×3 kernel instead of our packed `BlasManaged.Gemm`, which hits **350–440 GFLOP/s** at the same shapes (near hardware peak; oneDNN is intentionally stubbed per #131, so this is all our own machine-code GEMM).

### 1. Forward — `Conv2DWithIm2ColFloat`
High-channel convs (`inChannels >= 256`, non-Winograd) now go straight to explicit im2col + `BlasManaged.Gemm`. The `>= 256` gate is conservative — every measured high-channel conv wins decisively (>300 GFLOP/s) while smaller/stem/depthwise convs stay on the existing direct cascade (e.g. 128→128@28 SIMD-direct 98 GFLOP/s actually beats im2col+BLAS 41 there). Winograd keeps its large-spatial (output ≥ 224) niche.

### 2. Backward — `Conv2DBackwardKernel` (float)
The #573 transpose-avoiding K-concat path is now also used at `batch == 1`. batch=1 was computing `gradKernel = gradOutput · im2colᵀ` as a `transB=true` GEMM with thin `M = outChannels` and cache-hostile transposed-B access (~30–63 GFLOP/s). Building `gradOutput` transposed makes it a non-transposed GEMM with `M = colH` → ~300–400 GFLOP/s (up to 13.7×). No cross-batch stacking at batch=1, so buffers are the same size the fallback already rents.

## Measured (Kandinsky decoder 1.17B, CPU-only, this repo + consumer end-to-end)

| | before | after |
|---|---|---|
| `Engine.Conv2D` decoder convs | 0.2–1.2 GFLOP/s | **247–369 GFLOP/s** |
| `Conv2DBackwardKernel` (batch=1) | 30–63 GFLOP/s | **240–407 GFLOP/s** |
| hot forward (full threads) | 163 s | **2.9 s** (56×) |
| **Predict, 10 steps, 2-thread (CI-sim)** | ~1100 s | **68 s — under the 120 s timeout** |

End-to-end validated by packing locally and running the real consumer model.

## Correctness
Numerically identical (same conv math — im2col+BLAS vs the direct cascade). **46 conv correctness tests pass** (forward fast-path, backward input/kernel, ResNet accuracy) on both net10.0 and net471.

## Scope / follow-up
Fixes the **inference** path (all Predict/Clone/contract tests). A separate, still-open bottleneck remains for **training**: `Conv2DBackwardInput`'s `Col2ImAccumulate` scatter runs serially (~52 GFLOP/s even on 32 threads), so a full train iter is still ~160 s — tracked as a follow-up.

## Performance Improvements
- Improved CPU convolution by adding an extra fast routing path for high input channel counts when Winograd isn’t used.
- Accelerated backward-pass computation by channel-parallel im2col/col2im with deterministic-safe scheduling.
- Improved backward-pass GEMM buffering so batch==1 avoids unnecessary multi-batch handling.
- Improved backward-pass transpose with row-block parallelization.
- Accelerated attention backward’s inner float operations using SIMD vectorization for dot/axpy-style updates.

## Reliability / Efficiency
- Reduced overhead for very small workloads by forcing inline execution for single-iteration ranges, yielding more predictable exception behavior.


ooples:
@coderabbitai full review

ooples:
@coderabbitai review

PR #684, opened 2026-06-24
fix(cpu): make seeded TensorDropoutMask bit-reproducible (per-element SplitMix64)

ooples:
## Summary

`CpuEngine.TensorDropoutMask` produced a **non-reproducible mask even when given a seed**, above the 10000-element parallel threshold. This is the root cause of the intermittent *"two trainings at the same seed diverge"* model-family test flakiness in AiDotNet (e.g. `CNNBiLSTMCRF` / `BiLSTMCRF` `MoreData_ShouldNotDegrade`, which pass in isolation but are ~50% flaky in the full suite).

## Root cause

The seeded parallel branch:

```csharp
var seeds = new int[CpuParallelSettings.MaxDegreeOfParallelism];
for (int i = 0; i < seeds.Length; i++) seeds[i] = baseRandom.Next();
CpuParallelSettings.ParallelForOrSerial(0, totalElements, totalElements,
    () => RandomHelper.CreateSeededRandom(seeds[Thread.CurrentThread.ManagedThreadId % seeds.Length]),
    (i, state, localRandom) => { resultData[i] = localRandom.NextDouble() < rate ? zero : scale; return localRandom; },
    _ => { });
```

has **two** independent sources of nondeterminism:

1. **Seed selection** — `seeds[Thread.CurrentThread.ManagedThreadId % seeds.Length]`. `ManagedThreadId` depends on thread-pool state, which varies run-to-run, so element `i` is drawn from a different seed across runs.
2. **Draw order** — each thread pulls `localRandom.NextDouble()` *in the order it processes its dynamically-assigned chunk*. Element `i`'s value depends on how many elements its thread happened to process before `i`, i.e. on the partitioner's chunk boundaries — not on `i`.

So a *seeded* mask was not reproducible. Worse, the small path (`<= 10000`, one sequential `Random`) used a different scheme entirely, so it didn't match the large path for the same seed either.

`DropoutLayer` already derives a deterministic per-call seed (#1383) and passes it to `TensorDropoutMask` — but this path defeated that intent for any layer with >10000 activations.

## Fix

For the **seeded** case, element `i`'s keep/drop decision is now a **stateless SplitMix64 hash of `(seed, i)`**:

```csharp
ulong seedState = SplitMix64Seed(seed.Value);             // mix seed once
... resultData[i] = SplitMix64Uniform(seedState, i) < rate ? zero : scale;
```

The value depends **only on `(seed, i)`** — never on the processing thread, the partitioning, or the chunk size. Consequences:

- **Bit-reproducible** across runs, thread-pool states, **and machines with different core counts**.
- The small (serial) and large (parallel) paths now produce **identical** masks for the same seed.
- Strictly stronger than the chunk-id-seeded scheme in `TensorRandomUniformRangeInto` (which is only reproducible at a fixed worker count).
- The seed is run through the SplitMix64 finalizer first so adjacent seeds (e.g. consecutive per-call dropout seeds) yield uncorrelated streams.
- SplitMix64 reuses the constants already established in `SimdRandom.SplitMix64`.

The **unseeded** path is unchanged — intentionally nondeterministic via the shared thread-safe RNG.

## Tests

5 new reproducibility invariants in `MathInvariantExtendedTests` (the seeded parallel path was previously **untested** — all prior dropout tests used tensors below the 10000 threshold):

- `DropoutMask_Seeded_ReproducibleAcrossCalls_LargeParallel` — 4 repeats of a 20 000-element seeded mask are byte-identical.
- `DropoutMask_Seeded_ReproducibleAcrossCalls_Small`.
- `DropoutMask_Seeded_PathIndependent_SmallMatchesLargePrefix` — small mask equals the prefix of the large mask at the same seed. **This reliably fails on the old code** (old small/large paths used entirely different RNG schemes).
- `DropoutMask_Seeded_DifferentSeeds_DifferMasks`.
- `DropoutMask_Seeded_DropRateApproximatelyCorrect_LargeParallel`.

All 16 dropout tests pass. Library builds clean on **net10.0** and **net471** (the net8.0 path is identical code).

## Scope / follow-ups

- **CPU only.** The GPU `DropoutGpu` path is a separate backend-kernel RNG (different signature, `ulong seed`); CI is CPU-only, so exact CPU↔GPU bit-parity of seeded masks is left as a separate GPU-backend change (it would need the kernel to use the same per-element hash and cannot be CI-verified here).
- The seedless convenience overload `Dropout<T>(input, rate, training, out mask)` is intentionally non-reproducible (no seed parameter) and is unchanged.
- **Downstream:** once this is released, bump `AiDotNet`'s `Directory.Packages.props` `AiDotNet.Tensors` version to consume it; that closes the residual `MoreData_ShouldNotDegrade` full-suite flakiness for high-dropout models.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/CpuEngine.cs
+++ b/src/AiDotNet.Tensors/Engines/CpuEngine.cs
@@ -32969,7 +32969,6 @@ public partial class CpuEngine : ITensorLevelEngine
         if (shape == null) throw new ArgumentNullException(nameof(shape));
 
         var numOps = MathHelper.GetNumericOperations<T>();
-        var random = seed.HasValue ? RandomHelper.CreateSeededRandom(seed.Value) : RandomHelper.ThreadSafeRandom;
         var result = AutoTensorCache.RentOrAllocate<T>(shape);
         int totalElements = shape.Aggregate(1, (a, b) => a * b);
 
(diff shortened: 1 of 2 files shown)


ooples:
Note for reviewers: the `avx512-verify` job can intermittently fail `MachineKernelGemmTests.Gemm_MachineKernel_Fp32_Matches_PackBoth_And_Reference` — this is a **pre-existing, unrelated** flaky correctness bug in the hand-emitted AVX-512 FP32 GEMM kernel, now tracked as #685. It is independent of this PR (which touches only `TensorDropoutMask`): `avx512-verify` passed on this PR's exact base commit `a2638fb9` on `main`, and passed here on rerun of the same commit. `build` (full suite incl. the new dropout tests) is green.

Issue #653, opened 2026-06-20
perf: saturate cores across the whole forward (conv done ~44-47%; attention/norms/elementwise + near-serial whole-forward remain) — staged plan toward >=80-90% efficiency


ooples:
Tracking issue for the staged core-saturation effort. Extends #642 (BLAS worker pool parked ~99.9% during a single foundation forward) and follows #647 (conv parallelization).

## Where we are
- Probe `tools/ConvParallelProbe` measures **parallel efficiency = speedup / N** (maxdop sweep).
- Baseline (from #647): a batch=1 SD-UNet forward parallelized 1->16 at only **~1.1-1.2x (~7.5% efficiency)**; ~89% of serial wall-clock was Conv2D.
- After #647 conv work: dominant hi-res conv `[1,256,64,64]` **22% -> 47%**; deep `[1,256,8,8]` **~6% -> 44%**. So conv is partially fixed (~44-47%), **not** at the 90% target.
- The systemic problem persists and is **not conv-specific**: most ops (attention per-head QK^T/AV, GroupNorm/LayerNorm, activations, elementwise, and the thousands of small/thin GEMMs) run below the fan-out grain and execute serially on the caller; there is **no inter-op parallelism** across independent forward-DAG nodes.

## Root causes (code-confirmed)
1. `CooperativeGemmScheduler.Dispatch`: below `DefaultSerialGrainSize = 32K` an op runs fully inline on the caller; single-chunk / nested (`_inScheduler`) calls also collapse to serial. Per-dispatch coordination is paid per tiny op, thousands of times.
2. Zero inter-op parallelism: independent graph nodes (parallel residual branches, per-head projections) run sequentially. The pool's shared queue *can* blend chunks from multiple jobs, but nothing submits independent ops concurrently.

## Plan (each phase its own PR; correctness gate = bit-identical output vs serial)
- **Phase 0** — utilization meter (busy-thread-seconds / wall / cores) + `--attnblock` probe + end-to-end forward driver; capture per-stage baseline to set optimization order.
- **Phase 1** — dispatch floor: near-free coordination + auto-tuned per-op grain + cheap fan-out fast path; lift every small op off the serial floor.
- **Phase 2** — batch the small GEMMs (per-head QK^T/AV as one batched GEMM; CPU FlashAttention over batch*head) + norm/activation/elementwise fan-out.
[...]

PR #688, opened 2026-06-25
perf(#653): route StreamingStrategy GEMM dispatch onto low-latency PersistentParallelExecutor (~2x DiT inference)

ooples:
## Summary

Route the **Streaming GEMM strategy's** parallel dispatch off the park-and-wait `StreamingWorkerPool` and onto the low-latency `PersistentParallelExecutor` — the same migration #531 (ParallelForOrSerial) and #1478 (SDPA) did for the same "park/wakeup dominates rapid ops" reason.

## Why

Profiling one **ICEdit (SiT-XL / DiT-XL, 1152-dim × 28 layers)** `Predict` on CPU (`dotnet-trace`, `AIDOTNET_DISABLE_GPU=1`) — the workload behind the foundation-scale diffusion CI timeouts (AiDotNet #1672) — showed **~85% of wall time in worker park/wakeup + semaphore handshake and only ~5–7% in actual GEMM/attention compute**. The Streaming strategy fires thousands of medium GEMMs in rapid succession per denoising step; `StreamingWorkerPool` spins `IdleSpinBeforePark=1000` (~µs) then parks on a `ManualResetEventSlim`, so the inter-GEMM gap exceeds the spin budget and workers pay a wakeup on nearly every GEMM.

`PersistentParallelExecutor` is the low-latency cooperative pool that `PackBoth`/`SimdGemm` already use; moving Streaming onto it removes the per-GEMM park/wakeup floor.

## Change

The 4 `StreamingWorkerPool.Dispatch(procs, action)` sites in `StreamingStrategy` become `PersistentParallelExecutor.Instance.Execute(procs, action)`. **Same M-chunk partitioning, same wait-for-all barrier — only the executor changes.** `StreamingStrategy` was the pool's sole consumer, so this also retires the park-and-wait dispatch path.

## Validation

- **Perf:** one ICEdit `Predict` (DiT-XL, 32×32 latent, 16-thread CPU box): **>300 s → 155 s (~2×)**.
- **Correctness:** `865/901` GEMM / BLAS / MachineKernel / MathInvariant tests pass, **0 failures** (36 skipped). No GEMM result drift from the executor change — important because `StreamingWorkerPool`'s own comments warn that mishandling its dispatch produces nondeterministic GEMM drift; routing onto the already-concurrency-safe PPE avoids that surface entirely.

## Scope

This is **iteration 1** of the diffusion GEMM-efficiency work (AiDotNet #1672 foundation-scale CI timeouts / #653 core saturation). It banks the dispatch-pool win; ICEdit is still 155 s vs a 120 s test budget, so a follow-up will reduce the **per-forward dispatch count / join latency** (coarser GEMM chunking) — tracked separately. The machine-code microkernel itself is already healthy (~66 GF/s; A-prefetch is a documented −14% dead end per #656, C-prefetch neutral), so it is intentionally untouched here.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/BlasManaged/Strategies/StreamingStrategy.cs
+++ b/src/AiDotNet.Tensors/Engines/BlasManaged/Strategies/StreamingStrategy.cs
@@ -129,7 +129,7 @@ internal static class StreamingStrategy
                 int ldaLocal = lda, ldbLocal = ldb, ldcLocal = ldc;
                 bool taLocal = transA, tbLocal = transB;
 
-                StreamingWorkerPool.Dispatch(procsLocal, p =>
+                PersistentParallelExecutor.Instance.Execute(procsLocal, p =>
                 {
                     int mStart = (int)(((long)p * mLocal) / procsLocal);
                     int mEnd = (int)(((long)(p + 1) * mLocal) / procsLocal);
@@ -191,7 +191,7 @@ internal static class StreamingStrategy
                     bool taLocal = transA, tbLocal = transB;
                     int procsLocal = procs;
 
-                    StreamingWorkerPool.Dispatch(procsLocal, p =>
+                    PersistentParallelExecutor.Instance.Execute(procsLocal, p =>
                     {
                         var (kStart, kLen) = KAxisDriver.GetThreadRange(k, procsLocal, p);
                         if (kLen <= 0) return;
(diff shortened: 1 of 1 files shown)


ooples:
**Perf-claim correction (integrity note).** Follow-up A/B tournament work on a high-load-variance box revealed that ICEdit Predict wall-clock swings ~96–309 s across sessions, so the "~2× (>300→155 s)" figure in this PR's description was measured **cross-session, not clean back-to-back A/B** and is **not reliably confirmed**. The change itself stands on its merits — it's correctness-clean (865/901 GEMM/BLAS tests, 0 failures) and principled (low-latency PersistentParallelExecutor vs the park-and-wait pool, matching #531/#1478) — but the *magnitude* needs a clean same-session A/B (prebuilt base DLL vs modified, MIN-of-N) before the perf number should be trusted. Treat this as a low-risk correctness-neutral dispatch-path change pending that confirmation, not a verified 2×.

PR #707, opened 2026-06-28
perf(#1715): bound streaming-pool resident set on the GetParameters/register path (foundation-scale OOM)

ooples:
## Draft — MEASURED root cause (corrects the initial deferred-drop hypothesis)

Foundation-scale diffusion predictors (Flux2 ≈ **6.5 B params / ~31 GB fp32**) fail the `GetSetParameters_RoundTrips` contract tests under the 16 GB CI runner (AiDotNet #1715, shard `Unit-03d`: Flux2 / FluxDoubleStream / SiT).

### What I measured (the initial hypothesis was wrong)
With the AiDotNet-side wiring active (engage streaming on the param path + register materialized weights — branch `fix/generated-am-modelbugs`), I instrumented `WeightRegistry.GetStreamingReport()` during `GetParameterChunks` enumeration:

| chunk | residentMB | evictions | diskWriteMB | ws MB | cumElemsMB |
|------|-----------|-----------|-------------|-------|-----------|
| 250 | 8102 | 19 | 552 | 1623 | 8655 |
| 500 | 8066 | 215 | 9627 | 5850 | 17694 |
| 900 | 8174 | 519 | 23599 | 7180 | 31775 |

**The streaming pool already bounds the resident set at the 8 GB cap — eviction works.** The read enumeration does **not** OOM; it **times out** (~10 min for one full ~31 GB pass, disk-bound). So the deferred-drop / "pool doesn't evict" hypothesis is **disproven**.

### The actual residual
1. **Write pass**: the round-trip writes a ramp via `Tensor.Data.Span`, which rehydrates each chunk (`EnsureMaterialized` → resident) and marks it dirty. The real test OOM'd here — rehydrated/dirty `_data` isn't re-tracked/re-evicted by the pool the way the freshly-materialized read path is. **This is the Tensors fix**: the `Rehydrate`/`EnsureMaterialized` path must keep rehydrated weights under the pool's resident accounting so they can be re-evicted (with dirty write-back).
2. **Inherent speed**: even perfectly memory-bounded, a 31 GB model round-tripped to disk is ~10 min/pass × 3 passes ≈ 30 min → times out regardless. Once memory-safe, these foundation-scale contract tests warrant a longer timeout / `HeavyTimeout` (matches #1709) — the #1715 "must-fix footprint" is satisfied by streaming; the residual is inherent foundation-scale slowness.

### AiDotNet-side wiring (keep — bounds the read path)
- `NoisePredictorBase`/`MMDiTNoisePredictor` `GetParameterChunks`/`SetParameterChunks` → `MaybeEngageWeightStreaming()`.
- `LayerBase.EnsureParametersMaterialized` → `RegisterStreamingWeightsWithPool()` after `EnsureInitialized()`.

Full analysis in `docs/investigations/1715-streaming-register-path-oom.md` (updated). Tensors code change (rehydrate re-tracking) + the two-repo verify loop is the remaining work.

## Summary by CodeRabbit

* **Documentation**
  * Added a new investigation write-up detailing a streaming tensor OOM scenario, key findings, and recommended follow-up for resident-byte accounting and write-back behavior.
* **New Features**
  * Added writable, growable memory-mapped storage for streamed weights and enabled writable mmap aliasing for training/param-IO mode.
  * Exposed the resident-byte budget and added in-place entry data replacement with automatic eviction re-evaluation.
* **Bug Fixes**
  * Improved resident-owner bounding and write-back so mutated data remains correct across eviction/materialization, and writable aliases persist changes.
* **Tests**
  * Added coverage for writable vs read-only mmap behavior, mutation persistence, and bounded streaming materialization under pressure.

### DIFF
--- a/docs/investigations/1715-streaming-register-path-oom.md
+++ b/docs/investigations/1715-streaming-register-path-oom.md
@@ -39,7 +39,12 @@ so the spill mechanism exists — the gap is that the **owner tensors' resident 
 dropped** on this path, so paging the pool's byte[] snapshots to disk doesn't help (the unfreed
 `_data` is the accumulator).
 
-## Proposed fix (Tensors)
+## Rejected hypothesis — deferred-drop finalization (DISPROVEN; see "MEASURED CORRECTION" below)
+> ⚠️ Historical context only. The deferred-drop path below was the initial hypothesis; the MEASURED
+> CORRECTION section proved the pool's snapshot eviction already works and that this path does **not**
+> make the round-trip tests green. Do not implement it — the confirmed fix is the mmap redesign
+> (`MemoryMappedWeightStore` + writable zero-copy alias). Kept for the investigation trail.
+
 Bound the resident set on the register-heavy materialization path by finalizing deferred drops as we
 go, so each layer's `_data` is freed once its snapshot is in the pool:
 1. After `RegisterWeight` defers a drop, attempt finalization opportunistically when the pool crosses
@@ -87,12 +92,14 @@ out* (~10 min for one pass; disk-bound). So the pool's snapshot eviction works; 
 hypothesis is **disproven**.
 
 **Write pass** (`chunk.Data.Span[i] = …`, exactly as `AssertParameterChunksRoundTrip`):
+
 | chunk | residentMB | evictions | diskWriteMB |
-|------|-----------|-----------|-------------|
-| 50  | 1489 | 0 | 0 |
+| --- | ---: | ---: | ---: |
+| 50 | 1489 | 0 | 0 |
 | 100 | 3361 | 0 | 0 |
 | 150 | 5090 | 0 | 0 |
-| 200 | 6746 | **0** | **0** |
+| 200 | 6746 | 0 | 0 |
+
 → **OOM ~chunk 230**, before the 8 GB snapshot-eviction trigger ever fires.
 
 ### The real mechanism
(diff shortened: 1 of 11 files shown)


ooples:
## Consumer-side validation (AiDotNet)

The AiDotNet wiring that exercises this fix is now committed (AiDotNet PR #1719, commit `ba29b44f3`): `NoisePredictorBase`/`MMDiT`/`DiT` `GetParameterChunks`+`SetParameterChunks` engage **full-precision** weight streaming on the param-IO path, and `LayerBase.EnsureParametersMaterialized` registers just-materialized streaming weights with the pool.

**Verified:**
- The new `StreamingMaterializedOwnerBoundTests` unit test (this PR) passes: the materialized-owner set stays bounded **and** a mutation survives shed→rehydrate exactly (the write-back-before-drop path).
- 218 existing streaming/weight tests green (no regression).
- End-to-end against a local nupkg of this branch: a **foundation-scale MMDiT (SD3.5-Medium, ~9.6 GB fp32)** `GetParameterChunks` round-trip runs **memory-bounded (no OOM)** under `DOTNET_GCHeapHardLimit=0x400000000` (16 GB) — it streams to disk and times out on inherent IO rather than OOMing, which is exactly the intended OOM→bounded conversion. Pre-fix it OOM'd.

The four foundation round-trip tests are tagged `HeavyTimeout` (nightly lane) since they are now memory-safe but inherently slow (tens of GB across disk). Their correctness depends on this PR merging + an AiDotNet Tensors bump.

PR #729, opened 2026-07-06
Fix single-core deep-model training via cooperative-pool backward parallelism

### LINKED ISSUE #728: GPU/CPU under-parallelization: small tensor ops serialize training (1 CPU core, ~10% GPU)
## Summary
On a 128-core box with an RTX 3080, deep-model training pins to **1 CPU core** and only reaches **~10% GPU** utilization, because the training op-chain is a long sequence of *small* tensor ops executed one at a time with host↔device round-trips. Raw large GEMMs are fine, so this is about small-op scheduling, not GEMM itself.

## Measured (AiDotNet.Tensors 0.110.0, DirectGpuTensorEngine auto-selected)
- `engine.TensorMatMul` / `Matrix.Multiply` on **2048²**: ~22 of 128 cores (parallelizes) — but a naive loop is transfer-bound (re-uploads host inputs each call).
- N-BEATS training, **BatchSize=32**: **1.0 core**, 0% GPU (ops too small → CPU, single-threaded).
- N-BEATS training, **BatchSize=512**: **1.0 core**, **~10% GPU** (ops now dispatch to GPU but round-trip per op; GPU never fills).

## Root cause (two facets)
1. **CPU path:** small/medium GEMMs (e.g. 512×96·96×512) run single-threaded — oneDNN multi-thread size threshold appears too high; medium GEMMs should fan out.
2. **GPU path:** per-op host↔device transfers + no fusion/graph-capture across the tape serialize the step; `DirectGpuTensorEngine.MissingKernels.cs` itself notes host round-trips are the killer. Keeping the forward+backward resident + fusing/capturing the op-chain is needed for the GPU to fill.

## Repro
Minimal harness trains `NBEATSModel<float>` (hidden 512, 3×3 blocks) on a 20k-pt series; sample per-process CPU-seconds and `nvidia-smi`.

## Asks
- Lower the CPU GEMM multi-thread size threshold (or expose a thread-count knob) so medium GEMMs parallelize.
- Reduce per-op host transfers / fuse or CUDA-graph-capture the training op-chain so small ops don't each round-trip.

ooples:
Closes #728.

Autodiff backward row-parallelism used a raw Parallel.For (high ForkJoin dispatch cost) so small per-op gradients ran serial -> deep-model training pinned to 1/128 cores. `ForRows` now dispatches through `CpuParallelSettings.ParallelForOrSerial` (the cooperative StreamingWorkerPool primitive) and lowers `MinWorkForParallel` 64K->8K so small backward ops fan out.

Verified: N-BEATS<float> (hidden 512, 3x3 blocks, forced CpuEngine) training core usage 1.0 -> ~8 cores (7.4-9.9 across four 5s samples), at BOTH batch 32 and 512.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/BackwardParallel.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/BackwardParallel.cs
@@ -43,7 +43,7 @@ internal static class BackwardParallel
     /// dispatch beats serial execution. Below this, the
     /// <see cref="Parallel.For"/> setup cost dominates the body work.
     /// </summary>
-    internal const long MinWorkForParallel = 64L * 1024;
+    internal const long MinWorkForParallel = 8L * 1024;
 
     /// <summary>
     /// Max parallelism for backward operations. Capped lower than
@@ -95,8 +95,11 @@ internal static class BackwardParallel
         _inBackwardParallel = true;
         try
         {
-            var po = new ParallelOptions { MaxDegreeOfParallelism = MaxDegreeOfParallelism };
-            Parallel.For(0, rows, po, body);
+            // Route through the cooperative pool (StreamingWorkerPool-backed) instead of a raw Parallel.For:
+            // its per-dispatch cost is far lower, so small per-op backward work (the autodiff tape's bread and
+            // butter — e.g. N-BEATS FC-block gradients) actually fans out instead of the ForkJoin setup cost
+            // dominating and forcing single-threaded execution. The pool applies its own grain-size gate.
+            CpuParallelSettings.ParallelForOrSerial(0, rows, totalWork, body);
         }
         finally
         {


ooples:
### Honesty correction: core count up, wall-clock ~flat

Timed before/after (N-BEATS<float>, n=8000, hidden 256, 20 epochs, batch 64, forced CpuEngine):
- **before (0.110.0):** 284.0s
- **after (this PR):** 274.7s  (**~3% faster**)

So this change increases core UTILIZATION (1 -> ~8 cores) but does **not** meaningfully speed up deep-model training — the wall-clock bottleneck is elsewhere (serial forward pass / per-op tape overhead, all host-side; GPU sits at ~0-1% because these ops aren't GPU-resident). This PR is a legitimate parallelism correction (raw Parallel.For -> cooperative pool) and helps where the backward IS the bottleneck, but it is NOT the fix for slow deep-model training. The real levers are (a) GPU residency for the tape forward+backward and (b) reducing the serial host-side per-op overhead. Flagging so this isn't mistaken for a perf win.

PR #734, opened 2026-07-06
perf: zero-alloc training via TensorArena reuse + shape-safe ring (DRAFT)

ooples:
**Draft — WIP.** Core mechanism validated; full regression/invariant test suite being finalized.

## What
Activate the `TensorArena` buffer-reuse path during training (per-step reset on top-level `GradientTape.Dispose`), so per-op forward/backward scratch is recycled instead of GC-allocated — the PyTorch caching-allocator equivalent.

## The bug this also fixes
The arena's tensor ring buckets pooled buffers by **element count, not shape**. On the post-reset reuse path, `TryRentTensor` returned the cached wrapper with its **stale shape**, so a same-count/different-shape request (e.g. N-BEATS's `[B,L]`↔`[L,B]` permute) crashed `TensorAdd` (`[3,256]` vs `[256,3]`). Fix: `TensorBase.ArenaReshapeInPlace` — a zero-copy, **non-tape-recording** in-place reshape (the public `Reshape` records an autodiff node, unusable mid-forward) that also resets `_rowMajorStridesCache` — called on the reuse path when shapes differ.

## Evidence (N-BEATS)
- **Correctness:** forecast + MAE **bit-identical** to no-arena (pure allocation change).
- **Allocation:** 11,714.9 MB → **1,707.9 MB (85.4% cut)**, reproducible.

## Still to land before un-drafting
- Regression (arena-on ≡ arena-off across the model zoo) + invariant tests.
- Review of the `readonly`-removal on core `Tensor` shape fields.

Pairs with the AiDotNet-side arena activation (separate draft PR). Contributes to ooples/AiDotNet#1804.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/GradientCheckpointing.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/GradientCheckpointing.cs
@@ -86,7 +86,11 @@ public static class GradientCheckpointing<T>
                     // SymbolicBackwardGraphBuilder.Analyze when the cached pattern
                     // does not match the current tape's reachable entries).
                     using var recomputeTape = new GradientTape<T>(
-                        new GradientTapeOptions { Persistent = false });
+                        // SuppressArenaScope: this inner tape runs mid-backward, where the outer
+                        // tape has nulled the thread-current tape, so it would otherwise look like a
+                        // step-boundary tape and Reset() the OUTER tape's arena — corrupting the
+                        // accumulating gradients (#734). It borrows the arena for scratch only.
+                        new GradientTapeOptions { Persistent = false, SuppressArenaScope = true });
                     var reInput = inputs[0];
                     // Detach the segment input so the recompute graph is SELF-CONTAINED: its backward
                     // stops at the segment boundary instead of following reInput's producer into an
--- a/src/AiDotNet.Tensors/Engines/Autodiff/GradientTape.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/GradientTape.cs
@@ -92,6 +92,19 @@ public sealed class GradientTape<T> : IDisposable
     private bool _ownsArena;
     private Helpers.TensorArena? _ownedArena;
 
+    // Opt-in (default OFF; #734). When enabled, a TOP-LEVEL tape with no arena already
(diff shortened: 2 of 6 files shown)


ooples:
## Validation complete — arena is bit-equivalent across architectures ✅

Multi-model equivalence via the coreprobe harness: each model trained TWICE with a fixed seed — arena OFF (`TensorPool.ForceFreshAllocations=true`) vs arena ON — predictions compared element-by-element.

| Model (path) | max abs diff (ON vs OFF) | train alloc OFF→ON | Result |
|---|---|---|---|
| N-BEATS (TimeSeries tape) | **0.0** (bit-exact) | 107.58 → 47.62 MB (−55.7%) | PASS |
| MLP (eager NeuralNetworkBase) | **0.0** (bit-exact) | 44.01 → 42.26 MB (−4.0%) | PASS |
| Facade `Optimize()` (the new widening) | **0.0** (bit-exact) | 21.69 → 2.03 MB (−90.7%) | PASS |

Plus: Tensors reshape regression unit test **5/5 green** (net10.0 + net471); N-BEATS full-run bit-identical with **85.4%** allocation cut.

**Honest caveat (not inflating numbers):** the eager `NeuralNetwork.Train` path only drops 4% — its scratch is dominated by non-arena Matrix allocations, so the arena barely reduces its total there (correctness still holds bit-exact). The large wins are on the tape and `Optimize()` paths, which is exactly where the reuse engages.

Un-drafting for review.

ooples:
## Added: global tape-owned arena (universal coverage)

Extended the mechanism so a **top-level `GradientTape` creates and owns an arena when none is active** — making zero-alloc training universal, including custom/raw backprop loops that bypass the model bases and `Optimize()`. Paths that already open a long-lived arena are unaffected (tape sees `Current != null`, doesn't create its own).

**Validated (coreprobe):**
- Bit-exact arena-on vs arena-off across N-BEATS / MLP / facade `Optimize()` — max abs diff **0.0**.
- Engages on a raw non-base training loop: cross-arena reuse hits **0 → 20**, final weights bit-exact.
- Safe for training: grads consumed in-scope, weights/moments are plain-heap; inference never takes a tape.

Honest caveat (inherent to any arena reuse, not new): an arena-backed op output deliberately retained past the tape scope and read after a later step will be recycled — no training path does this.

PR #760, opened 2026-07-10
fix(compile): zero grad buffers when a specialized backward accumulates (multi-consumer)

ooples:
## Problem

The compiled training plan skips **all** gradient-buffer zeroing when every backward step is specialized (`genericBackwardCount == 0`), on the assumption that a specialized delegate always **overwrites** its gradient buffer (beta=0).

That assumption is false for a **multi-consumer** tensor. A tensor consumed by N ops receives the **sum** of N gradient contributions, so specialized delegates that write such a buffer **accumulate in place** (`TensorAddInto` / `+=`) — the `BroadcastAdd` bias-grad path is the clearest case (its `accumA`/`accumB` branches). When that buffer is never zeroed, each step's accumulation lands on top of the **prior** step's gradient, so the gradient — and the Adam update — grows without bound while the loss barely moves.

This is exactly the failure the AiDotNet time-series **N-BEATS** resident path documented: its doubly-residual stack makes the residual tensors inherently multi-consumer, the graph compiles to an all-specialized backward, and the un-zeroed accumulators blew the weights up to ~1e13 (so the model validation-rejected the resident run and fell back to eager every time).

## Fix

Only take the skip-all-zeroing fast path when **no tensor has multiple consumers** (no accumulation can occur). Otherwise fall back to the existing clear-all path that zeroes every gradient buffer at step start. Feed-forward graphs keep the optimization; multi-consumer graphs are now correct.

## Verification

New regression test `MultiConsumerGradZeroingTests` (runs on `CpuEngine`, no GPU needed): a multi-consumer bias broadcast-added to two inputs has analytic gradient `2*batch = 8` on **every** step.

- **Before the fix:** step 1 → 8, step 2 → **16** (8 accumulated onto the un-zeroed 8) — the exact "grows every step" blow-up.
- **After the fix:** stays 8. ✅

Confirmed fail-before / pass-after on `net10.0` and `net471`.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -4478,11 +4478,23 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
             }
         }
 
-        // If all backward steps are specialized (overwrite), we can skip gradient zeroing entirely.
-        // Under pooling, force the clear-all path (null) so every physical buffer is
-        // zeroed at step start (covering each shared buffer's FIRST tenant); the
-        // re-zero schedule clears it again before each subsequent tenant.
-        int[]? genericGradIndices = (!useGradPool && genericBackwardCount == 0) ? new int[0] : null;
+        // If all backward steps are specialized we CAN skip gradient zeroing — but ONLY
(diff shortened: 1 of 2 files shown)


ooples:
Consolidated into #759 (per maintainer direction). #759 now carries the same fix — made **precise** (zeroes only the accumulating multi-consumer buffers, so no perf cost on correct paths) — plus this regression test AND the original single-step parity tests, and fixes the net471 build error. Superseded here to keep one PR.

PR #767, opened 2026-07-10
perf(arena): persistent-pool deep-model buffers + fused-optimizer Adam moments

ooples:
Two related changes to the cross-arena **persistent pool** — the tier that lets a fresh-`TensorArena`-per-training-step trainer reuse large backing arrays instead of re-allocating them each step (#478).

## 1. Raise the per-size retention cap (4 → 64)

The pool capped retention at **4 buffers per (type, size)**, assuming "a training step rents each distinct size O(1) times". True for shallow nets, **wrong for deep transformers**: an N-layer stack produces N same-shaped hidden-state / gradient tensors per step. At cap 4, a fresh-arena-per-step trainer dropped N−4 of them on `Dispose` and re-allocated them next step — PerfView measured **~2 GB/step of `Double[]` churn** on an 18-layer VALL-E-X-clone (`TensorArena.TryRentTensor`).

Raise to 64. Retention is **self-limiting** — a `Dispose` only returns as many same-size buffers as the step actually used — so shallow models are unchanged; only deep models retain more, bounded by their own working set.

## 2. Pool fused-optimizer Adam moments through the persistent tier

`ConfigureOptimizerDouble/Float` allocated the Adam `m`/`v`/`vMax` buffers as raw `new double[len]` per param. `ConfigureOptimizer` re-runs every time the compiled plan is invalidated (repeatedly during warmup as lazy layers materialize), re-allocating the **entire** moment set each time — PerfView attributed **~2 GB** to `ConfigureOptimizerDouble` on the VALL-E-X-clone warmup.

The moments are long-lived (must survive the transient per-step arena — the ring would recycle them mid-step, and `RentPinned` would re-allocate every step), so the persistent tier is the right fit. Expose `TensorArena.RentPersistentZeroed<T>` / `ReturnPersistentBuffer<T>` (zeroed on rent, re-zeroed on reuse — preserving the moments' `new[]` zero-init) and route the m/v/vMax allocations through it; a prior config's moments return to the pool on each plan-invalidate / re-configure. GPU-init temporaries and bf16/int8 moment buffers keep their own lifecycles.

## Tests

- `ManySameSizeBuffers_PerStep_AllReusedAcrossLifetimes` — 16 same-size buffers per cycle all reused next cycle (fails at cap 4).
- `RentPersistentZeroed_ZeroesPoolsAndRezeroesOnReuse` — the new API's zero/pool/re-zero contract.
- Full fused-optimizer + convergence suites (108 tests) stay green, proving the pooled moments are never recycled mid-step.

## Summary by CodeRabbit

* **Performance**
  * Improved memory reuse for training optimizer buffers, reducing repeated large allocations during optimizer reconfiguration and plan disposal.
  * Increased persistent buffer retention to better support deep models and repeated allocation patterns.
  * Ensured reused buffers are reset to zero before being provided again.

* **Tests**
  * Added coverage for persistent buffer reuse, zero-initialization, and deep allocation scenarios.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -218,6 +218,7 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
         foreach (var buf in _gpuOptimizerBuffers)
             buf.Dispose();
         _gpuOptimizerBuffers.Clear();
+        ReturnPooledMoments(_optimizerRuntimeState);
         _optimizerRuntimeState = null;
 
         if (_residentGradBuffers is not null)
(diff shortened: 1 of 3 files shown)


ooples:

PR #768, opened 2026-07-10
perf(simd): single-pass AdamW + hoist step-global bias-correction Math.Pow

ooples:
Two bit-identical optimizations to the fused CPU Adam/AdamW SIMD kernels, both surfaced by PerfView on a large-model (VALL-E-X-clone) training step.

## 1. Fuse AdamW into a single pass over the parameter array

Adam's fused kernel is one pass (load `param/grad/m/v`, update, store). `AdamWUpdateSimd` did a **separate** full pass over `param` (read+write every element) for the decoupled weight decay `param *= 1-wd·lr`, then called `AdamUpdateSimd` which read+wrote `param` **again** — two memory-bound passes where Adam does one. Fold the decay into the Adam loop (in-register) so `param` is touched once. **Bit-identical** — the scaled param is a float/double, so keeping it in a register equals the value the old pre-scale pass stored and reloaded.

## 2. Hoist the step-global bias-correction `Math.Pow` out of the per-parameter loop

The kernels computed `bc1 = 1-β1^step`, `bc2 = 1-β2^step` at the top of **every call**, but they're invoked **once per parameter tensor**, and the corrections are step-global. A deep model with many small tensors paid ~2×(param count) `Math.Pow` per step for a value that should be computed twice. Split each kernel into a step-taking wrapper + a `bc1/bc2`-taking core, and hoist the two `Math.Pow` to **once per step** in all four fused-optimizer closures (float/double × grouped/non-grouped). **Bit-identical** — same corrections, computed once instead of per-param.

## Tests

- `FusedAdamW_{Double,Float}_BitIdenticalToTwoPass` — 28 raw-bit cases, fused vs the reconstructed two-pass reference.
- `AdamAndAdamW_BcOverload_BitIdenticalToStepOverload_Double` — step ↔ bc overload bit-equivalence across SIMD-tail lengths/steps.
- Full fused-optimizer + `OptimizerConvergence` suites (145 tests) stay green, exercising the hoisted closures over real multi-step Adam/AdamW training.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -2188,6 +2188,8 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
         _optimizerUpdate = () =>
         {
             _optimizerStep++;
+            float stepBc1 = 1f - MathF.Pow(b1, _optimizerStep);
+            float stepBc2 = 1f - MathF.Pow(b2, _optimizerStep);
             // #739 review: retire any captured on-device step graph before the CPU fused optimizer
             // mutates host weights and invalidates their resident device buffers (MarkHostWeightMutated
             // below) — otherwise a later LaunchCapturedGraph would replay against freed/stale device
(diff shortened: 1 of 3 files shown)


ooples:

PR #769, opened 2026-07-10
perf(simd): multi-tensor (foreach) fused AdamW optimizer

ooples:
First phase of the fused multi-tensor optimizer (PyTorch `foreach` / `fused` style). Bulk-processes the parameter set to kill per-tensor overhead.

## This PR (Phase 1 — CPU foreach, double AdamW)

The CPU fused optimizer rebuilt all eight AVX constant broadcasts (β1, 1-β1, β2, 1-β2, -lrAdj, eps, 1/bc2, wd-scale) **plus a method call** once **per parameter tensor** — redundant work a deep model with hundreds of small tensors pays every step.

- `AdamWUpdateSimdMulti` (double + float): builds the step-global SIMD constants **once**, loops the tensor list internally. **Bit-identical** to the per-parameter `bc1/bc2` overload.
- Wired the non-grouped double-AdamW fused closure to it (its pure CPU fp32-moment path has no GPU-resident / bf16 / int8 branch to preserve).

**Tests:** `AdamWMultiTensor_BitIdenticalToPerParam_Double` (multi vs per-param, raw bit pattern, SIMD-tail + empty tensors); fused-optimizer + `OptimizerConvergence` suites (139) green.

## Follow-on phases (scoped, not in this PR)

- **1c — float + grouped closures.** The float closure carries GPU-resident + bf16/int8 branches; wiring the multi-tensor path there needs per-case guards.
- **2 — contiguous flatten (amortize per-tensor SIMD scalar tails).** The full `fused=True` win needs params/grads laid out contiguously; the fused path currently binds per-tensor (bypassing the eager `ParameterBuffer`), so this is a consumer + Tensors change.
- **3 — GPU `multi_tensor_apply`.** The high-value case (each per-tensor op is a kernel launch). Needs a CUDA box to verify (this dev box is AMD/OpenCL).

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -3324,6 +3324,19 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
             double stepBc1 = 1.0 - System.Math.Pow(b1, _optimizerStep);
             double stepBc2 = 1.0 - System.Math.Pow(b2, _optimizerStep);
             double lr = lrSchedule.GetLr(_optimizerStep);
+            // Multi-tensor (PyTorch foreach) fast path: update EVERY parameter in one
+            // fused-AdamW call — the SIMD constant vectors are built once instead of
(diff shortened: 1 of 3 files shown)


ooples:

PR #799, opened 2026-07-17
feat(opencl): paged-attention decode kernel (P1) — GPU-validated

ooples:
First slice of **P1 (vLLM-core paged attention)**: an OpenCL decode kernel computing single-query attention that reads K/V from a physical block pool (`[maxBlocks, blockSize, heads, headDim]`) via a block table — so the KV cache need not be contiguous (the mechanism that lets vLLM pack many sequences without fragmentation).

- `PagedAttentionKernels.paged_attention_decode`: online-softmax (FlashAttention-style) single pass, one work-item per head, headDim ≤ 256. Compiled with SafeMathFlags (relies on INFINITY/exp). Block table bound as an int buffer.
- `OpenClBackend.PagedAttentionDecode` dispatch.
- `PagedAttentionOpenClTests` validates vs a standard-attention CPU oracle (K/V gathered through the block table).

**Validated on AMD gfx90c: 3/3** (varied heads/headDim/blockSize/seqLen), `AIDOTNET_REQUIRE_GPU_TESTS=1`. Contract matches the CPU `PagedKVCache` block layout.

Follow-ups: prefill (multi-query), GQA, on-device block allocator, and the other 5 backends.


ooples:
Superseded by #801 (consolidated single PR for automated review). Closing.

PR #810, opened 2026-07-18
feat(autodiff): public ComputeGradientsFromSeed for manual-backward layer wrapping

ooples:
## What

Adds `GradientTape<T>.ComputeGradientsFromSeed(output, outputGradient, sources)` — a public entry point to run the reverse pass seeded by a supplied upstream gradient at `output`, instead of the default "ones at the loss" seed.

## Why

The tape already supported seeding the backward with an arbitrary gradient (the `internal` `seedOverride` path used by the mixed-precision bridge), but it was not reachable publicly. That capability is exactly what a classic **manual-backward layer** needs to wrap a tape-recorded forward:

- `Forward`: run the op(s) with a tape active (they auto-record).
- `Backward(outputGradient)`: call `ComputeGradientsFromSeed(output, outputGradient, [params, input])` to obtain the parameter **and** input gradients in a single tape walk — no hand-written adjoint code.

This lets a tape-native operator be packaged as an `ILayer`/`LayerBase` (`Backward(outputGradient)` contract) without re-implementing autodiff by hand.

## How

Thin, validated wrapper over the existing internal seeded backward. Null + shape guards; no change to any existing path (the default `ComputeGradients` is untouched).

## Tests

`GradientTapeSeededBackwardTests` (4): correct seeded chain-rule for `y = w⊙x`; equivalence with the default `sum(y)` backward when seeded with ones; shape-mismatch and null-arg guards. Green on **net10.0** and **net471**.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/GradientTape.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/GradientTape.cs
@@ -489,6 +489,39 @@ public sealed class GradientTape<T> : IDisposable
         => ComputeGradients(loss, sources, createGraph, seedOverride: null);
 
     /// <summary>
+    /// Reverse pass seeded by a supplied upstream gradient at <paramref name="output"/>, instead of the
+    /// default "ones at the loss" seed. This is the entry point for wrapping a tape-recorded forward as a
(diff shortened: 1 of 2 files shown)


ooples:
Closing — unnecessary. I have full source access to this repo, so a manual-backward layer can call the existing internal seeded-backward directly (or via InternalsVisibleTo); no public API addition is needed for that. Reverting to keep the surface area minimal.

Issue #775, opened 2026-07-11
CPU vs GPU training divergence: BCE-with-logits amplifies a CPU/GPU forward numerical difference in the ViT path


ooples:
## Summary

Training on the **CPU engine (`CpuEngine`)** diverges while the **GPU engine (`DirectGpuTensorEngine`)** converges, for a plain-ViT model (`PatchEmbeddingLayer` → transformer/dense) optimized with **`BinaryCrossEntropyWithLogitsLoss`**. The loss climbs and the logits explode on CPU; the identical model, init, and optimizer converge on GPU.

Surfaced while making AiDotNet's `EfficientTAM` paper-faithful (plain ViT encoder). Worked around downstream by switching that model's mask loss to MSE, but the underlying CPU/GPU numerical divergence is a real Tensors-level concern that can bite any ViT + logit-loss model on CPU.

## Repro (minimal)

Model = `PatchEmbeddingLayer(patchSize=16, embed=64)` → `DenseLayer(64, ReLU)` → `DenseLayer(1)`, input `[4,3,32,32]`, random target in `[0,1)`, AdamW default LR 1e-3, 100 memorization steps on a fixed (input, target). **No dropout** (so train-mode == eval-mode).

| Loss | GPU (DirectGpu) | CPU |
|------|-----------------|-----|
| **BCEWithLogits** | converges 0.84 → **0.56** (logit ~3) | **DIVERGES** 0.84 → **12.4** (logit **24.7**) |
| **MSE** | converges 0.58 → 0.0007 | converges 0.58 → 0.003 (≈ identical to GPU) |

Lowering the LR to 1e-4 only *slows* the CPU divergence (0.84 → 4.1), confirming it is systematic, not simple overshoot.

## Diagnosis

- **Not a backward bug.** Finite-difference of the BCE loss vs the first Adam step direction agrees **93%** on CPU (the 4/61 disagreements are small-gradient noise). `TensorPermute` / `Reshape` / conv backward all gradient-check clean on CPU (err ~2e-10).
- Each engine's backward is **self-consistent with its own forward**, but the **CPU and GPU forwards differ numerically** in the ViT path (candidates: patch-embed projection, attention softmax, `LayerNorm`). `DifferentInputs` on GPU shows a suspiciously small 2.9e-6 sensitivity.
[...]

PR #829, opened 2026-07-21
fix(gpu): add tape-gradient parity and correct backward kernels

ooples:
## What this does

Adds a **tape-gradient parity phase** over the whole op registry, and fixes the defects it found.

### Why it was needed

`OpParityHarness.CheckBackward` is **dead code**: no test method invokes it, and **zero of the 475 registry cases** pass the `runFloatGrad` delegate it requires, so it would skip everything even if wired. The practical consequence is that **GPU backward kernels have never had CPU-vs-GPU coverage**. This phase is the first thing that compares them, and it needs no per-case delegate — it discovers tape leaves and pairs them by content.

### Defects fixed

| Fix | Impact |
|---|---|
| **CUDA `InstanceNormBackward`** launched one kernel with mismatched args — wrote `gradInput` into the **`gradBeta` buffer**, read uninitialised buffers as its input sums, reinterpreted `float epsilon` as `int W`, and never launched the required sums kernel | `d(input)`/`d(gamma)` were **all zeros** — InstanceNorm learned nothing and blocked gradient flow upstream |
| **Autodiff use-after-free** — gradients built via `gradOutput.Clone()` sat on pool-backed COW storage recycled while `grads` still referenced them (5 backwards) | gradients silently zeroed; this is the **true root cause** of the original `Scatter` defect |
| **CPU CIoU had no aspect-ratio gradient** — `v` materialised as a constant leaf | CIoU was training as **DIoU** |
| **IoU-family tie-breaking** — fused backwards used strict `>`/`<` where `MaxBackward` gives the full gradient to `inputs[0]` at ties | wrong gradients wherever box corners share coordinates |
| **SELU `>=`** finished on CUDA/HIP/Metal (`#775` had only landed on OpenCL/WebGpu) | factor-of-`alpha` error at `0` and **`-0.0`** |
| **4 backward crashes** — `Threshold` boxing, `Norm` axis, `ScatterAdd` arity, `BatchOuterProduct` rank | `Threshold` crashed **every float model** on its first backward |

### Results

- tape-gradient parity: **13 failed → 4 failed**, 547 passed, 6 skipped
- forward parity: **22 failed / 1056 passed — unchanged, no regressions**
- `Scatter` passes with its baseline entry **removed**

### Known-not-fixed (4)

`TensorSELU`, `TensorNormalize`, `TensorSquash`, `FusedLinearCrossEntropyWithLogits`. All stable across three consecutive runs (real, not flaky). For SELU the forward is proven identical with and without a tape on both engines, and its input-leaf gradient agrees to `2.98E-08`, so the reported `3.523E+00` is not yet accounted for by any measured leaf.

### Caveats

- `InstanceNormBackward` is verified against the CPU reference on **CUDA only**; HIP/Metal/Vulkan/WebGpu could not be exercised on this machine and may share the pattern.
- The IoU `predArea` ReLU-indicator change is a **verified no-op** on current parity data (every parity box has `px2 > px1`) — a correctness improvement this suite does not yet cover, flagged rather than claimed.
- One commit (`3885c00`, `power_scalar` NaN) is **cherry-picked from `codex/pr-827-all-backends`**, authored there. Expect it to dedupe at merge.
- The suite has residual nondeterminism unrelated to these ops (Tanh/Mish); one run reported 5 failures while naming 4.

## Summary by CodeRabbit

* **New Features**
  * Added complex tensor interleaving/deinterleaving operations across GPU backends.
  * Improved batched GPU RFFT/IRFFT with correct tape-based autodiff.
* **Bug Fixes**
  * Fixed gradient correctness for scatter/indexing, axis-aware norms, IoU-family losses, and multi-input additions.
  * Corrected GPU behavior for InstanceNorm, Threshold, SELU at `x == 0` (including `-0`), negative-base power, and loss/equals semantics.
  * Restored CIoU aspect-ratio gradient flow.
* **Performance**
  * Refined CPU FFT execution for efficiency.
* **Tests**
  * Expanded GPU/CPU parity, tape-gradient parity, boundary, and GPU fallback diagnostics coverage.


ooples:
### Correction to `3aed217`'s commit message

`3aed217` reports **"Conv : 4 failed / 417 passed -> unchanged"**. That number is not meaningful and should not be relied on in review.

It came from `--filter "FullyQualifiedName~Conv"`, which matches `CliConvertTests`, `TrainingConvergence` and `HrrBindBatch_MatchesCircularConvolutionReference` — a substring, not a test selector. The set it measured was mostly CLI serialization and licensing tests.

**The real figure, class-scoped:** `Conv2D` — **0 failed / 189 passed**, before and after that commit. Convolution was unaffected by the `im2col`/`col2im` layout change, which is what the line was trying to establish.

Leaving the commit message as-is rather than rewriting pushed history; recording the correction here.

Two related notes for whoever reviews this branch:

- The same substring trap produced a phantom "12 trig failures" later on (`~Sin` matches `Licen`**sin**`g`, `Mis`**sin**`g`). Those 12 are 6 Licensing + 5 Safetensors + 1 Gguf, all environment-driven: they assert no-key trial behaviour and fail when `AIDOTNET_LICENSE_KEY` is set. With it unset, `PersistenceGuardTests` passes 16/16 and `CliConvertTests` 6/6.
- `b2a4337` corrects `5a4bda3`, which I had reported as fixed and was not — it replaced an uninitialised-memory bug with a wrong-values bug in the mean-pool backward path. Both are described in those commit messages.

ooples:
@coderabbitai review

PR #831, opened 2026-07-21
perf(fft): wire NativeComplexFFT butterfly to SIMD radix-2 (test-gated)

ooples:
## What
Routes the tensor-level `NativeComplexFFT` per-row transform (shared by `NativeComplexFFTComplex`/`FFTND`/`NativeSpectralFilter` and the GPU CPU-fallback) through the existing SIMD radix-2 kernel (`FftKernels.IterativeRadix2NoCache`) for power-of-2 lengths, instead of the hand-rolled scalar butterfly. The fast SIMD + tested `Fft.*` module already existed but was disjoint from the `Native*FFT` engine ops — this bridges them.

- `CpuEngine.NativeFFTInPlaceDoubleSpan` / `NativeFFTInPlaceFloatSpan`: delegate pow2 transforms to the SIMD stage loop under `#if NET7_0_OR_GREATER`; `[ThreadStatic]` scratch so the batch loop allocates at most once per thread. Non-pow2, net471, and generic-`T` keep the exact scalar path.
- Contract preserved exactly: `IterativeRadix2NoCache` is raw/unnormalized, matching the scalar butterfly (callers apply 1/n) — identical layout + normalization.

## Correctness (test-gated)
- Multi-TFM build: **net10.0 + net8.0 + net471 all pass, 0 warnings**.
- FFT suite (`~Fft`): **net471 261/0, net10.0 271/0**. `FftApiTests`+`FftGradcheckTests`+`FftKernelsInvariantsTests`+`FFT2DAndNDTests`: 193/0.
- Added `NativeComplexOpsTests.FFT_MatchesNaiveDft_{Double,Float}` (E=256/512/1024) validating `NativeComplexFFT` against an independent O(n²) DFT reference.

## Measured (honest)
Kernel-level (isolated butterfly): **~1.5–2.2× (double), ~1.3–1.4× (float)** vs the scalar path — the scalar butterfly already used cached twiddles, so this is a modest, real win, not a huge one. Note: a naive "40×" earlier compared lean `Fft.RFft` against the *full* `NativeComplexFFT` tensor-op; profiling shows most of that gap is tensor-op **overhead** (Complex result allocation, per-element indexing, AutoTracer plan machinery), NOT the butterfly. Those allocation/indexing hot paths are a **follow-up** (PerfView-guided) on top of this change. Bench numbers are contention-noisy on the dev box; profile on a quiet box.

## Risk
- float FFT now computes butterflies in double then rounds on store → strictly closer to truth than the all-float scalar path (all tests pass; no bit-exact-float asserts observed).
- All callers route through the shared span core; `FFT2DAndNDTests` + `SpectralFilterTests` pass unchanged.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/CpuEngine.cs
+++ b/src/AiDotNet.Tensors/Engines/CpuEngine.cs
@@ -48035,10 +48035,42 @@ public partial class CpuEngine : ITensorLevelEngine
         }
     }
 
+#if NET7_0_OR_GREATER
+    // Interleaved [re,im,re,im,...] scratch buffer reused across the rows of a
+    // batched transform so the SIMD radix-2 delegation allocates at most once
+    // per thread (grown on demand). Never handed out; local to a single call.
+    [ThreadStatic] private static double[]? _fftSimdScratch;
+#endif
+
     // Span-based overload — zero-copy entry point used by NativeComplexFFTSpan hot paths.
     private static void NativeFFTInPlaceDoubleSpan(Span<Complex<double>> data, bool inverse)
     {
         int n = data.Length;
+#if NET7_0_OR_GREATER
+        // Delegate the power-of-2 transform to the SIMD radix-2 kernel used by the
+        // tested Fft.* module. IterativeRadix2NoCache is raw/unnormalized — exactly
+        // the contract of this scalar butterfly (callers apply any 1/n scaling
+        // themselves), so output layout and normalization are identical.
+        if (n >= 2 && AiDotNet.Tensors.LinearAlgebra.Fft.FftKernels.IsPowerOfTwo(n))
+        {
+            var scratch = _fftSimdScratch;
+            if (scratch is null || scratch.Length < 2 * n)
+            {
+                scratch = new double[2 * n];
+                _fftSimdScratch = scratch;
+            }
+            var buf = scratch.AsSpan(0, 2 * n);
+            for (int i = 0; i < n; i++)
+            {
+                buf[2 * i] = data[i].Real;
+                buf[2 * i + 1] = data[i].Imaginary;
+            }
+            AiDotNet.Tensors.LinearAlgebra.Fft.FftKernels.IterativeRadix2NoCache(buf, n, inverse);
+            for (int i = 0; i < n; i++)
+                data[i] = new Complex<double>(buf[2 * i], buf[2 * i + 1]);
+            return;
+        }
+#endif
         int bits = 0;
         for (int tmp = n >> 1; tmp > 0; tmp >>= 1) bits++;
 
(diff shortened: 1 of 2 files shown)


ooples:
### Follow-up commit `56977dd` — allocation reduction (measured with `GC.GetAllocatedBytesForCurrentThread`)

The SIMD wiring above is a modest ~1.5–2× kernel win; profiling showed the real per-call cost on the spectral-FFN path is **allocation**, not the butterfly. Measured bytes/call (`[128,1024]`, pure CpuEngine, AutoTracer off):

| Op | Before | After | Δ |
|---|---|---|---|
| `NativeSpectralFilter` | 26,274,462 B | **1,049,064 B** | **−96.0%** |
| `NativeComplexFFT` | 3,170,840 B | 3,170,840 B | unchanged (return-bound; honestly skipped) |

`NativeSpectralFilter` was allocating ~8 full-size `Complex<T>` intermediate tensors (FFT2D rows + 2 transposes + col-FFT + filtered + IFFT2D 2 transposes + col-IFFT). New internal `NativeSpectralFilterFused` runs the whole FFT2D→multiply→IFFT2D-real round trip over **three `[ThreadStatic]` scratch buffers**, allocating only the returned real tensor (1,048,576 B = theoretical floor). Same `NativeFFTInPlace` core, same 1/H·1/W scaling; column FFTs use gather/scatter (value-identical to transpose→row-FFT→transpose). Gated on `!GraphMode.IsActive` — the composed body runs byte-for-byte under lazy-graph recording.

`NativeComplexFFT`'s allocation is the *returned* `Complex[]` — not reducible without a risky output-parameter overload, so left as-is.

**Verification:** all 3 TFMs build 0-warn; new `SpectralFilter_Fused_MatchesComposed_2D/_Batch4D` assert bit-exactness (<1e-9) on net10 *and* net471; FFT+Spectral suites net471 316/0, net10 330/0.

ooples:
### Honest measurement follow-up: this PR does NOT move the HRE full-model M2 path

A controlled same-session A/B (local nuget of this branch vs its parent, HRE `M2_FullModel_WallClock_StackedBlocks`) shows the C# full-model wall-clock is **unchanged** by this PR — fix vs no-fix are statistically indistinguishable inside the box's ±2× farm-load noise. Reason: the HRE FFN/attention call `NativeComplexFFT` + manual complex-multiply, and the SIMD butterfly this PR wires is only **~5–15%** of `NativeComplexFFT`'s wall-time; the other **~85–95%** is `Complex<T>` result-tensor allocation + non-blittable element-wise fill/read + tracer overhead, which this PR does not touch. `NativeSpectralFilter` (the alloc-fix target) is not on this harness's path.

**This PR is still correct + worth landing** (SIMD butterfly + bit-exact equivalence tests; the `NativeSpectralFilter` −96%-alloc fix for *its* consumers). It's just not the HRE-M2 lever. The real lever is a **blittable interleaved-buffer `NativeComplexFFT` + pooled `Complex<T>` result** (kills the dominant per-call allocation) — a separate follow-up.

PR #863, opened 2026-07-21
feat(ptx): fuse residual LayerNorm GELU

ooples:
## 2026-09-02 merge of main after #859 (`279f772a0`)

Only `run-direct-ptx-ncu.ps1` conflicted (the `-Target` set, unioned with `dense-linear`, and one reworded comment merged two-sidedly). All other touched files auto-merged; verified by building: net10.0 and net471 0 errors, tests and benchmarks 0 errors, dense-linear + normalization + pointwise gates **189 passed / 133 hardware-gated skips / 0 failed**. Pre-existing on main, not from this merge: the script validates `geglu` and `geglu-backward` targets that have no switch entries (#864).

---

## 2026-09-02 merge of main after #864 (`459126672`)

#864 merged first and carried stacked copies of this branch's normalization files, so main held an older residual LayerNorm+GELU kernel (baked epsilon, six launch arguments, rows 2048/8192 promoted) beside this branch's evolved one (epsilon as a launch parameter, seven arguments, promotion held pending p15 re-timing). The seven shared files take this branch's versions, `CudaBackend.DirectPtx.cs` keeps this branch's normalization dispatch and takes main's twelve pointwise GLU members, and the feature gate, parity registry, CLI and profiler script are unioned. Three members both sides had added at different positions were de-duplicated.

> [!WARNING]
> **Consequence to confirm:** once this merges, `CudaBackend.FusedResidualBiasLayerNormGeluD64` returns from promoted (2048/8192 rows, via #864's copy) to HOLD, because this branch's retained topology has not been re-timed under p15. Keep #864's promotion only by re-timing this kernel or by restoring the older kernel and its cubins.

Verification on this host: net10.0 and net471 0 errors; tests and benchmarks 0 errors; normalization + pointwise gates **54 passed / 20 hardware-gated skips / 0 failed**.

---

## 2026-09-01 synchronization and review remediation (`181937d9a`)

Merged current `origin/main` through #1000 (release 0.129.8); merge-tree clean. Adversarial review fixes:

[...]

PR #886, opened 2026-07-23
feat(ptx): direct-PTX kernel foundation (#841) - 5 emitter gaps closed + convolution family, full measured evidence

ooples:
Direct-PTX convolution family for #841, rebased onto `main` (no longer stacked on #863), with the full measured evidence attached. **The evidence is mixed and is reported as such: six kernels beat cuDNN, four lose, and the losses are one coherent family.**

Protocol **p4** throughout — RTX 3080 locked at 1770 MHz, true fp32 (`allow_tf32=False`), competitor in its **CUDA-graph** lane (the strongest form), paired within-sample ratio, batched timed regions, best of 3 at a steady clock, SM clock sampled at both ends of every measurement.

---

## 1. Correctness — 13/13 against an fp64 oracle

Every kernel is emitted from a `CodegenKernelSpec` and compared on-device against the fp64 interpretation of **that same spec**, at the shape that is released (not a smaller proxy — verifying a proxy shape shipped two unexercised lowerings on this branch before the shapes were unified).

```
kernel                              regs  lowering/staged  guards  max abs dev   result
depthwise_conv2d_3x3_bias_relu        40   unroll/none          140    0.000E+000     PASS
depthwise_conv2d_3x3                  40   unroll/none          139    0.000E+000     PASS
conv2d_1x1_bias_relu                  64   unroll/none          452    0.000E+000     PASS
conv2d_3x3_bias_relu                  48   loop x1/weights      168    0.000E+000     PASS
maxpool2d_2x2                         18   unroll/none           20    0.000E+000     PASS
conv_transpose2d_3x3_stride2          77   unroll/none          115    0.000E+000     PASS
depthwise_conv2d_3x3_bwd_data         40   unroll/none          139    0.000E+000     PASS
conv2d_1x1_bwd_data                   40   unroll/none          832    0.000E+000     PASS
conv2d_3x3_bwd_data                   56   loop x1/weights      164    0.000E+000     PASS
depthwise_conv2d_3x3_bwd_weights      40   loop x2/none         225    5.589E-004     PASS
conv2d_1x1_bwd_weights                70   loop x2/none         256    0.000E+000     PASS
conv2d_3x3_bwd_weights                40   loop x2/dOut         150    0.000E+000     PASS
conv2d_1x1_deep_epilogue              54   unroll/none          456    0.000E+000     PASS

verify: 13 passed, 0 failed
```

Twelve are **bit-exact**. The one non-zero row is fp32 accumulation order over 100,352 summed terms, and under the split lowering it *improves* to `3.066E-007` because the split sums 4 partials of 25,088 rather than one serial run.

The nine backward kernels are **derived** from their forward specs by `CodegenAdjoint`, never hand-authored, so a gradient cannot drift from the operator it differentiates.

## 2. Machine code — 16/16 zero-spill

Driver-linked cubins, content-addressed, audited through `nvdisasm`. Sixteen rather than thirteen because three kernels ship a partial and a combine pass; releasing only the partial would leave half the shipped program unaudited.

```
release: 16 zero-spill, 0 spilling
```

## 3. Competitor — six wins, four losses

`artifacts/competitor-ratios.tsv`. Ratio > 1 means we are faster.

| kernel | ours µs | cuDNN µs | ratio |
|---|---|---|---|
| `depthwise_conv2d_3x3_bias_relu` | 75.1 | 224.6 | **2.99×** |
| `depthwise_conv2d_3x3_bwd_data` | 74.0 | 203.5 | **2.75×** |
| `depthwise_conv2d_3x3` | 75.5 | 156.9 | **2.08×** |
| `maxpool2d_2x2` | 163.5 | 230.7 | **1.41×** |
| `conv2d_1x1_deep_epilogue` | 33.3 | 44.9 | **1.35×** |
| `conv2d_1x1_bias_relu` | 31.5 | 39.3 | **1.25×** |
| `conv2d_1x1_bwd_data` | 33.8 | 34.5 | 1.02× |
| `conv_transpose2d_3x3_stride2` | 101.3 | 101.2 | 1.00× |
| `conv2d_1x1_bwd_weights` | 69.7 | 64.3 | **0.92×** |
| `depthwise_conv2d_3x3_bwd_weights` | 248.8 | 195.1 | **0.78×** |
| `conv2d_3x3_bias_relu` | 63.9 | 41.3 | **0.65×** |
| `conv2d_3x3_bwd_data` | 86.8 | 48.9 | **0.56×** |
| `conv2d_3x3_bwd_weights` | 125.8 | 42.1 | **0.33×** |

**The losses are one family: dense 3×3 at large channel counts** — forward 0.65×, data gradient 0.56×, weight gradient 0.33×. That is cuDNN's strongest ground.

## 4. Limiter — which unit is saturated

`artifacts/limiter.tsv`, Nsight Compute counters. A kernel at a roofline cannot be improved by any change to the code generator, only by changing what it has to move.

| kernel | limiter | % of peak | status |
|---|---|---|---|
| `depthwise_conv2d_3x3` | L1 | 93.4 | at roofline |
| `depthwise_conv2d_3x3_bias_relu` | L1 | 93.4 | at roofline |
| `depthwise_conv2d_3x3_bwd_data` | L1 | 93.3 | at roofline |
| `maxpool2d_2x2` | DRAM | 94.6 | at roofline |
| `conv_transpose2d_3x3_stride2` | SM | 82.2 | at roofline |
| `conv2d_3x3_bwd_weights` | L1 | 78.0 | at roofline |
| `conv2d_1x1_bwd_weights` | L1 | 77.1 | at roofline |
| `depthwise_conv2d_3x3_bwd_weights` | L1 | 74.5 | at roofline |
| `conv2d_1x1_bias_relu` | L2 | 64.8 | headroom |
| `conv2d_3x3_bias_relu` | L1 | 59.0 | headroom |
| `conv2d_1x1_deep_epilogue` | L2 | 54.5 | headroom |
| `conv2d_1x1_bwd_data` | L2 | 52.7 | headroom |
| `conv2d_3x3_bwd_data` | L1 | 36.6 | headroom |

**13 of 13 carry both a competitor ratio and a limiter verdict**, which is the release gate. It read 0 of 13 before this evidence was gathered.

## 5. Lowering choices are measured, not modelled

`artifacts/autotune.tsv`. A static cost model chose lowerings four times on this branch and **lost to the hardware every time it was checked** — it predicted a 2.78× occupancy penalty where 1.46× was measured, picked a 4×8 tile slower than the 4×4 it replaced, and called the transposed-conv tile worse on two separate metrics when the hardware said it was 1.12× faster. So candidates are emitted, required to agree numerically, timed, and the winner recorded.

| kernel | modelled µs | best µs | winner | gain |
|---|---|---|---|---|
| `depthwise_conv2d_3x3_bwd_weights` | 4068.9 | 245.1 | `split:4` | **16.60×** |
| `conv2d_1x1_bwd_weights` | 2129.8 | 60.7 | `split:3` | **35.08×** |
| `conv2d_3x3_bwd_weights` | 244.9 | 119.4 | `split:5` | **2.05×** |
| `conv2d_1x1_bias_relu` | 33.3 | 30.2 | `tile2` | 1.10× |
| the other nine | — | — | a tile knob, `no-vector`, `input-staging` | ≤ 1.03× |

### Split-K: a 1081× structural gap, and what it did not buy

`depthwise_conv2d_3x3_bwd_weights` ran **4068.9 µs against a 3.8 µs compute roofline**. The cause was structural, not a tuning miss: dW is `[C,3,3]` = 576 elements, so 576 threads, so **3 blocks on 68 SMs**. The autotuner tried every tile candidate and none moved it, because tiling redistributes work among threads that exist.

`CodegenSplitReduction` gives the reduction its own parallelism — two passes, not `atomicAdd`, because fp32 atomics are order-nondeterministic and would force the exact-agreement gate down to a tolerance.

**But 16.60× over ourselves is still 0.78× against cuDNN.** The split recovered a pathology of our own making and landed *short* of the competitor. Stated the other way: without it we were 21× behind on that kernel; with it we are 1.3× behind. The release gate was right to withhold these three.

## 6. What was tried and failed

Reported because it bounds what this layer can do.

Per-dimension shared-memory staging was implemented, verified correct (13/13 at `0.000E+000`, including the 2D block that previously returned `5.277`), and **falsified its own premise**:

| | L1 | DRAM | SM |
|---|---|---|---|
| baseline | 64.08% | 4.31% | 53.24% |
| staging | **77.45%** | 3.36% | 44.43% |

L1 went **up** — on NVIDIA hardware shared memory *is* L1TEX, so `ld.shared`/`st.shared` are counted by the very metric the lever was meant to relieve. Register reuse is also dead: L1 is flat at 64.27/64.11/64.20 across coarsening 2/4/8.

The warp-stall breakdown says why both were doomed:

| stall reason | % of warp-active cycles |
|---|---|
| wait — fixed-latency / FMA dependency | 17.12 |
| long_scoreboard — global memory dependency | 10.35 |
| short_scoreboard — shared / MIO dependency | 5.13 |
| **mio_throttle — load-pipe queue full** | **3.03** |

`mio_throttle` at 3% means **the load pipe was never the bottleneck**. Warps issue ~64% of the time with no unit above 64%: the kernel is *balanced*, so a 1.5× gap to cuDNN is an **algorithm** gap — implicit-GEMM or Winograd, where F(2,3) alone removes 2.25× of the multiplies. That is outside the index-map layer.

**Conclusion: dense 3×3 is not winnable by this code generator.** The three dense-3×3 losses are recorded as algorithmically out of reach for this layer and routed to the Winograd path rather than retried here. `COMPETITOR_TARGETING_MAP.md` said so before the attempt.

## 7. Scope — no promotion in this PR, and a family-scoped plan for the next one

Nothing dispatches to these kernels today. `DirectPtxFeatureGate.ConvolutionExperimentOverride` defaults to `false` and is set only by the kernel tools; `IsConvolutionEnabled` requires an explicit environment variable. The measured wins are therefore **unrealised outside the benchmark harness**.

That is deliberate, and the agreed plan is to promote **per family, on that family's own evidence** — never globally:

| family | ratio | decision |
|---|---|---|
| depthwise 3×3 (forward, bias+relu, bwd_data) | 2.08–2.99× | **promote** |
| `maxpool2d_2x2` | 1.41× | **promote** |
| 1×1 family (bias+relu, deep epilogue, bwd_data) | 1.02–1.35× | **promote** |
| dense 3×3 (forward, bwd_data, bwd_weights) | 0.33–0.65× | **excluded** — algorithmically out of reach for this layer (§6); route to Winograd |
| weight gradients (depthwise, 1×1) | 0.78–0.92× | **excluded** — behind cuDNN even after a 16.6×/35.1× split |

Promoting a kernel measured at 0.33× would knowingly route work to something three times slower than the alternative, which is why the exclusions are explicit rather than left to a global switch. The dispatch wiring itself is follow-up work, not part of this PR.

## 8. Reproducing

```bash
nvidia-smi -lgc 1770,1770
dotnet run --project tests/AiDotNet.Tensors.Benchmarks -c Release -f net10.0 -- --kernel-verify
dotnet run --project tests/AiDotNet.Tensors.Benchmarks -c Release -f net10.0 -- --kernel-release
dotnet run --project tests/AiDotNet.Tensors.Benchmarks -c Release -f net10.0 -- --kernel-autotune all
dotnet run --project tests/AiDotNet.Tensors.Benchmarks -c Release -f net10.0 -- --kernel-limiter
python tools/bakeoff/run_bakeoff.py
```

Timings are **refused** on a busy GPU rather than reported — an earlier run of the front-end check produced a 64 µs 16K-element ReLU and a 466 µs 512-element reduction while another process held 84% of the SMs, and those numbers were discarded, not published.

## 9. Verification summary

- 282 codegen tests pass
- 13/13 on-device verify (12 bit-exact; one at `5.589E-004`, documented)
- 16/16 cubins zero-spill
- 13/13 carry both evidence columns
- `net10.0` / `net8.0` / `net471` all build

## 10. Relationship to #891

[#891](https://github.com/ooples/AiDotNet.Tensors/pull/891) extracts the **reusable system** from this branch — index-map IR, emitter, front end, split-K, conveyor — as an additive PR against `main`, so the other open direct-PTX PRs can adopt it without waiting on the convolution result argued here. That PR carries no claim that our convolution beats cuDNN. This one holds the per-kernel evidence, including the losses.

### Known limits — the state this PR started from

`docs/SYSTEM_GAP_AUDIT.md` recorded what the system could not express: 11 of 66 IR ops
translated, reduce kinds `Sum`/`Max` only, activations `None`/`ReLU`, zero transcendentals,
fp32 only, one output per kernel, and no tensor-core instructions.

**Those are the gaps section 11 closes.**

---

## 11. The code-generator foundation — five capability gaps closed

The convolution result above is one family. The gaps below are what decided whether *any*
future kernel could be generated at all rather than hand-written, and they were the stated
mandate for this PR. Every one is verified on device against an fp64 oracle.

| gap | was | now | device evidence |
|---|---|---|---|
| **fp16 / bf16** | fp32 hardcoded in 69 places; bindings carried no element type | storage per **binding**, arithmetic always fp32 | fp16→fp32 matmul `0.000E+000` |
| **Tensor cores** | no `wmma`/`mma` emitted at all | `PtxTensorCoreEmitter`, m16n16k16, fused epilogue | 7 shapes at `0.000E+000` |
| **Gather / scatter** | documented "deliberately not expressible" | `CodegenIndirectIndex` on the binding, atomic scatter | 6 rows `0.000E+000`, incl. 4096-way contention |
| **Complex / quaternion** | not expressible | `CodegenAlgebra`, product from a checked table | 5 rows `0.000E+000` |
| **3+ outputs** | exactly two (primary + one argmax) | any number; legacy pair folds into the same list | 3 rows, 2/3/4 outputs |

Three design calls worth naming, each because the alternative fails silently:

- **Storage is per binding, not per kernel.** That is what lets one kernel read fp16
  activations against fp32 weights — the common decode shape. bf16 *shifts* rather than
  converting (no `cvt.f32.bf16` exists here), and narrowing rounds to nearest even, because
  truncation biases every value toward zero.
- **An index tensor is a distinct element type**, refused as an arithmetic operand **at the
  spec**. The first version of that guard sat in `EmitLoad` and a test caught a kernel reading
  the index buffer as fp32 anyway: the emitter has scalar, vectorised, staged and coarsened
  load paths, so a guard on one is a guard on none.
- **A scatter store is atomic by structure**, not by request. An output dimension addressed at
  run time cannot be proven injective; a plain store passes every single-threaded check and
  then loses gradients when warps collide.

Two paths are deliberately kept **separate from the proven real path** — tensor cores
(warp-collective, opaque fragment layout) and complex/quaternion arithmetic (component
vectors). Both refuse what they cannot express exactly, with a reason, and fall back. The
real path is byte-for-byte unchanged, which is why the original 13/13 still verifies at
`0.000E+000`.

### Tensor cores: correct, and NOT competitive

Reported as a loss, because it is one. Against cuBLAS on an idle GPU at locked clocks:

| shape | ours | cuBLAS | ratio | ours | cuBLAS |
|---|---|---|---|---|---|
| 1024³ | 190.1 µs | 66.6 µs | **0.35×** | 11.3 TFLOP/s | 32.2 TFLOP/s |
| 2048³ | 1455.4 µs | 333.8 µs | **0.23×** | 11.8 TFLOP/s | 51.5 TFLOP/s |
| 4096³ | 45810.5 µs | 2384.1 µs | **0.05×** | 3.0 TFLOP/s | 57.6 TFLOP/s |

It beats *our own scalar lowering* 2.05–6.77×, but it loses to cuBLAS and loses worse as the
shape grows. The 11.8 → 3.0 TFLOP/s collapse is diagnostic: one warp per 16×16 tile with no
staging moves `O(M·N·K)` operand bytes instead of `O(M·N·K / tile)`, so when the reused bands
outgrow L2 every warp goes to DRAM. cuBLAS holds ~93% of device peak.

This refines the §1 finding rather than contradicting it: dense GEMM was losing for **two**
reasons, and only one of them was instruction selection. **The next lever is named** —
shared-memory staging of the A/B bands plus multiple tiles per warp — and it is a separate PR
with its own evidence.

**The tensor-core path is therefore wired into no dispatch site.** Anywhere a caller could
reach cuBLAS instead, routing to us would be a 3–20× regression, and the fused-epilogue
advantage saves one launch — nowhere near enough to close it. The capability ships; the
promotion does not. This is the same per-family rule §6 applies to convolution.

### Measurement errors caught before publication

Recorded because each looked publishable:

1. **A WDDM submission floor** had every shape from 64³ to 1024³ reporting 43–87 µs, and had
   cuBLAS's fp16→fp32 lane beating its fp16→fp16 lane despite doing strictly more work. Both
   sides now batch, and small-shape ratios are not reported as findings — 256³ measured 21.7,
   38.0 and 77.5 µs across three runs.
2. **An `O(M·N·K)` CPU oracle** silently stalled the 4096³ run with the GPU idle. Verify
   shapes and timing shapes are now separate; the verified shapes exercise every path the
   large ones use.
3. **An argmax buffer downloaded as float** reported a deviation of exactly 15.0 — the largest
   index in a 16-wide reduction. It looked exactly like a broken kernel and was a broken
   harness.

### Foundation gates

```
codegen tests               282 / 282
--kernel-verify              13 / 13   at 0.000E+000
--frontend-check             24 / 24
--gather-check                6 / 6    at 0.000E+000
--algebra-check               5 / 5    at 0.000E+000
--multiout-check              3 / 3
net10.0 / net8.0 / net471    all build
```

`docs/KERNEL_BLUEPRINT.md` §7 is updated: **nothing on the capability list is still a reason
to hand-write a kernel.** What remains open is performance, not expressiveness.


ooples:
## Promotion status update — remaining tasks

**Cubin committed → strict gate active.** Both cubins (v1 + register-blocked c64) are compiled via the #863 driver-`cuLink` pipeline and committed, isolated in `Artifacts/sm86/convolution/` (a shared-dir bug that deleted the 61 normalization cubins was found and fixed — cleanup is now manifest-aware, and per-operator subdirs prevent collision). `--verify-direct-ptx-convolution-cubins` passes (2 identities / 2 cubins); `direct-ptx-convolution-gate.yml` now runs strict PTX-identity + pinned-`nvdisasm` SASS enforcement in CI. Locally `ptxas -v` already showed **0 spills**.

**Production embedded-cubin path proven.** The register-blocked c64 kernel loads from the **embedded committed cubin with the JIT fallback OFF** and matches the fp64 oracle on-device — i.e. PTX → cubin → committed → embedded → loaded-in-production is end-to-end correct.

**Public dispatch route wired.** `DirectGpuTensorEngine.FusedConv2D` now tries the register-blocked route first (beats cuDNN ~1.60× on its shape), then v1, then the established path — gated on 1×1/stride1/pad0/dilation1 + FP32 + ReLU + bias, fail-closed on every other contract. Capture-safe cache/prewarm/disposal mirror the tested v1 wiring.

**torch.compile competitor lane — environment-blocked here, reported honestly.** Per #863's rule ("attempted as a separate lane, never silently relabeled when unavailable"): torch.compile is degraded on this **Windows** box — triton needs a full CUDA toolkit + MSVC it can't find, so all modes (366–532 µs) run *slower* than eager cuDNN. WSL has no Linux distro and the Docker CUDA path hangs on this machine. The strongest **working** competitor is therefore **cuDNN autotuned in a CUDA graph** (its genuine best), which the register-blocked kernel beats **1.60×** (and beats cuDNN conv-*only*, doing less work, by 1.09×). A fair torch.compile lane needs a Linux GPU runner (triton works there) — best added as a CI job on a GPU runner rather than forced on this Windows dev box.

**Still open (cross-branch):** general dispatch for shape *families* + on-device tile autotune — lands when the #887 autotune foundation (`ConvTileAutotune` / `GpuFirstRunAutotuner`) merges; this exact-shape route is the promotion it generalizes.

PR #927, opened 2026-08-01
test(autodiff): ratchet the gradcheck sweep at zero harness skips

ooples:
## Problem

`DifferentiableOpsGradCheckSweep` asserted on mismatches and on ops that record no gradient, but **not on skips**. A skip means the harness could not drive the op at all, so its gradient is simply **unverified** — indistinguishable from a pass in a green build.

That mattered because the skip count is hard-won: this suite went from **114 skips to 0** by adding per-op argument tables. Without a ratchet, the next op whose parameters the synthesizer cannot handle silently rejoins the unchecked set and nothing reports it.

## Fix

Fail the sweep when `skipped.Count > 0`, with a message naming every offending op and the two ways to resolve it.

Zero is the honest floor, not a number chosen to make the assertion pass — measured **0 skips with 251 ops verified on BOTH target frameworks** (net10.0 and net471), which is why this is a hard 0 rather than a per-TFM budget.

An op that genuinely cannot be gradient-checked keeps its supported escape hatch: the `Exempt` table, which records the reason and keeps the exclusion visible. So a new op must end up either **covered** or **documented**, never silently unchecked.

## Verification

The ratchet was checked to actually **fire**, not merely to stay green. Temporarily disabling the `TensorMatMul` argument-table entry produced:

```
1 could not be driven by the harness at all, leaving their gradients unverified
(add a per-op entry to the argument table so the op can be invoked, or add it to
the Exempt table with a documented reason).
SKIPPED TensorMatMul|T,T: 2D threw ArgumentException: Matrix dimensions incompatible: [2,3] x [2,3]
```

The entry was then restored. Green on **net10.0 and net471** (6/6 including `RfftAndLogVarianceGradTests`).

Current sweep state on both TFMs:

```
gradient-checked OK : 251
MISMATCH            : 0
NO GRADIENT         : 0
exempt (documented) : 7
skipped (harness)   : 0
```

### DIFF
--- a/tests/AiDotNet.Tensors.Tests/Engines/Autodiff/DifferentiableOpsGradCheckSweep.cs
+++ b/tests/AiDotNet.Tensors.Tests/Engines/Autodiff/DifferentiableOpsGradCheckSweep.cs
@@ -796,12 +796,37 @@ public class DifferentiableOpsGradCheckSweep
         _out.WriteLine("");
         foreach (var s in skipped.OrderBy(x => x)) _out.WriteLine("skip  " + s);
 
-        if (mismatches.Count > 0 || noGradient.Count > 0)
+        // A SKIP is an op the harness could not drive at all, so its gradient is simply UNVERIFIED.
+        // That is materially different from a pass, and it is invisible in a green build unless it
+        // is asserted on: this suite went from 114 skips to 0 by adding per-op argument tables, and
+        // without a ratchet the next op whose parameters the synthesizer cannot handle would slip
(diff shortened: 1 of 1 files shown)


ooples:
Refreshed this branch onto current \main\ in \1bfcd336\ after confirming the PR had 0 review threads and all existing checks were green. The documented focused verification still passes after the merge: 6/6 on net10.0 and 6/6 on net471, with 0 skipped on both frameworks. \git diff --check\ is clean.

PR #928, opened 2026-08-04
perf(blas): BCL Vector<T> FP64 streaming microkernel — net471 GEMM was fully scalar

ooples:
## Problem

net471 has no `System.Runtime.Intrinsics`, so `Avx2Streaming` and `Avx512Streaming` compile to stubs whose `IsSupported` is a constant `false` and which delegate straight to `ScalarStreaming`. **Every FP64 GEMM on net471 ran a scalar triple loop** — even though `System.Numerics.Vector<T>` *is* available and JIT-vectorized there (`Vector<double>.Count == 4` on this x64 host).

Found while chasing AiDotNet's `ShiftNetTests.MoreData_ShouldNotDegrade` (250 training iterations): **33.3 s on net10.0 vs a >120 s timeout on net471**.

## Measurement, not intuition

PerfView `/ThreadTime`, leaf (exclusive) CPU in `testhost.net471`, 230,928 samples:

| leaf | % | |
|---|---|---|
| `(clr)!?` | 32.21% | 84.6% native-only stacks; the resolved remainder is xunit discovery/reflection, not model code |
| `ScalarStreaming.RunFp64` | 29.77% | **this PR** |
| `ScalarFp64_4x4.Run` | 12.00% | next lever, still scalar |
| `ScalarFp64_4x4.RunStridedB` | 2.58% | |
| `ScalarPack.PackA/PackB` | ~1.2% | |

GC was **ruled out** rather than assumed: 1,914 collections totalling 15,061 ms of a 223.7 s window = **6.73%**. So this is genuine FLOPs on the wrong kernel, not allocation churn.

## Change

Adds `PortableSimdStreaming` as a dispatch tier between `NeonStreaming` and `ScalarStreaming`, so it also covers any host whose intrinsic sets are unavailable rather than being net471-specific.

Median-of-9, net471 **Release**, x64:

| shape | before | after | speedup |
|---|---|---|---|
| 64×64×16 | 0.2240 ms | 0.0104 ms | **21.5×** |
| 128×128×16 | 0.8999 ms | 0.0376 ms | **23.9×** |
| 256×256×8 | 1.8762 ms | 0.0742 ms | **25.3×** |
| 64×256×24 | 1.4152 ms | 0.0498 ms | **28.4×** |

That exceeds the 4× the lane count alone gives, and **not all of it is SIMD**: the scalar kernel's `i→j→kk` order reads `b[kk*ldb + j]` with stride `ldb` down its inner loop — a cache miss per element — while `i→kk→j` streams `b` and `c` contiguously. The rest comes from hoisting the `transA`/`transB` branches out of the inner loop and dropping bounds-checked index arithmetic for pinned pointers. Sanity check: 0.0104 ms at 64×64×16 is ~12.6 GFLOPS, consistent with 4-lane fp64 AVX without FMA.

## Bit-exact, asserted on bits

Vectorizing across `j` leaves each element's accumulation over `kk` in ascending order, one rounded multiply and one rounded add per step — exactly `ScalarStreaming.RunFp64`'s sequence. The different loop nesting does not reorder any single element's sum.

Vectorizing across `k` instead would need a horizontal lane sum, which reassociates the additions and would break the bit-exact contract `ScalarStreaming` documents against `ScalarFp32_4x4` for the tiny-shape bypass guarantee. The tests compare `BitConverter.DoubleToInt64Bits`, not a tolerance, so a tolerance could not hide such a reassociation — and bit comparison also distinguishes `-0.0` from `0.0`.

## Deliberate scope limits

- **`transB == true` stays scalar** — `b[j*ldb + kk]` strides by `ldb` along `j` and would need a gather. `transA` needs no special case because `aval` depends only on `(i, kk)`.
- **FP32 stays scalar** — `ScalarStreaming.RunFp32` accumulates in `double` on purpose to stay bit-exact with `ScalarFp32_4x4`; a `Vector<float>` path would accumulate in float. The measured cost is fp64 anyway.
- **No `aval == 0` early-out** — it would suppress `0 * Infinity => NaN` and the signed-zero result the scalar kernel produces. Both are covered by tests.
- **Not built on `Compatibility/SimdCompat.cs`** — its net471 `Vector128`/`Vector256` shims are `T[]`-backed, loop element-wise and allocate per `Create`, so they would be *slower* than the scalar kernel they replaced. They exist so intrinsics-shaped code compiles, not so it runs fast.

## Verification

- 20 new tests pass on **both net471 and net10.0**.
- Full `BlasManaged` suite on net471: **208 passed, 41 skipped, 0 failed**.
- Both TFMs build clean, Debug and Release.

## Pre-existing issue found, not caused by this PR

`ManagedVsNativeGemmAudit.Audit_ManagedVsNative_AcrossHotShapes` hard-crashes the net471 test host (`--blame` records `Completed="False"`; the 208-passed summary only counts what finished before it). Verified pre-existing by stashing this change and reproducing on a clean tree. It compares managed GEMM against native OpenBLAS over P/Invoke, so a native access violation would kill the host. Tracked to be fixed in this same PR.

## Summary by CodeRabbit

* **New Features**
  * Added portable SIMD acceleration for FP64 matrix multiplication on supported hardware.
  * Improved vectorized processing for packed and strided layouts, including transposed matrices and scalar tails.
  * Preserved bit-exact numerical behavior, including NaN propagation and signed-zero results.

* **Tests**
  * Added comprehensive coverage for supported hardware, transposition, padded layouts, special values, edge cases, and bit-exact results.
  * Added validation for gradient accumulation across branching autodiff computations.


ooples:
Folded the test-sizing fix from #929 into this branch (fast-forward to e40150ee). `ZeroTimesInfinityStillPropagatesNaN` and `NegativeZeroSignIsPreserved` now size `n` from `Vector<double>.Count` (min 4) and build B/C from it, so both reach the vector body on 2/4/8-lane runtimes, with a guard assert that fails loudly if `n` is ever hard-coded below the lane count again. Resolves the CodeRabbit finding. Superseding #929 since this branch now contains its commit.

ooples:
Superseded by #933 (MERGED). This PR became unmergeable due to GitHub stacked-PR machinery — a sticky flag from #929 (which had been stacked on this branch) made every merge API refuse it and blocked changing its base. #933 is the identical content on a fresh non-stacked branch and is now on main. Closing.

PR #930, opened 2026-08-07
feat(tensor): deferred-storage tensors — shape without allocation

ooples:
## What

Adds `WeightLifetime.Deferred` and `Tensor<T>.CreateDeferred(shape, initializer)`: a tensor that knows its shape but has **not allocated storage**. `Length`, `Shape` and `Rank` are exact and free; the buffer — and the pending initializer — appear the first time the tensor's values are read or written.

```csharp
var w = Tensor<float>.CreateDeferred([10_000, 10_000], HeInit);

w.Length;              // 100,000,000 — exact, allocates nothing
w.IsStorageDeferred;   // true
w[0];                  // buffer created + HeInit applied, exactly once
```

## Why

This separates a tensor's **shape** from its **storage**, so a caller can size a checkpoint, plan a streaming budget, or report a model's parameter count before a single byte is allocated.

Crucially the size comes from the *same construction* that later allocates, so it cannot drift from what you eventually get. A hand-written parameter formula can, and did: one consumer reported 32,385,924 parameters against a real 295,840,220 — nine times out — and because a flat parameter vector is paired by length on restore, a checkpoint silently restored into the wrong tensors.

PyTorch offers this capability as `torch.device('meta')`, but only as an explicit mode the caller opts into and converts back out of; asking an uninitialized lazy parameter for `numel()` raises instead. Here it is automatic — a deferred weight behaves exactly like an allocated one the moment anything touches it.

## How

The implementation is small because **the seam already existed**. `TensorBase` already stored `Length`/`Shape` independently of `_data`, already used `_data.Length != Length` as its "not resident" test, and already funnelled value access through one `EnsureMaterialized()` gate that realizes lazy nodes and rehydrates paged-out streaming weights.

Deferred storage is one more branch in that gate, plus a shape-only constructor whose sole omission is `_data = new Vector<T>(totalSize)`.

The **initializer is deferred along with the allocation** — that is the point. A layer's He/Xavier fill would otherwise touch the buffer immediately and allocate anyway, so the fill is carried with the tensor and applied once, when storage first appears.

## Two things this turned up

- **Four value-reading paths bypassed the materialization gate** and would have observed an empty buffer: `RawStorageSpan` (CpuEngine GEMM, op serialization), `CloneDeepCopy`, and both `Transform` overloads. All now guarded. `DataVector` is deliberately left alone — it is documented as a GPU cache-residency probe that must *not* materialize.
- **Publishing storage before running the initializer is a race**: a second thread sees resident storage, skips the lock, and reads a half-filled buffer. Readiness is now a separate flag published only after the fill completes, inside the lock. The concurrency test covers it.

## Verification

- 10 new tests in `DeferredStorageTensorTests` — metadata exact while unallocated; first read materializes and initializes; initializer runs exactly once including under `Parallel.For`; write-triggered materialization does not clobber the write; `CloneDeepCopy`/`Transform` see initialized values; a 100M-element deferred tensor stays unallocated.
- Full suite: **10,680 passed / 683 skipped / 1 failed**. The single failure, `SmallShapeStreamingDispatchTests.SmallTotalWork_RoutesToStreaming(64,64,64)`, **fails identically on clean `main`** (verified by stashing this branch's changes, rebuilding, and re-running).

## Consumer

`AiDotNet` will point `LayerBase.AllocateLazyWeight` at this — the single chokepoint all 106 weight allocations flow through — so parameter counts become exact and allocation-free across the library. That change needs a package version bump and lands separately.

## Summary by CodeRabbit

* **New Features**
  * Added deferred tensor creation, making shape information available before memory is allocated.
  * Tensor storage and optional initialization now occur automatically on first access.
  * Added deferred weight lifetime tracking.
  * Deferred tensors support cloning, transformations, copying, reductions, zero initialization, and thread-safe one-time initialization.
  * Initialization failures are consistently reported on subsequent accesses.

* **Tests**
  * Added coverage for lazy allocation, initialization, concurrency, copying, transformations, dot products, reductions, and large untouched tensors.


ooples:
Addressed all three critical review findings in d0f76059:

1. **Value-access paths bypassing materialization** — added `EnsureMaterialized()` (plus the operand's, for binary ops) to every public method that touches `_data`/`_storage` directly: `CopyTo`, `AsVector`, `Fill`, `RawWritableStorageSpan`, and the direct-`_data` arithmetic/reduction methods (`Sum`, `Max`, `Mean`, `DotProduct`, `Add(Vector)`, `SubtractInPlace`, `AddInPlace`, `ElementwiseMultiply`, `GetRow`, `GetSlice`, `SetSlice` ×3, `DivideInPlace`, `MaxOverAxis`, and the axis-reduction helpers). Methods already routing through the guarded `AsSpan`/indexer or the `Contiguous()` `return this` fast-path (which keeps a deferred tensor deferred) are unchanged.

2. **Deferred state keyed off mutable `Lifetime`** — replaced the discriminator with an immutable `_storageDeferred` flag set once in the deferred constructor; both `IsStorageDeferred` and the `EnsureMaterialized` gate now use it, so flipping the public `Lifetime` setter can no longer strand the tensor on empty storage.

3. **Readiness published during recursive init / swallowed failures** — a separate `_deferredInProgress` flag makes the re-entrant (initializer-writes-through-indexer) call a no-op that never sets `_deferredReady`; `_deferredReady` is published only after the outermost initializer succeeds. Initializer exceptions are cached in `_deferredFailure` and rethrown on every later access instead of marking a half-filled buffer ready.

Added 7 regression tests (17/17 green on net10.0); builds clean on net10/net8/net471.

PR #943, opened 2026-08-13
test(engine): strengthen complex scan gradient coverage

ooples:
## What changed

- validates all eight complex-scan input gradients under a non-uniform projected objective
- validates the Megalodon-style width-one complex EMA composition through LayerNorm

## Why

The original sum-of-output finite-difference check uses a uniform upstream seed. A projected objective is a stronger VJP guard because it catches output-indexing and adjoint-routing errors that a uniform seed can hide. The composed check also proves the merged scan can serve complex EMA layers without a detached recurrence or a new backend-specific operation.

## Validation

- ComplexDiagonalSsmScanTests: 5/5 pass on net10.0
- AiDotNet Megalodon exact tape-vs-finite-difference input VJP: pass in 852 ms against the merged scan implementation


ooples:
@coderabbitai full review

PR #945, opened 2026-08-14
fix(streaming): make paged tensor access mutation-safe

ooples:
## Summary
- rehydrate paged-out streaming tensors before multi-dimensional reads and writes
- reject mutation through every common writable-storage path for bf16/int8/int4 inference stores
- re-encode native and lossless stores before owner eviction so optimizer/parameter updates survive rehydration
- leave writable owners resident if write-back fails instead of silently losing their current values
- clear both no-upcast int8 and int4 resident forms when their owners are dropped

## Why
`TensorBase<T>.this[params int[]]` bypassed the shared materialization gate even though flat indexing and spans use it. Adaptive base-layer code could therefore index a logically valid but resident-empty streamed tensor.

The original write fix exposed a second integrity contract: bf16/int8/int4 pool snapshots cannot persist arbitrary mutations without compounding quantization, while training's lossless tier must persist them exactly. This change makes quantized inference stores explicitly immutable and implements exact variable-size lossless write-back.

## Verification
- focused access + mutation contracts: 16 passed, 0 failed
- all streaming-focused tests: 225 passed, 0 failed (`net10.0`, Release)
- exact writable round trips cover both FullPrecision and Lossless under forced owner shedding
- mutation rejection covers Bf16, Bf16Stochastic, Int8, and Int4 plus indexer, `SetFlat`, `CopyFromArray`, and writable `Memory`

## New Features

- Streaming tensors automatically restore paged-out data for multidimensional reads and writes.
- Training mode promotes quantized inference storage to writable, lossless storage.
- Lossless streaming data preserves exact values after edits and write-back.
- Matrix multiplication and multi-head attention process quantized weights without floating-point expansion.
- Conv3D operations use an optimized, memory-conscious CPU path.

## Bug Fixes

- Unsupported mutations on quantized streaming tensors now fail clearly.
- Read-only embedding lookups remain safe during eviction.
- Streaming cleanup preserves successfully written changes.
- CPU parallel processing better respects configured thread settings.

### Quantized projection and timeout evidence (2026-08-14)

The latest common-engine revision `7f4983d6fa54d2e04377e100b8bdffa297a36ae0` removes full-weight expansion from int8/int4 2-D and batched TensorMatMul plus fused multi-head attention. It also fuses streaming-store encoding and keeps the wide-output route bounded. No model-specific dispatch was added.

- Release solution build: **0 errors**.
- Focused streaming/quantized suite: **86/86** on .NET 10 and **65/65** on .NET Framework 4.7.1.
- Regression coverage asserts no-upcast behavior and numerical parity for int8/int4 TensorMatMul and MHA.
- Exact UNet3D workload improved from **175,143 ms** to **11,285.9 ms** (**15.5x**).
- A post-fix Transfusion trace no longer contains `TensorBase.AsSpan` / full int8 dequantization in the hot stack. Its remaining paper-scale CPU cold forward is measured by AiDotNet's per-model census rather than hidden by a larger timeout.
- Current GitHub checks: solution build, AVX-512 verification, Codecov, and completed GPU parity shards pass.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/TapeStepContext.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/TapeStepContext.cs
@@ -349,7 +349,7 @@ public sealed class TapeStepContext<T>
                 throw new InvalidOperationException(
                     $"SetFlatParameters requires contiguous parameter tensors. " +
                     $"Call Contiguous() on non-contiguous parameters before using flat access.");
-            var dst = p.DataVector.AsWritableSpan().Slice(p._storageOffset, p.Length);
+            var dst = p.AsWritableSpan();
             src.Slice(offset, p.Length).CopyTo(dst);
             offset += p.Length;
         }
--- a/src/AiDotNet.Tensors/Engines/Autodiff/TensorPool.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/TensorPool.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Concurrent;
+using System.Collections.Concurrent;
 using AiDotNet.Tensors.LinearAlgebra;
 
 namespace AiDotNet.Tensors.Engines.Autodiff;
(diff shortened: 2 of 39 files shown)


ooples:
Addressed the outside-diff data-integrity finding in `071a6848`:

- bf16, bf16-stochastic, int8, and int4 stores now reject mutation at the common writable-storage gate (indexers, `SetFlat`, `CopyFromArray`, writable `Memory`, internal spans/backing arrays, and in-place kernels)
- native and Lossless stores now re-encode current values before eviction; Lossless retains variable-size exact encoding for training
- failed native/Lossless write-back leaves the owner resident, preventing persistence failures from becoming silent parameter loss
- forced owner-shed round trips pass for FullPrecision and Lossless; all 225 streaming-focused tests pass

The finding was outside GitHub's diff range, so it did not create a review thread to resolve manually.

PR #956, opened 2026-08-16
feat(fused): dispatch linearized ADMM through the plan

ooples:
ADMM (Boyd et al., 2011) splits the objective and alternates three sub-steps:

```
x = x - lr*(g + rho*((x - z) + u))     gradient step on the augmented Lagrangian
z = soft_threshold((x + u)/rho, l1)    proximal step on the split variable
u = u + (x - z)                        dual ascent on the multiplier
```

All three are elementwise given `z` and `u`, so this is an ordinary fused kernel despite not being an SGD-shaped update: the state is two per-parameter buffers (`z` = m, `u` = v) and two scalars. The linearized x-step (Boyd §6.3) is what makes it one step rather than an inner solve.

`l1 = 0` makes the soft-threshold the identity, which is the correct prox for an unregularized split, so one parameter covers both cases without a mode flag. `weightDecay` is rejected at configure time: regularization enters ADMM through the prox, and accepting a second, differently-scaled shrinkage would silently change the objective being solved.

## Also fixes three dropped Extras fields

`LbfgsMemorySize` and `TrustRegionRadius` were added in #953 without reaching `CloneFusedOptimizerExtras` or the checkpoint serializer. So an L-BFGS plan configured with a memory of 20 **reported and restored as the default 10**, and a trust region of any radius restored as 1.

This is the same trap #949 hit with `Nesterov` and `FtrlBeta` — the clone is what the runtime state, and therefore the checkpoint, records, so a field that doesn't reach it makes the plan describe an optimizer it isn't running. `AdmmRho` is wired into both from the start.

## Tests

5 new, in `AdmmFusedTests`:

- multi-step agreement with an independent transcription at length 67, so the SIMD body and scalar tail both run and `z`/`u` accumulate enough for errors to separate;
- the rho coupling reaching the parameter update, on a zero-gradient fixture where a kernel ignoring it wouldn't move at all;
- `l1 = 0` leaving the split variable exactly `(x + u)/rho`;
- a non-zero `l1` driving small coordinates to exactly zero;
- the dual variable accumulating the **post**-update residual, asserted against the pre-update value a plausible-but-wrong ordering would produce.

Existing optimizer + checkpoint suites: 371 passed, 0 failed.

## Enum numbering

`ADMM = 25`, skipping 24, which is reserved for `ConjugateGradient` — that kernel is written and lands separately, and renumbering afterwards would change the meaning of values already written into serialized checkpoints.

Unblocks the `IFusedOptimizerSpec` for ADMM in ooples/AiDotNet#1930.

## Summary by CodeRabbit

* **New Features**
  * Added ADMM optimizer support for compiled float, grouped-float, double, and grouped-double training workflows.
  * Added SIMD and scalar processing with L1 regularization, parameter coupling, and configurable rho settings.
  * Preserved optimizer-specific settings across checkpoint saves and restores.

* **Bug Fixes**
  * ADMM now rejects unsupported nonzero weight decay configurations.

* **Tests**
  * Added comprehensive coverage for grouped and ungrouped updates, state persistence, validation, and numerical consistency.


ooples:
Rebased onto main (which now carries the merged LARS #955 and ConjugateGradient #957) and merged with the review-feedback commit that had been pushed to this branch — that commit is preserved, not overwritten. `ConjugateGradient` keeps enum value 24, `ADMM` takes 25.

All three review findings are addressed:

- **z/u buffers were never allocated.** ADMM was in neither the `needsMomentum` nor the `needsSecondMoment` predicate, so both slots were `Array.Empty<float>()` and the first `Step()` took a fixed pointer to an empty array. The kernel tests could not see it — they allocate their own arrays. Now in both predicates on every configure path.
- **`AdmmRho` validated** at configure time. `rho = 0` divides by zero in the coupling scale; a negative rho pushes the parameter *away* from the split variable. Both previously surfaced as NaN several steps into training.
- **Format version 3 → 4.** Four values were appended to the checkpoint that no v3 writer emits (`LbfgsMemorySize`, `TrustRegionRadius`, `AdmmRho`, and the CG beta denominator from #957), so a v3 file would short-read under the v4 reader.

Two tests added on top of the ones already here, both covering gaps those could not:

- the **grouped** configure path, which keeps its own copy of the allocation predicates and which ADMM can reach because it is not on the global-state exclusion list;
- **cross-step state**, compared against an independent four-step transcription — buffers reallocated or zeroed each step would still give a finite descending trajectory, so "it ran and nothing was NaN" does not prove z and u persist.

Suites: 716 passed. The one failure, `PlanStitchingAllocationProbe.StitchedExecute_AfterWarmup_AllocatesZeroBytes`, fails identically on a clean checkout of `origin/main` and is unrelated to this branch.

ooples:
@coderabbitai review

All three findings from the previous review are addressed and every thread is resolved; the branch has also been rebased past the merged #955 and #957. Requesting a re-review so the stale changes-requested decision can be cleared.

PR #959, opened 2026-08-16
feat(fft): generic-precision, arbitrary-length GPU FFT path

ooples:
## Problem

`IFftBackend` ships **float32 radix-2 only**, and its contract states that callers with a non-power-of-two length "must route through the CPU Bluestein path." Two costs follow from that:

1. **Length.** A device-resident tensor whose transform length is not a power of two is copied to the host, transformed, and copied back. Qwen2.5-0.5B has hidden width **896 = 2⁷ × 7**, so the model this library is most often pointed at lands on exactly that path — a spectral operator running per layer pays the round trip every layer.
2. **Precision.** Every buffer is float32. A caller holding bf16 or fp16 activations must widen before the transform and narrow after: two extra full passes over the activation tensor, on kernels that are memory-bound to begin with.

The second one is measurable. Running a spectral FFN under `torch.autocast` for comparison — which has the same limitation, leaving FFTs in fp32 — bf16 made the operator **25% slower**, because every FFT got wrapped in bf16→fp32→bf16 conversions. The conversions cost more than the narrow arithmetic saved.

## What this adds

- **`FftElementType`** — `Float32`, `Float16`, `BFloat16`, with byte size, kernel suffix, and minimum compute capability.
- **`IFftGenericBackend`** — a capability interface alongside `IFftBackend`, following the same optional-interface pattern (declared separately rather than as a default interface method, because the library targets net471).
- **`CudaFFTGenericKernels`** — CUDA source generic over storage type and unrestricted in transform length.

### Narrow storage, float32 arithmetic

This is the load-bearing design decision. An *n*-point Cooley-Tukey transform is a chain of log₂(n) accumulation stages; accumulating in a 7-bit mantissa loses roughly a bit per stage, and a 1024-point transform accumulated in bfloat16 retains almost nothing.

So the narrow type appears **only at the load/store boundary**. Every butterfly widens on load, computes in float32 registers, and narrows on store. That is where the benefit is anyway: each butterfly reads two complex values and writes two back, so halving the element width halves the bytes moved, and these kernels are bound by traffic rather than arithmetic.

The tests pin this as an invariant (`ArithmeticIsFloat32ForEveryElementType`) so a future "optimisation" to narrow accumulators fails CI rather than silently degrading transforms.

### Device-side Bluestein

Arbitrary *n* now stays resident: chirp-multiply → zero-pad to *m* = next_pow2(2n−1) → radix-2 → pointwise multiply by the pre-transformed chirp → inverse → chirp-multiply.

The chirp convolution kernel is **extended symmetrically** — `b[j] = c[j]` for `j < n` *and* `b[m−j] = c[j]` for `1 ≤ j < n`. This is not cosmetic. An *m*-point FFT computes a **cyclic** convolution; Bluestein needs a **linear** one, and the two agree only when the kernel is mirrored into the upper tail. Omitting the mirror leaves the transform correct at k=0 and wrong everywhere else — the kind of bug that passes a smoke test on DC.

A reference implementation of that exact omission measures **0.71 relative error at n=896**, so it is carried in the suite as an explicit negative control (`BluesteinRequiresSymmetricChirpTail`) rather than left as a comment.

The chirp phase is also computed as `k²  mod 2n` in **64-bit before the float conversion** — `k*k` overflows `int` past n ≈ 46341, and reducing after conversion loses the low bits that carry the phase.

### Twiddle tabulation

The existing butterfly calls `cosf`/`sinf` per element per stage. Twiddles are now tabulated once per transform and indexed as `tw[wing * (n/stride)]`, removing log₂(n) transcendental pairs per element and leaving the butterfly purely load/store bound.

## Tests

34 tests, all green on net10.0. Because the CUDA kernels need a device, the tests lock down the two things verifiable without one — and they are where the bugs actually live:

- **The algorithm**, executed in managed code with the *same* twiddle indexing and the *same* Bluestein construction, checked against a direct DFT: radix-2 at n = 2/8/64/256, Bluestein at n = 3/5/12/100/**896**/900/1023, forward and inverse, tolerance 1e-9.
- **The negative control** proving the symmetric tail is load-bearing.
- **Source generation** per element type: storage type, conversion intrinsics, header includes, and that every declared entry point is present in the emitted source.
- **Helper invariants**: `BluesteinLength`, `IsPowerOfTwo`, byte sizes, compute-capability requirements.

The reference validation that caught the missing symmetric tail during development is at `hre_port_tests/fftref.py` (research repo, not shipped here).

## Build

`AiDotNet.Tensors` builds clean on **net10.0, net8.0 and net471**, 0 errors.

## Scope

This adds the kernel source, the element-type model and the capability interface. **Backend wiring is deliberately not included** — `CudaBackend` implementing `IFftGenericBackend`, module compilation and launch plumbing are a follow-up, so this PR stays reviewable and the algorithm can be scrutinised on its own. Nothing here changes any existing code path; `IFftBackend` and `CudaFFTKernels` are untouched.

## Summary by CodeRabbit

* **New Features**
  * Added generic complex FFT support for arbitrary transform lengths, including inverse transforms and batched data.
  * CUDA acceleration now supports Float32, Float16, and BFloat16 storage where hardware allows.
  * Added FFT capability detection for supported storage types.
  * Added automatic handling for power-of-two and non-power-of-two transforms.
  * FFT operations now integrate with automatic differentiation.

* **Bug Fixes**
  * Unsupported FFT configurations now report clear errors instead of silently falling back to CPU execution.

* **Documentation**
  * Updated FFT documentation to describe arbitrary-length CUDA support and backend limitations.


ooples:
@coderabbitai full review

PR #960, opened 2026-08-17
fix(fused): RAdam rectification precision and Nadam bias-correction placement

ooples:
Two fused-kernel update rules that disagreed with AiDotNet's eager implementations. Both were found by AiDotNet's fused-vs-eager parity gate **after fixing a bug that had made that gate compare fused against fused** — it disabled compilation by mutating `TensorCodecOptions.Current`, which returns a fresh instance per read when unset, so every numerical assertion in it had been passing vacuously since it was written.

## RAdam: rectification scalars in float

`r_t` depends on `sqrt(rho_t - 4)`, and `rho_t - 4` is near **zero** on the first rectified step — so float error in `rho_t` is amplified there rather than damped. At the default `beta2 = 0.999`:

| step | rho_t (double) | rho_t (float) |
|---|---|---|
| 3 | 2.9987 | 3.0112 |
| 5 | 4.9960 | 4.9865 |

At step 5 that's a ~1% error in `(rho_t - 4)` exactly where `r_t` is most sensitive to it. The branch itself (`rho_t > 4`) was *not* flipping — both precisions agree on which step rectification starts — so this is purely the magnitude of `r_t`.

Measured divergence after 40 training steps: **2.4e-3 → 5.1e-7**, against a 6.1e-7 float-ordering control.

These are per-step scalars, not per-element, so double costs nothing measurable; only the narrowed `r_t`, `bc1` and `bc2` enter the element loop. Applied to `SparseRAdamUpdate` too, so a parameter doesn't take a different trajectory depending on whether its gradient arrived sparse — the dense/sparse divergence that made the LARS kernel wrong in #955.

## Nadam: the two bias corrections were swapped

Dozat (2016), Algorithm 8:

```
m_bar = (1 - b1) * g / (1 - b1^t)  +  b1 * m / (1 - b1^(t+1))
```

The **gradient** term takes the `t` correction; the **momentum** term takes `t+1`. This kernel had them the other way round. At `b1 = 0.9`, step 1, those divisors are 0.19 and 0.1 — each term off by nearly 2×, in opposite directions, largest exactly where Nesterov momentum is meant to help most.

AiDotNet's eager Nadam was *also* wrong, differently (it used `t` for both). Both sides are corrected in the same change, so the paths agree with each other **and** with the paper rather than merely with each other.

Measured divergence: **0.182 → 4.6e-5**, against a 4.6e-7 control.

## Verification

Tensors optimizer suites: 351 passed, 0 failed. On the AiDotNet side the full parity gate is 22/22 with these kernels, and the optimizer suite is stable at the 8 known pre-existing failures across consecutive runs.

Refs ooples/AiDotNet#1930.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected Nadam optimizer bias correction for more accurate parameter updates.
  - Improved RAdam’s bias-correction and rectification calculations, increasing numerical accuracy for dense and sparse training workloads.
  - Updated optimizer behavior consistently across supported execution paths.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/FusedOptimizer.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/FusedOptimizer.cs
@@ -1558,14 +1558,23 @@ internal static class FusedOptimizer
         float* param, float* grad, float* m, float* v, int length,
         float lr, float beta1, float beta2, float eps, int step)
     {
-        float bc1 = 1f - MathF.Pow(beta1, step);
-        float bc2 = 1f - MathF.Pow(beta2, step);
-        float rhoInf = 2f / (1f - beta2) - 1f;
-        float rhoT = rhoInf - 2f * step * MathF.Pow(beta2, step) / bc2;
-        bool rectified = rhoT > 4f;
+        // These are PER-STEP scalars, not per-element, so double costs nothing measurable and buys real
+        // accuracy where it matters most. In float, rho_t carries ~0.4% error (at beta2 = 0.999, step 3:
+        // 3.0112 against a true 2.9987), and r_t depends on sqrt(rho_t - 4) — a term that is near ZERO on
+        // the first rectified step, so that relative error is amplified rather than damped. Measured
+        // against AiDotNet's eager RAdam, which computes these in double, the float version diverged by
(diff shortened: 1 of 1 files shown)


ooples:
@coderabbitai review

ooples:
@coderabbitai full review

Issue #837, opened 2026-07-21
[PTX] Port quantized, mixed-precision, and structured-sparse linear kernels


ooples:
Parent epic: #833

Foundation PR: #832

## Operation scope

Implement PTX blueprint kernels for FP8, INT8/W8A8, INT4, BF16/FP16 mixed precision, dequantization GEMM, `Enforce2x4Sparsity`, `Decompress2x4Sparse`, semi-structured sparse dispatch, `FusedSparseLinear`, sparse GEMM+bias/activation, and supported quantize/dequantize conversions.

Fuse scale/zero-point decode, sparse metadata decode, matrix accumulation, bias, and activation where semantically possible. Compare against the strongest cuBLASLt/cuSPARSELt/cuSPARSE and PyTorch implementations available for each format.

### Inventory and semantics

- [ ] Enumerate every public AiDotNet operation and existing CUDA/NVRTC/library-backed kernel in this scope; the issue is not complete while any scoped entry point is unassigned.
- [ ] Define the intended fusion boundaries, supported forward/backward modes, dtypes, logical shapes, physical layouts, numerical modes, and architecture matrix.

### PTX blueprint and ABI

- [ ] Implement the operation as direct Driver-API-loaded PTX, ported from the existing kernel or written from scratch; CUDA C/NVRTC hidden behind a new wrapper does not count.
- [ ] Give every admitted specialization a versioned `DirectPtxKernelBlueprint` with PTX SHA-256, GPU UUID/SM/driver identity, block geometry, resource budget, and occupancy.
- [ ] Declare formal tensor contracts for logical/physical extents, dtype, layout identity, alignment, byte offset, allocation/view extent, access, aliasing, and padding.
- [ ] Validate the contract once before dispatch. Admitted PTX must contain no runtime stride/layout checks and no generic dynamic-shape branches.
- [ ] Keep Ampere, Ada, Hopper, and Blackwell specializations separate. Unmeasured architectures and semantics must fail closed to the established backend with an exact reason.

### Fused dataflow

- [ ] Document every global-memory read/write and identify which intermediates remain in registers or shared memory.
- [ ] Eliminate avoidable global intermediates and repeated reads; use `cp.async`, Tensor Cores, warp reductions, vectorized transactions, or architecture-specific instructions wherever they are appropriate and measurably beneficial.
- [ ] Prove required alignment/coalescing and shared-memory banking for every physical layout; semantic padding may not be introduced silently.

[...]

PR #967, opened 2026-08-17
feat(ptx): bf16 tensor-core wmma GEMM (naive + staged)

ooples:
## What

Adds **BF16** multiplicand support to `PtxTensorCoreEmitter` alongside the existing FP16 path, on **both** the naive and staged wmma lowerings. This is the BF16 GEMM slice of #837 (the direct-PTX mixed-precision linear kernels), split out of the #860 stack because the tensor-core emitter is already on `main`, so it can land independently.

## How

- **Recogniser** (`TryPlan`): accepts matched bf16 multiplicands with fp32 accumulate; refuses bf16 below **sm_80** (bf16 wmma is Ampere+, fp16 is sm_70+) with a named reason. `Plan.AbType` carries the type.
- **Emission** (naive + staged): `.global`/`.shared.bf16` fragment loads and the **explicit** `wmma.mma…m16n16k16.f32.bf16.bf16.f32` (bf16 is not the PTX default). Byte math is unchanged — bf16 is 2 bytes like f16. FP16 emission is byte-identical.

## Tests

New codegen tests on both paths: `PlainBf16MatMul_IsEligible`, `EmittedKernel_Bf16_UsesBf16Wmma`, `Bf16_BelowSm80_IsRefused`, `StagedKernel_Bf16_UsesBf16Wmma`. `CodegenTensorCore` suite **74/74 green** off `main`, no fp16 regression.

GPU championship (p15 + Nsight zero-spill) for the bf16 path remains pending on Ampere+ hardware; this PR lands the assembler-correct codegen. Part of #837.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/Codegen/Ptx/PtxTensorCoreEmitter.Staged.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/Codegen/Ptx/PtxTensorCoreEmitter.Staged.cs
@@ -94,6 +94,12 @@ public sealed partial class PtxTensorCoreEmitter
     private int _warpTilesM = 2;
     private int _warpTilesN = 2;
 
+    // Multiplicand PTX type for the staged wmma path: "f16" (sm_70+) or "bf16" (sm_80+). Set from the plan at the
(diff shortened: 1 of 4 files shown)


ooples:
@coderabbitai review

PR #972, opened 2026-08-20
fix(compilation): skip a fused optimizer step whose gradients are not finite

ooples:
## Outcome

Compiled training now treats non-finite gradients as an atomic, recoverable skipped step on both CPU and GPU. The original iteration protected only materialized CPU arrays; GPU-resident gradients, uploaded host gradients for GPU parameters, grouped plans, and the multi-tensor fast path could still update with NaN or infinity. This revision closes those gaps.

## New capabilities

- Detects `NaN`, `+Infinity`, and `-Infinity` before any fused optimizer update.
- Covers CPU gradients and device-resident gradients through the shared direct-GPU backend contract used by CUDA, HIP, OpenCL, Metal, Vulkan, and WebGPU.
- Exposes `LastStepSkippedNonFiniteGradients` and `NonFiniteStepsSkipped` so callers can distinguish an intentionally discarded step from stalled learning.
- Allows the next finite step to recover normally from the last valid model and optimizer state.
- Preserves the multi-tensor Adam/AdamW fast path for finite gradients.

## Existing behavior fixed

- A bad gradient can no longer partially update earlier parameters before a later tensor is inspected.
- Skipped steps leave parameters, optimizer moments, Schedule-Free visible weights, and Adam's bias-correction step unchanged.
- The update consumes the exact gradient source that the preflight checked. This fixes the grouped route's former unconditional trust of a stale resident gradient when eager backward had produced newer host values.
- Lazy gradients that materialize after optimizer configuration are rebound before inspection.
- GPU-resident parameters no longer bypass the guard merely because their host parameter arrays are empty.
- Per-tensor, grouped, and multi-tensor GPU optimizer routes now have the same skip semantics as the CPU route.

## GPU design and performance contract

The GPU path does not download gradients. It classifies each device gradient into an IEEE-bitwise 0/1 finite mask, combines every mask on-device, and downloads one scalar reduction result per distinct backend.

- CPU-authoritative gradients are scanned first, so a known host failure launches no GPU work.
- GPU masks are aggregated before readback, preventing one device synchronization per parameter.
- Persistent scratch is bounded to two buffers sized to the largest tensor per backend, rather than one allocation per gradient or the sum of model sizes.
- `Min(mask) == 1` is used instead of summing mask values, so a single zero cannot disappear through FP32 rounding for tensors larger than 2²⁴ elements.
- The mock contract asserts two GPU gradients produce two classifications but only one scalar reduction, zero full-buffer downloads, and no optimizer call when rejected.

## CUDA defects found by hardware validation

The physical-GPU probe exposed two pre-existing scalar-reduction bugs that mock tests could not reveal:

1. The final warp used a partial active mask while shuffling from inactive lanes. A four-element all-ones mask could reduce to zero and falsely reject every optimizer step.
2. Multi-pass reductions retained a scratch buffer whose physical capacity exceeded the final logical size, then attempted to download it into a one-element destination.

The CUDA sum/min/max final-warp reduction now uses identity-filled lanes with a full warp mask, and scalar readback always lands in an exact one-element device allocation.

## Validation evidence

| Evidence | Result |
|---|---:|
| Focused .NET 10 suite | 26/26 passed |
| Focused .NET Framework 4.7.1 suite | 25/25 passed |
| Related fused dispatch, GPU residency, and GPU classification regressions | 69/69 passed |
| Physical GPU | NVIDIA GeForce GTX 1660 Ti, CUDA compute capability 7.5 |
| Hardware values | finite, NaN, +Infinity, -Infinity |
| Hardware reduction lengths | 1, 4, 31, 32, 33, 255, 256, 257, 1025 |
| Full gradient readbacks in GPU route tests | 0 |
| Scalar reductions for two gradients on one backend | 1 |
| Merge with current `main` | clean |

The boundary sizes cover scalar/small inputs, both sides of a CUDA warp boundary, both sides of the 256-thread block boundary, and multi-pass reduction. The live probe executes the same Fill → ClassifyFloat → on-device mask aggregation → scalar Min composition used by compiled training.

## Route coverage

- CPU float and double, including grouped schedules
- GPU-resident and host-authoritative gradients for GPU parameters
- stale resident-gradient selection
- per-tensor SGD route
- multi-tensor Adam/AdamW route
- grouped GPU route
- skip recovery and optimizer-step accounting
- Schedule-Free restoration
- SIMD vector boundaries and scalar tails

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
+++ b/src/AiDotNet.Tensors/Engines/Compilation/CompiledTrainingPlan.cs
@@ -6679,14 +6679,44 @@ internal sealed class CompiledTrainingPlan<T> : ICompiledTrainingPlan<T>
                     else { var r = eng.TensorAdd(a, b); r.AsSpan().CopyTo(o.AsWritableSpan()); }
                 };
             }
-            // Broadcast path: copy a into output, then in-place add b (b is the
-            // smaller operand and broadcasts up to a's shape).
-            return eng =>
+            // Broadcast path: seed the output with whichever operand ALREADY CARRIES THE OUTPUT
+            // SHAPE, then broadcast the other one up into it.
+            //
+            // WHICH OPERAND IS THE LARGER ONE IS NOT KNOWN IN ADVANCE, and this path used to assume
+            // it was always `a`. Seeding from `a` when `a` is the SMALLER operand copies only a's
+            // elements into the full-size buffer and leaves the remainder holding whatever was
+            // there before -- zeros from a fresh allocation, or a previous step's bytes from a
+            // recycled one. Addition is commutative, so seeding from the full-shape operand
+            // instead is the same arithmetic with every element actually written.
+            //
+            // This is reached constantly: TensorBroadcastTo expresses a genuine expansion as
+            // TensorBroadcastAdd(input, zeros(targetShape)), whose FIRST operand is by construction
+            // the small one, so every such broadcast replayed through a compiled plan was wrong.
+            // MEASURED on [4,1] -> [4,3]: a correctly SHAPED 12-element output holding the 4 source
+            // values contiguously and 8 stale slots -- a silent wrong answer at exactly 1/3 of the
+            // true sum when the buffer was fresh, and NaN/1e35 when it was recycled, which then
+            // poisons every parameter through the gradient.
+            bool aIsFullShape = ShapeMatchesBuffer(a._shape, o._shape);
+            bool bIsFullShape = ShapeMatchesBuffer(b._shape, o._shape);
+            if (aIsFullShape || bIsFullShape)
             {
-                a.AsSpan().CopyTo(o.AsWritableSpan());
-                if (eng is CpuEngine cpuEng) cpuEng.TensorBroadcastAddInPlace(o, b);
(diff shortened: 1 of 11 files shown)


ooples:
**Correction to the motivation, from measurement.**

I opened this citing a downstream AiDotNet model (BiaffineNER) whose test class is green in isolation but produces NaN in a full run, and I attributed that to this defect. I built the fix, packed it locally as `0.129.0-nanguard`, pinned AiDotNet to it and ran that class three times. **It did not fix it** — identical failures, and the message says why:

```
Loss became NaN during training: short=NaN, long=NaN
```

The **loss** is already NaN, so the NaN originates upstream of the optimizer. This guard stops a non-finite gradient from being written into the weights; it cannot help when the forward pass produced NaN before any update. My inference — `ComputeGradients` clean but `Train` NaN therefore the fused update must be corrupting weights — skipped checking whether the loss was non-finite *before* the update. That was the error.

**What this PR does and does not claim now:**

- It **does** fix a real defect: every fused kernel reads `grad` and writes `param` in one pass with no finiteness check, so a single non-finite gradient is written into the weights permanently, with no way for a caller to detect it. `torch.amp.GradScaler` exists precisely to prevent this. The tests demonstrate the behaviour directly.
- It is **defensive hardening**, not the fix for that downstream model. I have no evidence yet that any specific failing model is cured by it.

The scope note stands as well: this gates the CPU path only; the multi-tensor GPU path needs an equivalent device-side reduction.

PR #975, opened 2026-08-21
fix(compilation): record RoPE dependencies in fused training

ooples:
﻿## What this fixes

Compiled/fused transformer training did not record `ApplyRoPEInterleaved` in `GraphMode`. Eager autodiff recorded the rotation, but compiled attention treated rotated Q/K tensors as trace-time leaves. Those leaves were allocated from pooled storage; once reused, the first downstream key permute intermittently read NaNs. The optimizer safety work in #972 correctly refused to mutate weights, but that guard could only contain the damage?it could not make training progress.

This PR fixes the producer instead of disabling fused training or weakening the non-finite guard. `ApplyRoPEInterleaved` is now a variadic compiled node with input/cos/sin dependencies and the existing inverse-rotation backward. CPU replay executes the CPU primitive; DirectGpu replay dispatches through the same `IEngine` entry point to the active GPU backend kernel.

## User-visible result

- RoPE-based transformers can use fused compiled training without a model-specific opt-out.
- Compiled forward replay no longer freezes trace-time rotated Q/K tensors.
- Compiled backward propagates gradients through RoPE to Q/K projections and upstream parameters.
- CPU keeps generic numeric-type behavior; coverage pins `double`, which is the failing AiDotNet DocOwl configuration.
- GPU keeps the speed-first float path and backend dispatch; coverage pins active-device `float` forward and analytical backward parity.
- The #972 failure-atomic non-finite optimizer guard remains enabled as defense in depth.

## Evidence

Root-cause trace on AiDotNet DocOwl:

- Eager gradient: 64,672 / 68,064 values non-zero, 0 non-finite.
- Broken compiled gradient: 68,064 / 68,064 values non-finite; #972 skipped the optimizer update, leaving parameters finite but unchanged.
- First corrupt compiled forward: `TensorPermute` after attention RoPE; its input had no compiled producer because `ApplyRoPEInterleaved` was tape-only.

Post-fix downstream stress:

- DocOwl fused-training regression: 12 / 12 sequential passes.
- Contended DocOwl fused-training regression: 4 / 4 simultaneous passes.
- No `SupportsFusedCompiledTraining = false` escape hatch was added.

Tensors verification:

- `CompiledRoPEInterleavedTests`: 3 / 3 pass on net10.0, including the active DirectGpu test (not skipped) and pre-capture invalid-rank validation; 2 / 2 pass on net471.
- RoPE focused suite: 5 pass, 1 environment-specific backend skip.
- RoPE + fused non-finite optimizer suite: 28 / 28 pass.
- Library builds: net10.0, net8.0, and net471 all pass with 0 errors.

The CPU regression replays eight per-step arenas and compares both loss and every input-gradient element to the closed-form RoPE derivative. The GPU regression compares compiled loss and every gradient element to the same CPU analytical control. Invalid input is now rejected before graph recording, and the GPU test skips only when the constructed engine explicitly reports no backend; unexpected constructor failures remain visible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compiled graph execution for interleaved rotary positional embeddings (RoPE).
  * Preserved correct forward results and gradient calculations during training.
  * Added validation for invalid RoPE configurations during graph capture.

* **Tests**
  * Added coverage for compiled CPU training with RoPE.
  * Added GPU validation where supported, including loss and gradient comparisons.


ooples:
﻿Addressed both review findings in `90fc2324`: RoPE validates rank/head shape before GraphMode records a node, backed by a pre-capture regression test; the DirectGpu test no longer blankets constructor failures. Verification: 3/3 net10.0 (active GPU, no skip), 2/2 net471, and net8.0 library build with 0 errors. Both threads are resolved.

@coderabbitai review

ooples:
Addressed the follow-up review in a4f43a84: GraphCapture_RejectsInvalidRoPERankBeforeRecording now asserts scope.NodeCount is zero after the expected validation exception. Focused CompiledRoPEInterleavedTests: 3/3 passed on net10. The review thread is resolved. @coderabbitai review

PR #1017, opened 2026-09-09
test: assert gemm determinism arithmetically, not just by flag

ooples:
DeterministicByDefaultTests covers the determinism contract at the level of
FLAGS: that Deterministic defaults to true, that BlasProvider.IsDeterministicMode
follows it, that a plan instance is reused. None of it runs a GEMM and compares
numbers, so a dispatcher that honours every flag and still returned different
arithmetic on two calls would pass the whole file.

That gap matters because BlasManaged selects between kernels on properties the
caller does not fix: whether pre-packed operands were supplied
(`options.PackedA is null && options.PackedB is null` gates the GotoGemm path),
the thread budget, and what the CPU reports (Avx.IsSupported,
Avx2.IsSupported && Nr == 16). Those kernels block the reduction differently and
floating-point addition is not associative, so two numerically equivalent routes
can disagree in the last bits. GotoGemm documents itself as "computed by one
thread in fixed K order => thread-count-independent, Deterministic/
DisableAutotune-contract safe", which states plainly that thread-count
independence is a property some paths have and others do not.

Added:
  - the same GEMM twice in one process is BIT-identical (3 shapes)
  - a GEMM is bit-identical regardless of the thread budget (2 shapes, 1 vs
    2/4/all cores)

Bit-identical rather than "close" on purpose. Both results are approximately
correct; the question is whether they are the SAME, because a consumer
comparing a model against its clone through an iterative sampler compounds any
difference at all. Inputs span several magnitudes so that reordering a sum
actually changes the result - values of one magnitude can sum identically in any
order and would hide a real reordering.

WHY: a downstream Clone() test fails intermittently on shared CI and passes on
idle machines. That signature is a run-to-run difference rather than a clone
defect, and it matches the signature already recorded in
Issue318CloneDriftDiagnosticTests - trained weights vs SetParameters-cloned
weights producing different MatMul output from byte-identical logical contents,
where "the drift persists" after PR #315.

These pass on the author's CPU, as do the Issue #318 diagnostics and the
downstream test itself. They are added so the mechanism is checked directly, in
CI, on the hardware where the symptom actually appears - rather than inferred
from a 22-step sampler amplifying it somewhere else.

Serialised via a DisableParallelization collection, matching the convention of
the other tests here that touch process-wide state: the thread-budget case
mutates the ThreadPool ceiling, which an unrelated class running in parallel
would otherwise observe.

Bits() uses BitConverter.GetBytes rather than SingleToInt32Bits because this
project multi-targets net471, where that overload does not exist.

---

## What this PR is and is not

**It is** arithmetic-level coverage for a contract that was only ever asserted at flag level, plus a CI run of that coverage on hardware I do not have.

**It is not** a fix. I could not reproduce the downstream failure anywhere I control:

| Environment | Result |
|---|---|
| Windows, this CPU, 6 runs | passes |
| Linux container, same CPU, clean tree | passes |
| `Issue318` clone-drift diagnostics, this CPU | pass |
| These new determinism tests, this CPU | pass |

Everything points to the failure needing the CI runner's specific ISA. These tests will run there.

## How to read the CI result

- **If they fail on CI** — the determinism contract is not being honoured by whichever kernel those shapes select on that hardware, the mechanism is isolated, and the downstream `Clone()` divergence needs no clone-specific explanation.
- **If they pass on CI while the downstream shard still fails** — the cause is narrower than a general GEMM ordering difference, and #318's `SetParameters`-vs-trained-buffer provenance becomes the strongest remaining lead.

Either outcome is worth the run, which is why it is a test-only PR.

## Correction worth recording

I initially proposed that `options.PackedA is null && options.PackedB is null` explained the divergence — pre-packed weights taking a different kernel than unpacked ones. That is a real gate, but the consuming library never sets those options (zero references to `PackedA`, `PackedB`, `PackingMode`, `WeightPack` or `PrePack` in its entire source), so the branch is unreachable from that path and cannot be the explanation. Recorded here so nobody re-derives it.

https://claude.ai/code/session_017otuSGr3GdmvPoYLbiaWR2

## Summary by CodeRabbit

- **Tests**
  - Added coverage to verify that matrix multiplication produces identical results across repeated executions.
  - Added validation that results remain consistent under different thread configurations, including larger matrix sizes.
  - Improved safeguards around deterministic numerical behavior across supported execution scenarios.

### DIFF
--- a/tests/AiDotNet.Tensors.Tests/Engines/GemmDeterminismAcrossThreadsTests.cs
+++ b/tests/AiDotNet.Tensors.Tests/Engines/GemmDeterminismAcrossThreadsTests.cs
@@ -1,0 +1,168 @@
+using System;
+using System.Threading;
+using AiDotNet.Tensors.Engines.Simd;
+using Xunit;
+
+namespace AiDotNet.Tensors.Tests.Engines;
+
+/// <summary>
+/// The same GEMM, run twice in one process, must return bit-identical results.
+///
+/// <para><b>What is already covered, and what is not.</b> <c>DeterministicByDefaultTests</c> asserts the
+/// determinism CONTRACT at the level of flags and options — that <c>Deterministic</c> defaults to true, that
+/// <c>BlasProvider.IsDeterministicMode</c> follows it, that a plan is reused. Nothing there runs a GEMM and
+/// compares numbers, so a dispatcher that honours every flag and still returns different arithmetic on two
+/// calls would pass all of it.</para>
+///
+/// <para><b>Why that gap is worth closing.</b> <c>BlasManaged</c> chooses between kernels on properties that
+/// are not fixed by the caller: whether pre-packed operands were supplied, the thread budget, and which ISA
+/// the CPU reports (<c>Avx.IsSupported</c>, <c>Avx2.IsSupported &amp;&amp; Nr == 16</c>). Those kernels block
+/// the reduction differently, and floating-point addition is not associative, so two "numerically equivalent"
+/// routes legitimately disagree in the last bits. The GotoGemm path documents itself as "computed by one
+/// thread in fixed K order ⇒ thread-count-independent, Deterministic/DisableAutotune-contract safe", which
+/// says plainly that being thread-count-independent is a property some paths have and not others.</para>
+///
+/// <para><b>What prompted it.</b> A consumer's Clone() test compares a model against its clone through a
+/// 22-step denoising sampler, where a last-bit disagreement compounds. It fails intermittently on shared CI
+/// and passes on idle machines — the signature of a run-to-run difference rather than a clone defect. These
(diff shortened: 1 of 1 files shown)


ooples:
Confirmed and fixed in 558c36b6.

Measured directly on a 32-processor host with `threads = 1`, running the exact sequence this method uses:

```
before:               min=32   max=32767
during:               min=1    max=1
after (min-then-max): min=1    max=32767   restoreMin=False
after (max-then-min): min=32   max=32767
```

The restoring `SetMinThreads` returned **false** and the minimum stayed at 1. These settings are
process-wide, so every later test in the run executed against a starved pool, and nothing threw to
say so.

The `finally` block now restores the maximum first, which is the ordering the third line above
shows working. Comment records the measurement so the ordering is not "tidied" back later.

Issue #905, opened 2026-07-30
IEngine: no tape-level inverse trig, STFT detaches via out-params, TensorMean silently ends the graph


ooples:
## Summary

Three gaps in the `IEngine` surface make it impossible to express some standard losses as differentiable graphs. Each was hit while implementing APNet2's paper objective (Du et al. 2023, arXiv:2311.11545) in the consumer repo, and each forced a workaround rather than a call.

Checked against **0.119.0, 0.120.1, 0.120.2 and 0.120.3** — all three apply to every version.

---

### 1. No tape-level inverse trigonometry

`Atan`, `Asin`, `Acos`, `Atanh` are exposed only for `Vector<T>`:

```
M:AiDotNet.Tensors.Engines.IEngine.Atan``1(Vector{``0})
M:AiDotNet.Tensors.Engines.IEngine.Acos``1(Vector{``0})
```

There is no `TensorAtan` / `TensorAtan2` / `TensorAcos`. The tensor surface has forward trig (`TensorSin`, `TensorCos`) but no inverses, so the round trip cannot be closed.

**Why it matters.** Any objective over angles needs an inverse tangent. Phase-prediction vocoders (APNet, APNet2, Vocos) compute

```
Phi(R, I) = arctan(I / R) - (pi/2) * Sgn*(I) * [Sgn*(R) - 1]
```

The `Vector<T>` overload cannot be used because it is outside the autodiff graph — calling it silently severs the gradient rather than failing, which is the worst possible failure mode: training still runs and quietly optimises nothing through that path.

**Workaround used.** Composed `atan` from `TensorMultiply` / `TensorAdd` / `TensorDivide` / `TensorWhere` / `TensorReciprocal` / `TensorClampMin` as a range-reduced degree-9 minimax polynomial (~1e-7). Correct, but every consumer needing an angle has to reinvent it.

**Ask.** `TensorAtan`, `TensorAtan2`, `TensorAsin`, `TensorAcos` with the same tape support as `TensorSin` / `TensorCos`. `TensorAtan2` in particular, since the four-quadrant form is what callers actually want and the sign-correction dance above is easy to get wrong.

---

### 2. `STFT` returns through `out` parameters, so it cannot appear in a differentiable graph

```
M:...IEngine.STFT``1(Tensor{``0}, Int32, Int32, Tensor{``0}, Boolean, Tensor{``0}@, Tensor{``0}@)
                                                                     ^ magnitudeOut  ^ phaseOut
[...]

Issue #257, opened 2026-04-27
Critical: tape-recorded gradients keyed by stale intermediate refs — Embedding grad + 3/4 MultiHeadAttention weights silently dropped on backward (parent: AiDotNet#1208)


ooples:
## TL;DR

After **#255** landed in 0.58.1 (`Engine.TensorEmbeddingLookup` now calls `DifferentiableOps.RecordUnary`), end-to-end Transformer training is **still broken** in the parent AiDotNet repo. Verified via instrumentation against 0.58.1: the tape backward records **43 gradients per training step**, but a `Tensor<T>` reference-identity lookup — `grads[trainableParam]` — finds **only 12 of 16 trainable parameters**. The four that miss are:

| Shape | What it is | Status |
|---|---|---|
| `[8, 16]` | Token embedding (`_embeddingTensor`) | Tape registered the op, gradient is computed, but `allGrads` does NOT contain a `[8, 16]` entry at all. |
| `[16, 16]` | MultiHeadAttention `_queryWeights` | Same — registered, computed, never appears in `allGrads`. |
| `[16, 16]` | MultiHeadAttention `_keyWeights` | Same. |
| `[16, 16]` | MultiHeadAttention `_valueWeights` | Same. |

The 4th MHA `[16, 16]` weight (`_outputWeights`) DOES match — confirming this isn't "MHA layer is buggy" but "some specific op-graph configuration breaks the reference link."

The user-visible symptom: ooples/AiDotNet#1208 — `Transformer<T>.Train()` reports the correct loss magnitude but the model emits identical logits across distinct token inputs because `dL/dE` and 75% of attention weight gradients are silently dropped on the backward walk.

## Repro

Smallest model that surfaces all four missing-gradient cases. Run inside the parent AiDotNet repo against `Directory.Packages.props` pinned to `AiDotNet.Tensors` 0.58.1; the failure is shape-based and reproduces for any `Transformer<T>` instance with `numHeads >= 1`.

```csharp
var arch = new TransformerArchitecture<float>(
    inputType: InputType.TwoDimensional,
    taskType: NeuralNetworkTaskType.SequenceClassification,
    numEncoderLayers: 1, numDecoderLayers: 0,
    numHeads: 2, modelDimension: 16, feedForwardDimension: 32,
    inputSize: 4, outputSize: 8, maxSequenceLength: 4, vocabularySize: 8);

var model = new Transformer<float>(arch,
[...]

PR #1019, opened 2026-09-10
feat: add differentiable tensor asin, acos, atan and atan2

ooples:
## What

Adds `TensorAsin`, `TensorAcos`, `TensorAtan` and `TensorAtan2` to `IEngine`, implemented on
`CpuEngine`, overridden on `DirectGpuTensorEngine`, and recorded on the gradient tape. Also
documents a related hazard on `TensorMean` (issue #905 item 3).

## Why

Inverse trigonometry was already in the library, but only where autodiff could not see it:

- `IEngine` exposes `Asin`/`Acos`/`Atan` over spans and over `Vector<T>`, outside the tape entirely.
- The one tensor-level entry point, `NativeAtan2`, is listed in `OpRegistry.NonDifferentiableOps`.

So an objective defined over an angle - a phase term in a vocoder, a heading, the angle between two
unit vectors - compiles, runs, and trains, while the gradient through the angle is silently dropped.
Nothing throws and nothing warns. The model just never learns that path, which is the worst shape a
defect of this kind can take. Issue #905 reports working around it by composing arctangent out of a
dozen primitive ops.

## How

`CpuEngine.InverseTrig.cs` follows the existing `TensorSin`/`TensorSinh` template: GraphMode and
AutoTracer checks, `scope.RecordUnary`, the issue #257 contiguity fix-up that preserves the
user-facing reference before `Contiguous()` discards `GradFn`, per-type dispatch, then
`DifferentiableOps.RecordUnary`/`RecordBinary`. The three single-argument functions share one body,
differing only in the scalar applied and the derivative recorded.

Derivatives added to `BackwardFunctions<T>`:

| op | derivative |
| --- | --- |
| `asin(x)` | `1 / sqrt(1 - x^2)` |
| `acos(x)` | `-1 / sqrt(1 - x^2)` |
| `atan(x)` | `1 / (1 + x^2)` |
| `atan2(y, x)` | `x / (x^2 + y^2)` w.r.t. `y`, `-y / (x^2 + y^2)` w.r.t. `x` |

`TensorAtan2` requires both operands to have the same shape, not merely the same length - equal-length
tensors of different shapes would be paired by flat index, which is almost never what the caller
meant. `NativeAtan2` keeps its existing non-differentiable classification; this is an addition beside
it, not a change to it.

### GPU overrides

`BackendCompletenessTests` holds a floor of **zero** tensor-returning `IEngine` ops lacking a
dedicated `DirectGpuTensorEngine` override, and these four were about to raise it. The floor is
right, so they got real GPU paths instead of an exemption. The backends already carried
`Asin`/`Acos`/`Atan` kernels and an element-wise `atan2`, so this is dispatch rather than new
kernels.

One trap worth flagging for review: `Atan2Elementwise` is declared `(real, imag)` and computes
`atan2(imag, real)`, so the operands cross over - our `y` is its imaginary part and our `x` is its
real part. `Atan2_CoversAllFourQuadrants` pins `3*pi/4`, which would come back as `-pi/4` if that
order were wrong.

### `TensorMean` (item 3)

`TensorMean` returns a bare `T` and `ReduceMean` returns a tensor. Both are reasonable in isolation,
but the names give no hint that one of them ends the graph, so the obvious spelling of a loss
compiles and then never trains. Added the doc-comment the issue asks for. No behaviour change.

## Tests

16 tests in `InverseTrigGradientTests`. Locally green across `Engines.Autodiff` (932 passed) and the
`OpParity` suite including the backend-completeness floor.

Forward correctness against `System.Math` in double and single precision, the tails of `atan`, the
`asin + acos = pi/2` identity, the `sin(asin(x))` round trip, all four quadrants of `atan2`, angle
recovery from a sine/cosine pair, and shape-mismatch rejection.

The load-bearing one is `Atan2_GradientReachesTheInput_NotSilentlyZero`. Every forward assertion in
the file would also pass against an implementation that computes the right angle and records nothing
on the tape - that is precisely the pre-existing behaviour. Only the gradient assertions distinguish
the two, so that test is what actually guards the fix.

## A note on the test tolerance

Double-precision assertions compare to `1e-6`, which is loose for double and deliberately so. Where a
GPU is present `AiDotNetEngine.Current` is `DirectGpuTensorEngine`, and on a device without native
fp64 its kernels evaluate a `Tensor<double>` in single precision. That is a pre-existing property of
the engine rather than of these ops - a bare `TensorAdd` of two doubles reproduces it with no inverse
trigonometry involved - and the repo's other gradient tests already use `1e-6` for the same reason.
It still discriminates everything these tests are for: every asserted value is O(1), so a wrong
derivative, a swapped `atan2` operand or a collapsed quadrant misses by a fraction of a radian, not
by a part in ten million.

## Not in this PR

Item 2 of #905 - `STFT` returning through `out` parameters, so it cannot appear in a differentiable
graph - is untouched. A returning, tape-participating overload is real work rather than a signature
change, and the asymmetry with `ISTFT` deserves fixing properly. The issue stays open for it.

Refs #905

https://claude.ai/code/session_016vmi2eqPuCXgNVYa9NBcM5

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
@@ -270,6 +270,72 @@ internal static class BackwardFunctions<T>
         DifferentiableOps.AccumulateGrad(grads, inputs[0], grad, engine);
     }
 
+    /// <summary>d(asin(x))/dx = grad / sqrt(1 - x^2)</summary>
+    internal static void AsinBackward(
+        Tensor<T> gradOutput, Tensor<T>[] inputs, Tensor<T> output,
+        object[] savedState, IEngine engine, Dictionary<Tensor<T>, Tensor<T>> grads)
+    {
+        var grad = engine.TensorMultiply(gradOutput, InverseSqrtOneMinusSquare(inputs[0], engine));
+        DifferentiableOps.AccumulateGrad(grads, inputs[0], grad, engine);
+    }
+
+    /// <summary>d(acos(x))/dx = -grad / sqrt(1 - x^2)</summary>
+    internal static void AcosBackward(
+        Tensor<T> gradOutput, Tensor<T>[] inputs, Tensor<T> output,
+        object[] savedState, IEngine engine, Dictionary<Tensor<T>, Tensor<T>> grads)
+    {
+        var slope = engine.TensorNegate(InverseSqrtOneMinusSquare(inputs[0], engine));
+        var grad = engine.TensorMultiply(gradOutput, slope);
+        DifferentiableOps.AccumulateGrad(grads, inputs[0], grad, engine);
+    }
+
+    /// <summary>d(atan(x))/dx = grad / (1 + x^2)</summary>
+    internal static void AtanBackward(
+        Tensor<T> gradOutput, Tensor<T>[] inputs, Tensor<T> output,
+        object[] savedState, IEngine engine, Dictionary<Tensor<T>, Tensor<T>> grads)
+    {
+        var numOps = MathHelper.GetNumericOperations<T>();
+        var x = inputs[0];
+        var denominator = engine.TensorAddScalar(engine.TensorMultiply(x, x), numOps.One);
+        var grad = engine.TensorDivide(gradOutput, denominator);
+        DifferentiableOps.AccumulateGrad(grads, x, grad, engine);
+    }
+
+    /// <summary>
+    /// d(atan2(y, x))/dy = grad * x / (x^2 + y^2), d/dx = -grad * y / (x^2 + y^2).
+    /// </summary>
+    /// <remarks>
+    /// The four-quadrant form differs from atan(y/x) in its value but not in its derivative, which
+    /// is what makes it usable in a loss: the branch cut moves, the slope does not.
+    /// </remarks>
+    internal static void Atan2Backward(
+        Tensor<T> gradOutput, Tensor<T>[] inputs, Tensor<T> output,
+        object[] savedState, IEngine engine, Dictionary<Tensor<T>, Tensor<T>> grads)
+    {
+        var y = inputs[0];
+        var x = inputs[1];
+
+        var denominator = engine.TensorAdd(engine.TensorMultiply(x, x), engine.TensorMultiply(y, y));
+
+        var gradY = engine.TensorDivide(engine.TensorMultiply(gradOutput, x), denominator);
+        DifferentiableOps.AccumulateGrad(grads, y, gradY, engine);
+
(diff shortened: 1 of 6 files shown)


ooples:
I could not reproduce this, and the premise does not hold on either engine.

**`TensorMultiply` and `TensorDivide` do not reject `[1]` against `[]`.** Probed directly, on the
`DirectGpuTensorEngine` that is current here and on a plain `new CpuEngine()`:

| engine | `TensorMultiply([1], [])` | `TensorDivide([1], [])` |
| --- | --- | --- |
| DirectGpuTensorEngine | 12 (correct) | 3 (correct) |
| CpuEngine | 48 (correct) | 3 (correct) |

Both broadcast rather than throwing, so the shape-mismatch exception the finding is built on cannot
arise through these two calls.

**The end-to-end cases return correct gradients.** Rank-0 input through each of the four new ops:

- `TensorAsin` on a rank-0 tensor: grad 1.1547005, expected 1.1547005
- `TensorAtan2` on rank-0 operands: dy 0.4, dx -0.2, both expected
- Control arm, `TensorSin` on the same rank-0 shape (its `SinBackward` also omits `AlignGradRank`):
  grad 0.87758255, expected 0.87758256 - identical behaviour, so nothing here is specific to the new
  ops

**The suggested trigger does not produce a rank-0 value.** `ReduceSum(x, null)` returns rank **1**,
not rank 0, so `reduction -> asin` never reaches the state described. That path also returns correct
gradients.

**Adoption rate is evidence too.** `AlignGradRank` is applied in 5 of 267 backward functions. The
262 that omit it include `SinBackward` and `CosBackward`, the closest analogues to this change.
Adding the guard to these four and not to those would be inconsistent without a reproducer
justifying it.

I am not making the change on this evidence. If you have a concrete construction that reaches
`"Tensor shapes must match. Got [1] and []"` through any of these four backward functions, I will
take it and fix it - the guard itself is cheap and I have no objection to it, only to adding it
speculatively for a failure I could not make happen.

Full suite after the change: 4103 passed, 0 failed, and CI is green across all 17 jobs including
every `gpu-parity-pocl` tape-grad shard.

ooples:
CI is settled: **18 of 19 green**, including all three required checks (`Validate PR title`, `build`,
`avx512-verify`) and every `gpu-parity-pocl` shard.

The one red is `codecov/patch`, which is not a required check. The dip is the
`DirectGpuTensorEngine` overrides: the coverage job has no GPU, so `TryRunUnary`/`TryRunBinary`
return null and the recorded-output branches inside each override never execute - they fall through
to the base engine, which *is* covered. That is the case `codecov.yml` already calls out by name as
acceptable, "API surface wired to existing implementations (e.g. CUDA wrappers that delegate to a
covered CPU path)". The remaining uncovered lines are the generic-`T` arms of the per-type dispatch,
which only run for numeric types outside `float`/`double`.

The behaviour those lines implement is covered on the GPU path by the `gpu-parity-pocl` shards, which
pass.

PR #1021, opened 2026-09-10
feat: make istft differentiable and add stftphase

ooples:
## What

Closes the remaining half of issue #905: **item 2**, the STFT round trip.

- `ISTFT` is now differentiable, with `IstftBackward` as the genuine transpose of overlap-add
  synthesis.
- New `StftPhase` op, the phase counterpart of `Spectrogram`, on both the CPU and GPU.
- A pre-existing transform defect fixed along the way (see below).

Together with `Spectrogram`, both inputs of `ISTFT` are now differentiable functions of the waveform,
so an STFT-consistency objective `|| S - STFT(ISTFT(S)) ||` trains end to end.

## Why

The issue reports that `STFT` returns through `out` parameters and so cannot appear in a
differentiable graph, while `ISTFT` "returns its result and composes fine". **That second half is not
true** - both were in `OpRegistry.NonDifferentiableOps`. The real division is that `STFT`/`ISTFT` are
deliberately unrecorded primitives and `Spectrogram`/`MelSpectrogram` are the recorded entry points
built on them. So the actual gaps were narrower and different from the report:

1. Synthesis had no gradient at all. A consistency objective is defined *across* the round trip, so
   half of it was untrainable while the loop still ran and reported a number.
2. No differentiable path returned **phase**. `Spectrogram` is magnitude-only by construction.

## How

**`IstftBackward`** transposes each forward stage in reverse: the window-sum normalisation (under
the same `1e-8` guard the forward uses, or the two disagree exactly where the forward declined to
divide), the windowed scatter-add inverted to a gather, the unnormalised inverse DFT whose transpose
is a forward one, the Hermitian mirror folded back onto bins `1..numFreqs-2`, and the polar map
`dL/dm = gRe cos p + gIm sin p`, `dL/dp = m (gIm cos p - gRe sin p)`.

**`StftPhase`** mirrors `Spectrogram` exactly. Magnitude and phase share every step of their adjoint
except the per-bin seed, so `MagnitudeStftAdjoint` becomes a thin forwarder onto one implementation
selected by an `StftAdjointTarget`. The phase seed is the magnitude seed rotated a quarter turn and
scaled by `1/|C|` - which is also the only reason the phase target needs a floor, since at zero
magnitude the angle is undefined rather than merely steep.

**Both engines record.** A GPU path computing the right values and recording nothing would make
trainability depend on which engine happened to be current, which is the same defect one level down.
The GPU `StftPhase` reuses the `StftMagPhase` kernel `Spectrogram` already runs.

**`GriffinLim` and `TimeStretch`** call `ISTFT` internally and stay non-differentiable, so their
internals are wrapped in `NoGradScope` - otherwise a tape-active `GriffinLim` would unroll 60
iterations onto the tape. On the GPU engine that scope also covers its momentum arithmetic, which was
*already* recording `TensorSubtract`/`TensorAdd` nodes while the CPU engine recorded none. The two
engines now agree.

## A pre-existing defect this surfaced

`DifferentiableOpsGradCheckSweep` failed on ISTFT, and the cause was not the adjoint.

**`FFTCore` zero-pads a non-power-of-two length to the next power of two**, which computes a
different transform - its bins sit at `k/nPadded` rather than `k/n`. This was already known and
documented on `UnnormalizedTransformForAdjoint`, but only the **adjoint** had been moved onto the
exact transform. At those lengths `STFT` and `ISTFT` were therefore disagreeing with their own
transposes, silently. Both now use it; identical call for power-of-two lengths, which is every
realistic one. The helper loses its `ForAdjoint` suffix since it is no longer adjoint-only.

## Verification

**4207 passed, 0 failed** across the autodiff, op-parity, completeness, FFT and audio suites.

Gradients are checked by central differences under **Richardson extrapolation**, on an explicit
`new CpuEngine()` so the check runs in true double rather than the fp32 a GPU engine uses for
`Tensor<double>` on a device without native fp64.

Two measurements say the residual is roundoff rather than a systematic error: extrapolation drops the
worst relative disagreement over 64 samples from **1.07e-6 to 1.41e-7**, and its sign is balanced
**31 positive / 33 negative**. A missing term would neither shrink with step size nor balance in sign.

## What adversarial review of the first commit changed

- **Every ISTFT gradcheck used `center: false`** - mine and the sweep's synthesized argument alike -
  leaving the centred path, with its negative write offsets and out-of-range trimming, unchecked.
  Both finite-difference tests are now theories over both settings.
- **The `NoGradScope` keeping `GriffinLim` off the tape had no test.** A scope that silently stopped
  working would surface nowhere, because the forward value is identical either way.
  `GriffinLim_StaysOffTheTape` covers it.
- **`StftPhase`'s GPU override had no coverage at all.** It is float-only and needs a backend, which
  the double-precision gradient tests never provide, and OpParity is a hand-maintained registry
  rather than reflective. Added parity cases - which then failed, exactly as intended.

## On the parity cases

They compare `cos` and `sin` of the phase, not the phase itself. Two correct implementations may
report `+pi` and `-pi` for the same angle and differ by exactly `2*pi`: comparing raw values, CPU gave
`3.1415927` and GPU `-3.1415923` at one bin, a 2157060020-ULP "disagreement" that is no disagreement.
`cos` and `sin` are continuous across the cut and the pair pins the angle uniquely.

**No tolerance was widened for this.** The tape-gradient bar is hardcoded at `1e-4` for all ops and is
untouched. The seed keeps every bin off the cut, so both cases run at the default tolerances - a
fixture fix rather than a relaxed bound.

Refs #905

https://claude.ai/code/session_016vmi2eqPuCXgNVYa9NBcM5

## Summary by CodeRabbit

- **New Features**
  - Added differentiable STFT phase output for CPU and GPU engines.
  - Added gradient support for inverse STFT, including magnitude and phase inputs.
  - Enabled phase-based audio analysis and reconstruction workflows.

- **Bug Fixes**
  - Improved STFT and inverse STFT accuracy for odd and non-power-of-two transform lengths.
  - Prevented Griffin-Lim processing from recording unsupported gradients.
  - Improved handling of near-zero-magnitude phase bins to avoid unstable gradients.
  - Improved consistency between CPU and GPU phase-processing results.

### DIFF
--- a/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
@@ -8929,10 +8929,10 @@ internal static class BackwardFunctions<T>
     /// imaginary parts of <c>FFT(gz_f)</c>.
     /// </description></item>
     /// <item><description>
-    /// <b>Hermitian extension and the polar map.</b> The forward copied bins
-    /// <c>1 .. numFreqs-2</c> to <c>nFft-k</c> with the imaginary part negated, so each of those
-    /// bins receives two contributions while DC and Nyquist receive one. Then <c>C[k] = m
-    /// exp(I*p)</c> gives <c>dL/dm = gCRe cos(p) + gCIm sin(p)</c> and
+    /// <b>Hermitian extension and the polar map.</b> The forward copied every bin with
+    /// <c>2k &lt; nFft</c> to <c>nFft-k</c> with the imaginary part negated, so each of those bins
+    /// receives two contributions while DC - and Nyquist, when <c>nFft</c> is even - receives one.
+    /// Then <c>C[k] = m exp(I*p)</c> gives <c>dL/dm = gCRe cos(p) + gCIm sin(p)</c> and
     /// <c>dL/dp = m * (gCIm cos(p) - gCRe sin(p))</c>.
     /// </description></item>
     /// </list>
@@ -9033,7 +9033,10 @@ internal static class BackwardFunctions<T>
                     double gradReal = numOps.ToDouble(adjointReal[k]);
                     double gradImag = numOps.ToDouble(adjointImag[k]);
 
-                    if (k >= 1 && k < numFreqs - 1)
+                    // Same bound as the forward mirror, for the same reason: 2k < nFft rather than
+                    // k < numFreqs - 1, which differ exactly when nFft is odd. The fold has to track
+                    // whatever the forward copies or this stops being its transpose.
+                    if (k >= 1 && (k * 2) < nFft)
                     {
                         gradReal += numOps.ToDouble(adjointReal[nFft - k]);
                         gradImag -= numOps.ToDouble(adjointImag[nFft - k]);
--- a/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
+++ b/src/AiDotNet.Tensors/Engines/Autodiff/BackwardFunctions.cs
@@ -8856,6 +8856,164 @@ internal static class BackwardFunctions<T>
     private static string Describe(object? o) => o is null ? "null" : o.GetType().Name;
 
     /// <summary>
+    /// Adjoint of <c>ISTFT</c>: maps a gradient shaped like the reconstructed waveform back onto
+    /// the magnitude and phase spectrograms it was synthesised from.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// The forward is four stages, and this transposes each in reverse. Writing <c>w</c> for the
+    /// window, <c>s(f)</c> for the write offset of frame <c>f</c>, and
+    /// <c>W[j] = sum_f w[j - s(f)]^2</c> for the window-sum envelope:
+    /// </para>
+    /// <list type="number">
+    /// <item><description>
+    /// <b>Normalisation.</b> <c>y[j] = a[j] / W[j]</c> where the envelope exceeds 1e-8. <c>W</c>
+    /// depends only on the geometry, never on magnitude or phase, so its transpose is the same
+    /// division applied to the incoming gradient, under the same 1e-8 guard. Dropping that guard
+    /// would make the adjoint disagree with the forward exactly where the forward declined to
+    /// divide.
+    /// </description></item>
+    /// <item><description>
+    /// <b>Windowed overlap-add.</b> <c>a[s(f) + i] += w[i] * z_f[i] / nFft</c>, summed over frames
+    /// and skipping out-of-range writes. The transpose of a scatter-add is a gather:
+    /// <c>gz_f[i] = w[i] * gy[s(f) + i] / nFft</c>, and zero wherever the forward wrote nothing.
+    /// </description></item>
+    /// <item><description>
+    /// <b>Inverse DFT, real part.</b> <c>z_f[i] = Re( sum_k C[k] exp(+2*pi*I*k*i/nFft) )</c>. The
+    /// transpose of an unnormalised inverse transform is an unnormalised FORWARD one, so
+    /// <c>gCRe[k] = sum_i gz_f[i] cos(2*pi*k*i/nFft)</c> and
+    /// <c>gCIm[k] = -sum_i gz_f[i] sin(2*pi*k*i/nFft)</c>, which are exactly the real and
+    /// imaginary parts of <c>FFT(gz_f)</c>.
+    /// </description></item>
+    /// <item><description>
+    /// <b>Hermitian extension and the polar map.</b> The forward copied bins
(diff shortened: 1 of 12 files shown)


ooples:
All four findings verified against the code and fixed. Two were not what the passing tests suggested.

### The round-trip test was vacuous

You were right, and the demonstration is worth recording. The loss was built from
`reconstructed - waveform`, which gives the waveform a direct edge to the loss. Severing the entire
analysis-synthesis chain with `StopGradient` and running the *same* assertion:

```
with the round trip DETACHED, waveform gradient non-zero = True
```

So it passed for a reason unrelated to what it claimed. The loss is now built from the reconstruction
alone and asserts on `magnitude` and `phase` as well. The same probe against the new form:

```
severed -> waveform=absent magnitude=absent phase=absent
```

It now fails when it should.

### The GPU override did lack inference-trace handling

`GraphCaptureParityTests` passes for both `StftPhase` cases, which reads as a refutation until you
open it - it constructs `new CpuEngine()`, so it exercises the base path and never reaches the GPU
override at all. Green, and non-discriminating. The override now bails to base on
`IsInferenceTrace`, matching the one existing precedent (`GroupNorm`), since base is the only
implementation that calls `CaptureInferenceKernel`.

### The buffer leak is real

`FinishGpuOp` takes ownership of the buffer it is handed, so a throw in the kernel launch, the
`Synchronize`, or the first handoff stranded both allocations while the `catch` fell back to the CPU.
Fixed with the handed-off/`finally` idiom used elsewhere in the file.

Extended to `Spectrogram`, which is the code this was copied from and carries the identical leak. The
inference-trace bail is deliberately **not** extended there: that changes which engine computes a
result, whereas the disposal fix changes nothing on the success path. Flagging rather than fixing it.

### Odd nFft was argued, not tested

The fold covers exactly the range the forward mirrors, so it is the transpose either way - but that
was an argument and every gradcheck ran at `nFft = 8`. Added gradchecks at `nFft = 5` and `7`:
magnitude and phase both agree with central differences to **1e-7**. That also exercises the
Bluestein path the transform fix in this branch enabled; before it, the forward zero-padded while the
adjoint did not, and these cases would have failed outright.

---

**4209 passed, 0 failed** locally after the fixes, with a full-solution build under
warnings-as-errors on net10.0/net8.0/net471.

### Still unaudited, for the record

The zero-padding defect fixed here for `STFT`/`ISTFT` also affects every other `FFTCore` caller:
`RFFT` (41573), `FFT` (41746), `IFFT` (41810), `FFT2D` (41881) and `IFFT2D` (41952, 41975). Each
computes bins at `k/nPadded` rather than `k/n` for a non-power-of-two length. Out of scope here -
this PR fixed the two where the forward disagreed with its own adjoint - but worth its own pass.

https://claude.ai/code/session_016vmi2eqPuCXgNVYa9NBcM5

ooples:
Both remaining findings are fixed. The odd-length one was correct and larger than I first credited.

### Odd-length Hermitian mirroring - real, and in the forward

I initially answered this with a gradcheck at `nFft = 5` and `7`, which passed. That was the wrong
instrument: a gradcheck compares the adjoint against the forward, so it passes whether or not the
forward is a correct inverse. It was measuring the wrong thing and I cited it as though it settled
the question.

Measuring the actual round trip instead, worst interior error of `ISTFT(STFT(x))`:

| nFft | parity | error |
| --- | --- | --- |
| 8 | even | 4.4e-16 |
| 16 | even | 6.7e-16 |
| **5** | **odd** | **0.055** |
| **7** | **odd** | **0.0035** |

Thirteen orders of magnitude. The forward stopped mirroring at `numFreqs - 1`, which is right only
when the last reported bin is Nyquist and self-conjugate. An odd `nFft` has no Nyquist bin, so the
top of the spectrum stayed zero, the inverse transform was not of a conjugate-symmetric spectrum, its
result was not real, and taking the real part discarded the difference.

The bound is now `2k < nFft` in the forward and the adjoint fold tracks it. That bound is provably
identical to the old one for every even length:

```
nFft  4 -> [1]           same        nFft  5 -> [1] vs [1,2]        CHANGED
nFft  6 -> [1,2]         same        nFft  7 -> [1,2] vs [1,2,3]    CHANGED
nFft  8 -> [1,2,3]       same        nFft 17 -> ... vs ...+[8]      CHANGED
nFft 16 -> [1..7]        same
```

So no power-of-two caller changes at all. `Istft_InvertsStft` now pins both parities and is the test
the gradchecks structurally could not be.

This was reachable because of the transform fix earlier in this branch - before it a non-power-of-two
`nFft` was zero-padded to the next power of two and wrong for an unrelated reason.

### Inference-trace handling

Fixed. The override now bails to base on `IsInferenceTrace`, as `GroupNorm` does, since base is the
only implementation that calls `CaptureInferenceKernel`. Worth noting that `GraphCaptureParityTests`
could never have caught this: it constructs a `CpuEngine`, so it exercises the base path and never
reaches the override - it was passing for both `StftPhase` cases throughout.

https://claude.ai/code/session_016vmi2eqPuCXgNVYa9NBcM5

